PR #26, opened 2011-09-29
Napster support

jlewin:
Hi David,

I've been working on updated Napster support and wondered if you'd take a look. I've added an option for the user to enable/disable the Napster module so it can be easily disabled independent of the main Chrome Scrobbler.

Additionally, I've put a simple mechanism in place to display album artwork on the notification windows that falls back to traditional behavior if the image data is absent. It's easy enough to pull but I thought it enhanced the experience and wanted to offer it up as a suggestion.

Thanks,
John

### DIFF
--- a/napster.js
+++ b/napster.js
@@ -111,9 +111,12 @@ function init() {
       for(var i=0;i<=1;i++)
       {
          var trackImage = document.getElementById('nx_large_image_' + i);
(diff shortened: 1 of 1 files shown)


Hekemon:
Napster website scrobbling stopped working today when the site updated. Any way to make it work again?

PR #327, opened 2014-03-03
Cleaner, more reliable connector for Digitally Imported

Daniel15:
Rather than listening to DOM mutations, this injects a script into the page's scope (via a similar method to what the Jango and PlugDJ connectors do) and listens directly to JavaScript events fired by their player.  Unlike the Jango and PlugDJ connectors which use a hacky method for communication (listening to DOM mutations on a hidden `div`), this one communicates to the extension via `window.postMessage` which is a lot cleaner. 

The script to inject the script is also cleaner and has been generalised in `inject.js` for easy reuse. The Jango and PlugDJ connectors could be updated to use this method at some point in the future. I've left todo comments in the code.

Open question: Is `window.postMessage` the cleanest way to do this? Should I be using an extension-specific API instead?  [This page](https://developer.chrome.com/extensions/messaging) documents `chrome.runtime.connect` but I can't seem to find this? `chrome.runtime` is null for me.

@pronskiy - What do you think?

### DIFF
--- a/connectors/difm-dom-inject.js
+++ b/connectors/difm-dom-inject.js
@@ -1,0 +1,23 @@
+/**
+ * Last.fm Scrobbler for Chrome - Digitally Imported connector
+ * @author Daniel Lo Nigro - http://dan.cx/
+ *
+ * This script runs in the context of di.fm itself. It listens for events fired
+ * by their JavaScript, and propagates these events to the extension via
+ * `postMessage`.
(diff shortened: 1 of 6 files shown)


inverse:
Is this obsolete now?

Daniel15:
It's still relevant, I just need to rebase it for any recent changes.

PR #438, opened 2014-08-20
Adds Napster Connector

hughrawlinson:


inverse:
@hughrawlinson we should close this and rewrite this in v2

PR #439, opened 2014-08-20
Adds Rhapsody Connector

hughrawlinson:


inverse:
@hughrawlinson same with this

PR #441, opened 2014-08-20
fixes rdio artist bug

hughrawlinson:
We noticed a bug caused by the order of rdio's dom manipulation when updating the player. They would update the track, causing us to updateNowPlaying before they update the artist. It caused incorrect scrobbles, so I've added a timeout to address this, as well as making sure a user has listened to a certain track for a reasonable amount of time for it to be considered a 'play'. I thought I had fixed this in #435, but it looks like I hadn't, so I'm closing that now.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,7 +3,7 @@
 /* global module, require */
 module.exports = function(grunt) {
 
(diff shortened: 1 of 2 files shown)


janbaykara:
We'd be better off building a v2 connector for Rdio, then going from there. Close this?

david-sabata:
That's right --> #781

PR #525, opened 2014-12-30
Musicase Support

joe50261:
* Added support for Musicase (http://musicase.me/)

Screenshots
https://drive.google.com/file/d/0B_2eSXlwVD17ZGxVS0NhTjY0cEE/view?usp=sharing
https://drive.google.com/file/d/0B_2eSXlwVD17SWF1aDFzN1pLMUk/view?usp=sharing
https://drive.google.com/file/d/0B_2eSXlwVD17R19objZESWVfTXM/view?usp=sharing


david-sabata:
I checked this out and it seems to scrobble all tracks right away. Are you seeing this behavior too?
Retrieving track duration from validation response and passing it to nowPlaying request should help.

If it doesn't, you can still try the new v2 connector format. It's still documented in the code only, but it's extremely easy and there are already some connectors in the `connectors/v2` folder.

joe50261:
This is because I scrobble the start of songs, instead of end or progress.
I did that because there are some difficulty of implementation.
I will try to implement based on duration again or the v2 API.
2015-1-24 上午5:27 於 "David Šabata" <notifications@github.com> 寫道：

joe50261:
The upgrading to v2 connector was almost done. It seems to work successfully with the built-in delay, even the information of duration is lacking.

Issue #528, opened 2015-01-18
Add support for bop.fm


MWYang:
Please and thank you!

inverse:
Feel free to run with the branch I've put up and let me know how it goes.

MWYang:
Seems to work great! I had no idea the code changes were so easy. Next time I'll know how to do it myself. :D

david-sabata:
Closed by #529

PR #529, opened 2015-01-18
Added support for bop.fm

inverse:
Simple implementation that uses the V2 connector.

I noticed that the bop.fm site also has a really nice `Bop.Player` JavaScript object which we could bind too and listen to events and get further meta data from, such as duration.

refs #528

### DIFF
--- a/connectors/v2/bopfm.js
+++ b/connectors/v2/bopfm.js
@@ -1,0 +1,10 @@
+
+Connector.playerSelector = '#bop-player';
+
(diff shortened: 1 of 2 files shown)


david-sabata:
Nice! That's the simplicity what I wanted v2 connectors to be :)

Hope you didn't hit any problems.

inverse:
Nothing wrong at all - was super easily! For sites that have such players it's great!

PR #533, opened 2015-01-26
Add support for listen.tidalhifi.com, closes #485

### LINKED ISSUE #485: Support for TIDAL?
New music service that launched yesterday. listen.tidalhifi.com.

theli-ua:


inverse:
refs #485

theli-ua:
@inverse can you access http://play.wimpmusic.com/  then? that is from the same company and I believe it uses the same frontend.
Also does http://listen.tidalhifi.com/ really not work for you from Berlin? If you don't log in you can test it in "demo" mode - ie just press play on something, playback will be limited to 20 seconds though.

inverse:
Both URLs don't work. They just show blank pages. However SSH tunnelling  to the UK works but I'm not up for creating an account :P

theli-ua:
I don't have WIMP account either.
If I open http://play.wimpmusic.com/  I can press play on any , say, playlists on front page and it'll start playing. Same thing for listen.tidalhifi.com . This should be enough to test, no?

inverse:
My bad - It turns out it was the adblocker (µBlock) causing both sites to load. 

Just tried it out and and success :) awesome work!

PR #534, opened 2015-01-28
Fixed Ghostly

inverse:
Fixed Ghostly and updated it to new format.

Is it worth removing the old connector as well?


david-sabata:
Yes please, remove the old one. Thanks for updating! ;-)

inverse:
Done :) Awesome to see your back!

PR #545, opened 2015-02-25
Fix for SC

inverse:
SC updated their code again and now the event bus is located @ index `16` again. Could be worth getting this validated by someone else before pulling this in though.

### DIFF
--- a/connectors/soundcloud-dom-inject.js
+++ b/connectors/soundcloud-dom-inject.js
@@ -12,10 +12,10 @@ window._ATTACHED = window._ATTACHED || false;
     // Exit if already attached.
     if (window._ATTACHED) return;
     // Attach event listeners to the event-bus.
-    webpackJsonp([1], 
-        { 
-          0: function(e, t, n) { 
-            bus = n(17);
+    webpackJsonp([1],
+        {
+          0: function(e, t, n) {
+            bus = n(16);
             bus.on('audio:play', function(e) {
                 window.postMessage({
                     type: 'SC_PLAY',


llisama:
I was linked into here, but sorry, I still don't understand how to fix the problem...

inverse:
@llisama If you are not comfortable with running unpacked extensions and using git then you'll have to wait for @david-sabata to review/merge this and get another release ready I'm afraid.

ledgeri:
I'm arrived with the same problem, so now i see the satus.. I would say a "Hurry Up", but i need to say "Thanks for the app!!"
@inverse: Whats are the "unpacked extensions" in this case?

inverse:
@ledgeri I've never used Opera so I don't know if this is possible.

ledgeri:
@inverse As i realised the keyword here maybe can be the "based on Chromium": the Chrome addons can be used on Opera, and Vica versa... So please tell me about what is your solution... i will try it, and whatever is the final result,we learned one more thing...

inverse:
Basically you will need to find a way to install the extension within Opera from the source code from the branch that I have put up here on github. I've not used Opera so I am unsure how this is done but I am sure it's pretty similar to in Chrome. Check the readme for more information.

llisama:
@inverse, hiya, is it easily explainable? to run unpacked extensions/git?

ledgeri:
@llisama https://developer.chrome.com/extensions/getstarted#unpacked that is what i found, but not tried yet... (i had other things to do..)

PR #548, opened 2015-02-27
Private Mode

clementlamoureux:
- Added private mode (no data will be passed to LastFm when the option is checked)
- Added option in options.html page
- Added custom pageAction status (In private Mode)
- Added quick menu to checked/unchecked option on other pageAction(s)

For a personnal need, need to be continued


janbaykara:
Could you not achieve exactly this by ... disabling/uninstalling the extension?

clementlamoureux:
I needed this to be quickly disabled, and re-enable as quickly. It's just for myself

Issue #541, opened 2015-02-21
Deezer not work


NAUSHNIK52:
Last.FM Scrobbler can't recognize Deezer song (I think this error appeared about two weeks ago) and suggest to enter name.
![image](https://cloud.githubusercontent.com/assets/4209275/6314614/dba53efe-b9f6-11e4-99e4-fbe8025bbbb8.png)

david-sabata:
Can someone else confirm this? I just played few random tracks and all were recognized just fine.

@NAUSHNIK52 some tracks are not recognized by L.FM, because their database is never complete. Can you post a link to track that is causing problems?

NAUSHNIK52:
@david-sabata All tracks cause problem. My deezer.js fixed scrobbling for me.

david-sabata:
I am on beta too. Is it possible that Deezer has some new UI that is open for some users only? Or maybe difference based on country?

NAUSHNIK52:
@david-sabata Oh, looks like it is. I am just changed language to English and my fix won't work anymore.

david-sabata:
That's interesting. Can you somehow detect language and make your fix active for RU only?

NAUSHNIK52:
@david-sabata I think I can't. I don't know JS very well (making fix for me took about 2 hours).
Maybe I am try later. I don't think Russian is one language with this problem.

david-sabata:
I understand, but I don't want to apply a change that breaks something that previously worked. I'll note this to the PR. Maybe someone will show up and make the fix language-dependent.

inverse:
The HTML tag binds the language attribute. Could use that to hook in a fix on a per language basis.

inverse:
should be fixed in next release

PR #550, opened 2015-02-28
Update deezer.js

NAUSHNIK52:
Last 3 weeks scrobbling won't work for me. There is fix for scrobbling Deezer.
This is my first fix ever so don't hate me please.

### DIFF
--- a/connectors/deezer.js
+++ b/connectors/deezer.js
@@ -91,7 +91,7 @@ function getCurrentTrack()
     else if ($('.player .control-pause').size() > 0) {
         return {
             track: $('.player-track-title').text(),
-            artist: $('.player-track-artist').text().replace($('.player-track-artist').children().first().text(), '').trim().replace(/,$/, ''),
+            artist: $('.heading-2-sub').children('.player-track-link').text(),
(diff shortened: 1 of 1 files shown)


david-sabata:
Waiting to see how #541 resolves

david-sabata:
Won't merge now. This fix works only for russian language version of Deezer and simultaneously breaks english version. 

An update to make this change for RU only is needed.

NAUSHNIK52:
@david-sabata I am just checked another languages in interest. Now my fix works with English and Russian at same time. Maybe Deezer updated design for another countries? Test please.

inverse:
Not a deezer user but I've just tested and the existing one is working in both EN and RU locales. Perhaps it's a geolocation rollout.

When putting `STATIC_VERSION` into the console what do you get returned? Here I'm getting `00342213` - Perhaps this is the build number.

NAUSHNIK52:
@inverse Suprisely, original Deezer.js now works fine for me.

janbaykara:
@NAUSHNIK52 if it works now, do you think it's worth closing this PR?

PR #563, opened 2015-03-27
Added support for somaFM

inverse:
This works and is fine. But is it better to try and calculating a rough play time based off a diff from the time the current track started (`HH:MM`) and now?

image


david-sabata:
@inverse you can always improve. I'm not sure measuring play time that way would work if someone pauses the track though.

PR #571, opened 2015-04-08
Add MixRadio Connector support for www.mixrad.io

nhtua:
I have clone a connector for www.mixrad.io, used code from Grooveshark Connector. It will resolve for issues 476, and my problem too, subscribe MixRad.io and log history by Last.fm.


david-sabata:
Hi @nhtua you are using an outdated connector format. Can you please migrate your PR to `v2`? The new connectors are much simpler and offers new features. You can take a look at the code in `connectors/v2/`

Issue #546, opened 2015-02-26
Soundcloud scrobble not working


llisama:
Hasn't worked for me for quite some time now. Wasn't sure if it was just me or overall. I've re-authenticated etc. can't seem to find the problem.

cheers

inverse:
Should be fixed with #545 - You could try it out and see if you are comfortable with running unpacked extensions + git.

munframed:
Same problem here. It shows it scrobblable, but doesn't scrobble. No problem with lots of other websites like youtube, google music player, or even tunein, but with Soundcloud, yes. Please fix it :/

inverse:
It's fixed. Just needs to be merged and released.

munframed:
You mean it's fixed but hasn't released yet? When should I expect it then? Thanks for your answer btw.

inverse:
@munframed  - It's been merged - should be in the next release.

david-sabata:
Bad news - SC seems to be broken again. I was just checking before release... :(

munframed:
:( thanks for your efforts I do appreciate it. Keep me posted please :)

inverse:
It's like cat and mouse :(

Yoshi8765:
Not working for me yet again. :-1:

apaulvs:
Hey, I don't know the status on this issue but as of the past few weeks, Soundcloud hasn't been scrobbling for me either at all. The extension doesn't even seem to be running on the website. :/

llisama:
Anything here? Soundscloud has just never worked for me. I'm running the unpacked version...

david-sabata:
@llisama can you post an example url of what you cannot scrobble? Also please try using regular package. A lot of changes were done for SC in last two releases.

For me the SC scrobbling seems to be working fine now.

nicoletheghoul:
not working again :/

PR #573, opened 2015-04-08
SC fix

inverse:
We need a more robust way as SC push code changes frequently which adjusts their packed resources.

Something like this would improve this:

```
			var i;
			for (i = 0; i < 255; i++) {
				var test = n(i);
				if (typeof test === 'object' && typeof test._events === 'object' && typeof test._events.all !== 'undefined') {
					break;
				}
			}

			bus = n(i);
```

Although it's only a basic test looking for the correct object but would provide more resilience. Opinions?
refs #546

### DIFF
--- a/connectors/soundcloud-dom-inject.js
+++ b/connectors/soundcloud-dom-inject.js
@@ -15,7 +15,7 @@ window._ATTACHED = window._ATTACHED || false;
     webpackJsonp([1],
         {
           0: function(e, t, n) {
-            bus = n(16);
+            bus = n(19);
             bus.on('audio:play', function(e) {
                 window.postMessage({
                     type: 'SC_PLAY',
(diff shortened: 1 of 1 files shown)


lourinaldi:
:+1:

AltarBeastiful:
+1, I just wrote the same patch before seeing this PR!

To avoid searching for the bus each time we could try the hardcoded index and fall back on searching it if it's not the bus event : 
````js
function isEventBus(f) {
    return typeof f === 'object' &&
           typeof f._events === 'object' &&
           typeof f._events.all !== 'undefined';
}

bus = n(19);
if (!isEventBus(bus)) {
	var i;
    for (i = 0; i < 255; i++) {
        var test = n(i);
        if (isEventBus(test)) {
  			bus = test;
  			break;
        }
    }
}
````

inverse:
@AltarBeastiful Thanks - see the updated commit

AltarBeastiful:
+1 , @inverse Thanks for the swift update.

Issue #540, opened 2015-02-19
Please scrobble the music at ektoplazm.com


w-sky:
Would be great if scrobbling would work for Ektoplazm. They have terabytes of (electronic) music, free to play in the browser.
http://www.ektoplazm.com/free-music

inverse:
Not sure if it's possible. The music is being played through a flash mp3 player which doesn't look like any of the DOM state gets updated.

bslotems:
I've created a connector for ektoplazm: https://github.com/david-sabata/web-scrobbler/pull/588/

bslote:
I've created a connector for ektoplazm: https://github.com/david-sabata/web-scrobbler/pull/588

PR #589, opened 2015-05-06
Ektoplazm connector

bslote:
Created connector for ektoplazm.com.

Requested in issue #540.


gordol:
:+1: still

david-sabata:
@bslote Hi! Sorry to say, but the connector is in the old format which we don't want to support anymore. Because of that no new connectors will be merged. Can you please migrate to `v2` connector format? You can check `connectors/v2` directory for examples and also `core/content/connector.js` for some documentation.

PR #591, opened 2015-05-07
Fallback cover art via MusicBrainz/CoverArtArchive API

janbaykara:
The code in this PR works for a fair number of (but not all) songs. I'm not sure that the file I've placed the logic in is the best place, but the mechanics of it work fine in practice.

I've only done this as I noticed the extension was no longer really ever getting imagery from Last.fm.

### DIFF
--- a/core/background/notifications.js
+++ b/core/background/notifications.js
@@ -87,7 +87,8 @@ define([
 						type: 'basic',
 						iconUrl: song.metadata.artistThumbUrl || 'icon128.png',
 						title: song.getTrack(),
-						message: 'by ' + song.getArtist()
+						message: 'by ' + song.getArtist(),
+						priority: 1
 					};
 
 					chrome.notifications.create('', options, notificationCreatedCb);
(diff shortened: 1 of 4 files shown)


david-sabata:
@janbaykara this looks cool :)

Only the place is not ideal. This way notifications could be delayed by slow response from MusicBrainz.
I think the best place is something I call "processing pipeline". Here's how L.FM metadata is loaded  https://github.com/david-sabata/web-scrobbler/blob/master/core/background/pipeline/lfm-metadata.js . MB should be next step in the pipeline and should fill the thumbnail in song's structure, if it's not filled yet.

Btw. I see no problems with current images - they are loaded almost every time. Of course they are supported for `v2` connectors, so that's probably what you mean.

janbaykara:
@david-sabata The deed is done, I've moved it to a pipeline module!

david-sabata:
@janbaykara great, thanks! :+1:

Issue #593, opened 2015-05-08
YouTube scrobbles as youtube.com/devicesupport


maxdaniel98:
When using YouTube, web-scrobbler scrobbles as youtube.com/devicesupport.

inverse:
dupe of #590

MansourMans:
On youtube it scrobbles track as  <youtube.com/devicesupport> and artist as <http>
![snap](https://cloud.githubusercontent.com/assets/12377342/7552839/df54f7c0-f6d6-11e4-9aa5-cbdb12782fa3.JPG)

Issue #590, opened 2015-05-06
YouTube has disabled their old data API, rendering the YouTube connector nonfunctional


soliloquial:
YouTube is now returning the following for all data API requests from the old API:

[...]

PR #594, opened 2015-05-08
Bump youtube connector to new API

aemxdp:
To fix recent youtube problems #590 and #593. As for me, the only problem here is API key and quota. I've generated this one specifically for web-scrobbler and don't mind even if you just merge it as-is, but I'm not sure how much people are using it and how much tracks they scrobble per day. As scrobble statistics of http://www.last.fm/music/https suggests its just few tens of thousands scrobbles from all the scrobblers on the web per day, so maybe quota is not a problem.
For details, take a look at:
https://developers.google.com/youtube/v3/docs/videos/list
https://developers.google.com/youtube/v3/getting-started#quota

### DIFF
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -155,7 +155,6 @@ function cleanArtistTrack(artist, track) {
 function parseDuration(durationString) {
   var parts = durationString.substr(2).split(/[HMS]/);
   var count = parts.length;
-  var duration = 0;
   switch (count) {
     case 2: return parseInt(parts[0]);
     case 3: return parseInt(parts[0]) * 60 + parseInt(parts[1]);
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -10,11 +10,6 @@ var lastUrl = '';
 // loaded asynchronously from storage; default to false for backwards compatibility
 var scrobbleMusicOnly = false;
 
-// we will observe changes at the main player element
-// which changes (amongst others) on ajax navigation
-var target = null;
-var feather = false;
-
 function handleMutation (mutation) {
 	// detect first mutation that happens after url has changed
 	if (lastUrl != location.href) {
@@ -27,23 +22,16 @@ var observer = new MutationObserver(function(mutations) {
     mutations.forEach(handleMutation);
 });
 
-
 var config = {
 	attributes: true,
 	attributeFilter: ['data-youtube-id', 'src']
 };
 
-target = document.querySelector('#player-api video');
-if (!target) {
-    feather = true;
-    updateNowPlaying();
-}
-else {
-	// trigger manually to detect change from '' -> 'something' on first regular page load
-	handleMutation(null);
+var target = document.querySelector('#player-api video');
 
-    observer.observe(target, config);
-}
+handleMutation(null);
+observer.observe(target, config);
+
 // bind page unload function to discard current "now listening"
 $(window).unload(function() {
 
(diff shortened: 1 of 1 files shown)


inverse:
:+1: awesome work! can confirm this working. Things to be wary of:

- API quote limits with it currently set to 50,000,000 per day this shouldn't be a real issue. I guess we could have a pool of API keys to try but that'll surely be against their T&Cs
- Key security - again there isn't much IMHO that can be done with that in such a web application. Aside having some code hosted doing the API calls and using it as a proxy.

EhsanKia:
Is this not applicable here? https://developers.google.com/identity/protocols/OAuth2UserAgent

Daniel15:
You could just make every user create their own API key and enter it on the settings page.

david-sabata:
This looks good, can you please rebase to current master?

I'll merge it now, but I'm not sure it will survive upgrade to v2. Not that it wouldn't work, but I don't like much relying on Google's API when all data is already there, rendered on the page in some way.

inverse:
I'd say go down the OAuth route instead of the API key for each user __if needed__. OAuth should solve all the limitation issues but adds another step during the installation of the extension.

aemxdp:
@david-sabata Ok, rebase done. I'd like to notice a problem I see with parsing page: video category. It seems you can't really determine if the video is music in general case without API. Though there might be some heuristics which will detect music fine enough.

david-sabata:
@aemxdp thanks! Just one more note came to my mind. The connector will probably stop working if something goes wrong with the API key, like Google bans it or its request limit is reached. How about making the API request optional - in a way that if it fails, classic DOM parsing will still take on?

aemxdp:
@david-sabata Nice idea, check last commit. As for "scrobble music only", it ignores this option and tries to scrobble if API request fails because I can't imagine any easy ways to determine if the video is music without API. Also, when adding duration parsing from page, I've found out other DOM parsing code is dealing with "feather" mode. I've tried to find such a mode on current youtube but it seems it was removed years ago, so I removed the "feather" stuff. To check if it does the right thing I've tried it with invalid url and invalid key (separately).

AltarBeastiful:
Looks good. Works with the new youtube player as well.

Could we get this merged?

connectors/youtube.js
```diff
@@ -148,137 +154,110 @@ function cleanArtistTrack(artist, track) {
...
-      var artist = null;
-      var track = null;
-      var artist_dom = '';
-      var track_dom = '';
-      if (!feather) {
-        // Use the #eow-title #watch-headline-show-title if available
-        track_dom = $('#eow-title').clone();
-        artist_dom = $('#watch-headline-show-title', track_dom);
+  xhr.onreadystatechange = function () {
+    var info = null;
```
inverse: Shouldn't we only continue if the readyState is 4?

e.g. add the following just below the function decoration?

```js
  if (xhr.readyState !== 4) {
  	return;
  }
```

aemxdp:
@inverse Of course, my fault. I didn't notice because always turning off notifications at the beginning and forgot to check them.

Daniel15:
If you use onload rather than onreadystatechange, you don't need to check
the status.

david-sabata:
Actually there's jQuery in the context, so its xhr wrapper can be used instead of low level `XMLHttpRequest`. (If Chrome doesn't have some kind of policy against this.)

But that's just for further improvements. I think we're good to merge now. Thanks all :)

PR #599, opened 2015-05-12
Use Last.fm's default cover art image as a fallback

janbaykara:
##### Ref
* http://cdn.last.fm/flatness/responsive/2/noimage/default_album_300_g4.png - Last.fm's default track icon as seen here:

##### Idea
* Use Last.fm's site default cover art, for tracks scrobbled with no cover art. See new (top) and old (bottom):

##### Why?
* Less garish and distracting than the extension's black/white shiny icon; cleaner
* More in keeping with Last.fm's aesthetic
* Fits with the chrome notification style


inverse:
Nice idea. I think It's pretty awesome :+1:

david-sabata:
Awesome, thanks!

PR #600, opened 2015-05-12
CI changes

inverse:
- Added v2 connectors to build
 - Declared Connector as ignore since its implicit with how it's being called
 - Made connectors have `use strict` on global level

Open to discussion if this is the best changes to make if you feel otherwise.


david-sabata:
:+1:

PR #607, opened 2015-05-17
V2 Connectors track art + BC implementation

inverse:
Added the ability for connectors to specify what image to display in the notification. 

Falls back to the current mechanisms if nothing specified.


janbaykara:
Ace idea, looks good +1

PR #608, opened 2015-05-17
Scrobble album title when using Google Play

markjustin:
There was a very small error in the new fix for Google Play. The new album selector tag is actually .player-album not #player-album (a class not an ID). I've tested this change out and now album titles are scrobbled as well as the artist and song title.


futchadoc:
Sorry - can you please give me more info on how to fix the album scrobbling? I am barely computer literate (OK that might be an understatement, but I've never done anything like this before). Thanks.

PR #610, opened 2015-05-18
Added Reddit Music Player support

inverse:
Title says it all :)

Totally digging v2 connector development!


david-sabata:
@inverse Love to hear that! It's awesome to see so many new connectors thanks to v2 :)

PR #623, opened 2015-05-27
Update Spotify connector to accommodate new layout

### LINKED ISSUE #620: Spotify overhaul ruined music recognition
Google chrome Version 43.0.2357.81 m using the webplayer. It was working fine yesterday and I was receiving notifs/scrobbling for each song played.

Paszt:
resolves #620

### DIFF
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -206,6 +206,13 @@ define(function() {
 		},
 
 		{
+			label: 'Spotify',
+			matches: ['https://play.spotify.com/*'],
+			js: ['connectors/v2/spotify-play.js'],
+			version: 2
+		},
+
+		{
 			label: 'Plug.dj',
 			matches: ['*://plug.dj/*'],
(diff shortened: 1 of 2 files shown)


core/connectors.js
```diff
@@ -200,7 +200,7 @@ define(function() {
 
 		{
 			label: 'Spotify',
-			matches: ['https://play.spotify.com/*'],
+			matches: ['https://player.spotify.com/*'],
```
hughrawlinson: So, this would appear to be a problem - Spotify seem to be splitting this change geographically. I guess this PR is for the new interface in the US(?), which hasn't been rolled out globally. We may have to keep both connectors around for the foreseeable future.

To clarify:
1: The changes to the player aren't being observed in the play.spotify.com player in the UK.
2: player.spotify.com redirects to play.spotify.com in the UK at least, probably elsewhere too.

Otherwise, :+1:, looks good.

PR #624, opened 2015-05-28
Update Amazon to V2

Paszt:


david-sabata:
@Paszt it's the UTF BOM ;-)

But nevermind and thanks for contributing!

Issue #622, opened 2015-05-27
Add support for kollekt.fm


inverse:
link: https://kollekt.fm

david-sabata:
Closed by #628

PR #628, opened 2015-05-31
Added kollekt.fm

inverse:
Added support to kollekt.fm

@Paszt @david-sabata  would you be able to take a look? How would be better to implement the checking for the existence of the player-controls element on the DOM?

I was thinking of changing the following in the starter to something like this:

```js
	console.log('Web Scrobbler: Setting up stateChangeObserver');

		var playerSelectorObserver = new MutationObserver(function() {
			if (document.querySelector(Connector.playerSelector) !== null) {
				playerSelectorObserver.disconnect();
				var stateChangeObserver = new MutationObserver(Connector.onStateChanged);
				stateChangeObserver.observe(document.querySelector(Connector.playerSelector), {
					childList: true,
					subtree: true,
					attributes: true,
					characterData: true
				});
			}
		});

		playerSelectorObserver.observe(document.body, {
			childList: true,
			subtree: true,
			attributes: false,
			characterData: false,
		});
```

Perhaps this could be implemented as a flag that controllers can override 

`Connector.waitForPlayer = true` or something.

Also I noticed that if the track/artist is null on the new connectors then notifications are null. Ill try to look into this later. I guess because null is a valid artist :/ will have to filter these out from the pipeline.

refs #622

### DIFF
--- a/connectors/v2/kollekt.js
+++ b/connectors/v2/kollekt.js
@@ -1,0 +1,39 @@
+'use strict';
+
+/* global Connector */
+
+(function() {
+	var playerObserver = new MutationObserver(function() {
+		if (document.getElementById('player-controls')) {
+			playerObserver.disconnect();
+			var actualObserver = new MutationObserver(Connector.onStateChanged);
+			actualObserver.observe(document.querySelector('#player-controls'), {
+				childList: true,
+				subtree: true,
+				attributes: true,
+				characterData: true
+			});
+		}
+	});
+
+	playerObserver.observe(document.body, {
+		childList: true,
+		subtree: true,
+		attributes: false,
+		characterData: false,
+	});
+
+})();
+
+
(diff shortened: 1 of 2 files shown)


david-sabata:
@inverse I think (hope) that this wouldn't be a frequented case. But you're right that some mechanism of waiting for player to be ready might come handy. Otherwise people would invent their own ways, which are often wrong :)

I don't see any better way than the one you presented. Using observers should be fine even in pages that change elements often and in high quantities. Only thing that gives me bad feeling is observing the `body`, but again, I don't think there's a better way. 

I'd consider starting the double-observing automatically in starter if the player element is not present at the time of injection. That way the flag won't be needed and we could prevent potential timing issues, which could otherwise cause headaches to people who try to add a connector but don't see what's actually happening under the hood.

What do you think?
I'm merging this anyway. We can continue in some new issue.

PR #629, opened 2015-05-31
Prevent null artist/track being scrobbled

inverse:
I noticed that if the artist and or track was `null` this would still attempt to go through the whole processing pipeline. This check will prevent it from going through as apparently null - null is a valid song on last.fm.

Perhaps this is abusing the `isLastfmValid` but the checks are strict type. I'm open to suggestions of where such rejection should take place? Perhaps higher up in the controller?

refs: #628

### DIFF
--- a/core/background/services/lastfm.js
+++ b/core/background/services/lastfm.js
@@ -246,6 +246,13 @@ define([
 	 * @param cb {Function(boolean)} callback where validation result will be passed
 	 */
 	function loadSongInfo(song, cb) {
+
+		if (song.processed.artist === null || song.prcoessed.track === null) {
+			song.flags.attr('isLastfmValid', false);
+			cb(false);
+			return;
+		}
+
 		var params = {
 			method: 'track.getinfo',
 			autocorrect: localStorage.useAutocorrect ? localStorage.useAutocorrect : 0,


david-sabata:
This fix should be fine. Returning `false` from the L.FM service correctly leaves user with unknown track and allows him to fix it manually.
I probably missed the fact that such song could exist on L.FM :))

Generally the idea is, that connector sends as much data as it can gather. Then controller and pipeline should get something from it and evenutally scrobble it. Theoretically (not implemented yet) the connector could send just `uniqueID` and the track info would be looked up in some internal storage.

inverse:
An internal 'cache' of valid tracks could be a good thing  :) although the cost of a HTTP request vs the often streaming of music is somewhat small. But every performance gain is good :+1:

david-sabata:
Actually I sometimes notice that L.FM servers have lower performance than usual. Resulting in delay between the time a track starts playing and the time the notification shows up. Which is totally a bad user experience, because it seems like the extension didn't notice the track is already playing.

On the other side it may be regional issue. Also the cache would be efficient mainly for "lovers" who play their loved tracks over and over, and less for "explorers". ... But still, I want to get this feature too some day.

PR #633, opened 2015-06-03
Fix SoundCloud injection

AltarBeastiful:
Default index changed to 2459. Also increased search range to 0-3491, above nothing is defined.

Now calling n() with an invalid index throws a TypeError, catching those.

Added some logs, if the default index change again we need to be notified quickly as the loop to search a new index can take up to 288ms.

### DIFF
--- a/connectors/soundcloud-dom-inject.js
+++ b/connectors/soundcloud-dom-inject.js
@@ -29,9 +29,8 @@ window._ATTACHED = window._ATTACHED || false;
             if (!bus || !isEventBus(bus)) {
                 var i;
                 for (i = 0; i < 3491; i++) {
-                    var test;
                     try{
-                        test = n(i);
+                        var test = n(i);
                         if (isEventBus(test)) {
                             bus = test;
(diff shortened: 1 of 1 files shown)


connectors/soundcloud-dom-inject.js
```diff
@@ -20,17 +20,33 @@ window._ATTACHED = window._ATTACHED || false;
...
+
+            if (!bus || !isEventBus(bus)) {
                 var i;
-                for (i = 0; i < 255; i++) {
-                    var test = n(i);
-                    if (isEventBus(test)) {
-                        bus = test;
-                        break;
+                for (i = 0; i < 3491; i++) {
+                    var test;
```
inverse: Why is `var test` declared like this?

AltarBeastiful: No idea  :)

inverse:
Looks good aside from this :+1:

gmeben:
@AltarBeastiful @david-sabata Any idea when this will be merged?

david-sabata:
Thanks! Hopefully this will protect us from SC's magic :)

PR #635, opened 2015-06-05
Fix youtube scrobbling

AltarBeastiful:
Api used seems to be deprecated. Updated the code to use v3 of youtube API.

Only thing missing is the API key from web-scrobbler. How are those handled usually?

### DIFF
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -183,28 +183,22 @@ function updateNowPlaying() {
       return;
    }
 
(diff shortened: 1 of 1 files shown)


inverse:
You should check out #594 - This is already requested to fix the issue.

AltarBeastiful:
Indeed, I'll close this and nag @david-sabata to merge #594 !

david-sabata:
@AltarBeastiful your nags have been heard ^_^

Issue #643, opened 2015-06-23
Soundcloud scrobbling broken (again)


jdb8:
#633 seemed to fix it, but now I'm getting `Cannot scrobble. Please report at https://github.com/david-sabata/web-scrobbler/issues` in the console.

Using v1.48 from the chrome store.

joshkaplan:
Might be worth referencing this, their soundcloud scrobbling works: https://github.com/msfeldstein/chrome-media-keys

inverse:
@joshkaplan  I think what they do is control the player which binds to DOM elements. We could do that with this scrobbler but binding to the event bus feels more pragmatic in this instance. The changes that are in PR #648  will make this a lot more robust @Paszt is doing some good work with bringing all the scrobblers to v2 as well!

maegentae:
Soundcloud is not scrobbbling again.

PR #648, opened 2015-06-25
Upgrade Soundcloud to V2 & new event-bus detection

Paszt:
New Soundcloud event-bus detection cycles through webpack modules to find the module with the correct functions instead of relying on the module number which will change each time they recompile the module. Also upgrade to V2 with getTrackArt  and getUniqueId methods included.

Should replace PR #645 (tried to fork that PR but couldn't).
ref #643


inverse:
Much better :+1:

AltarBeastiful:
Scrobbling is perfectly fine. We even gained track art!

I got 2 console logs : 
```
starter.js:52 Web Scrobbler: Connector.playerSelector is empty. The current connector is expected to manually detect state changes
starter.js:14 starter already loaded
```
They might be normal tought. 

@david-sabata What's the use of those?

inverse:
@AltarBeastiful The first notice is more for developer reasons to indicate that the connector is using a non-conventional route of binding audio events. Normally it would be binding to the 'player' and watching for modifications but in this instance @Paszt has bound to the SC event bus.

The second notice  is because of the controller has already been started. Perhaps this is something we can avoid doing? @Paszt opinions?

Paszt:
@inverse: I don't have a great understanding of the web-scrobbler core js, but I've just traced back the error.  It originates with the chrome.tabs.onUpdated event.  According to the [docs](https://developer.chrome.com/extensions/tabs#event-onUpdated), a tab status can be loading and complete.

I don't think it's anything in the connector but instead something different about soundcloud's website.

david-sabata:
Great job, thanks!

inverse:
:+1:

PR #649, opened 2015-06-25
Update Soundcloud bus index

AltarBeastiful:
Also increased max search index to another hopefully large enough value.

For now the scrobbling is broken again so we should push a new version quickly.
Altought after this one we should finally be fine. Maybe.

Possible improvements could be that we are automatically notified when the script starts to notice a new bus index. Can we send ourselves a message from the extension somehow?


inverse:
It's fixed

AltarBeastiful:
Yup, again forgot to look before posting :)

Closing this one.

inverse:
No worries its good to see people being active :)

PR #651, opened 2015-06-29
Add isStateChangeAllowed function, update slacker2 & mixcloud

Paszt:
isStateChangeAllowed function allows the mutationObserver to ignore certain states, for example if an advert is playing.  Updated slacker2 & mixcloud connectors to include isStateChangeAllowed.  Added trackArtImageSelector to slacker2.

### DIFF
--- a/core/content/starter.js
+++ b/core/content/starter.js
@@ -34,7 +34,11 @@
 	if (Connector.playerSelector !== null) {
 		console.log('Web Scrobbler: Setting up observer');
 
-		var observer = new MutationObserver(Connector.onStateChanged);
+		var observer = new window.MutationObserver(function () {
+			if (Connector.isStateChangeAllowed()) {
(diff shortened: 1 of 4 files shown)


inverse:
:+1: Awesome work!

core/content/starter.js
```diff
@@ -34,7 +34,11 @@
 	if (Connector.playerSelector !== null) {
 		console.log('Web Scrobbler: Setting up observer');
 
-		var observer = new MutationObserver(Connector.onStateChanged);
+		var observer = new window.MutationObserver(function () {
```
inverse: Do we need to add `window.`?

Paszt: jshint checks for undefined variables.  It wouldn't pass the grunt test without including window.

david-sabata:
Looks great, thanks! :+1:

PR #665, opened 2015-07-13
Adding Tradiio connector

andremoniz:
Adding Tradiio.com connector to web scrobbler.
Tradiio is a new music discovery App, where users decide and help upcoming artists rise. 
Become more than a passive listener. Play your part in your new favourite artists’ careers.

Artists that perform well, get opportunities such as label deals, festival perfomances, music videos and even radio play.

### DIFF
--- a/connectors/v2/tradiio.js
+++ b/connectors/v2/tradiio.js
@@ -11,11 +11,11 @@ Connector.trackSelector = '.leftmusicplayer .musicname a:first';
 Connector.playButtonSelector = '.leftmusicplayer .section-controls .btn-play';
 
 Connector.getDuration = function() {
-    return Connector.stringToSeconds($('.leftmusicplayer .songDuration').text()) || null;
+	return Connector.stringToSeconds($('.leftmusicplayer .songDuration').text()) || null;
 };
 
 Connector.currentTimeSelector = '.leftmusicplayer .progress-play';
 
 Connector.getTrackArt = function () {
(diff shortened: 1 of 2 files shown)


andremoniz:
Connector is not defined?
Hi David is this a major problem?

Paszt:
This looks good.  

The reason it's failing jshint is because you changed the inline configuration.  The following line tells the  jshinter that `Connector` is a global variable:

    /* global Connector */
  
which you changed to `/* Tradiio Connector */`.  Just change it back and it should pass.

To make the connector more complete, I would suggest adding the current time and track art functionality.  Current time will allow the extension to detect repeated plays.

    Connector.currentTimeSelector = '.songDuration.s-progress';
    
    Connector.getTrackArt = function () {
    	return $('.cover-small img').attr('src').replace('-square.jpg', '-normal.jpg');
    };

Cheers.

andremoniz:
Hi @Paszt Many thanks for your help.
Fixed

Paszt:
@andremoniz - Why did you change the value of currentTimeSelector?  It won't work with the changes you made.  currentTimeSelector is expecting a jquery selector that will find the element containing track current time in h:m:s format.  The value of `'.leftmusicplayer .progress-play'` is the progress bar which will always have a text value of white space.  `'.songDuration.s-progress'` will give you the current time.

andremoniz:
Fixed

david-sabata:
@andremoniz did you push the change?

andremoniz:
I did a new pull request

PR #683, opened 2015-08-05
Add dreamfm.biz support

fedorg:


connectors/v2/dreamfm.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function getTrack() {
+	var track = $('#tracktitle').text() ||
+				null;
+	return cleanText(track);
+}
+
+/* @returns {{artist, track}} */
+Connector.getArtistTrack = function () {
```
Paszt: This method is not really needed since it is called only when either getArtist or getTrack returns an empty result.

/blob/master/core/content/connector.js#L154
/blob/master/core/content/connector.js#L263

connectors/v2/dreamfm.js
```diff
@@ -0,0 +1,70 @@
...
+/* global Connector */
+
+Connector.playerSelector = '#player';
+
+/**
+ * remove zero width characters & trim
+ * @param  {string} text to clean up
+ * @return {string} cleaned up text
+ */
+function cleanText(input) {
```
inverse: We should probably put this in the base connector now that two extensions are using similar functionality.

fedorg: I don't know if there is a real need for this cleaning. This site delivers tracks from Bandcamp, so it made sense to copy Bandcamp's connector behavior. Also, as vaporwave titles usually contain a lot of junk characters, a little sanitization is probably ok.

PR #702, opened 2015-08-21
Song info popup, Last.fm love/unlove UI

janbaykara:
Building on @herrbengtsson's UI popup, fixed up and added a way to love/unlove tracks from within it (as discussed in #598).

Some styling could be done in future, along with other song info / interaction features (e.g. the API data we receive has Track-User Play Count), but I feel like this is a good start.


janbaykara:
To do for next PR request:
[ ] Style popup to match Chrome notification
[ ] Add play count / other user-specific statistics

david-sabata:
@janbaykara this is awesome! Exactly what I had in mind :)))

More ideas 
* [ ] countdown before the track gets scrobbled
* [ ] edit track info, even though it was recognized
* [ ] button to "never scrobble this track" (would require `uniqueID` to be implemented by current connector)

janbaykara:
Groovy, will work on these points in the next few weeks.

Re: editing track info - surely that would involve re-scrobbling a song, as the initial scrobble will have been sent off - any idea how we'd work around spamming multiple scrobbles?

david-sabata:
The correction is already there, even for v2 connectors, just in another popup.

I think users usually notice bad track info soon, mostly before time to scrobble. So the track is only marked as "now playing", which is harmless. And if they edit the info later there's not much we can do. I would go to library and delete the last scrobble. Sure it can be done, but it seems as too much work for too few users to me.

inverse:
Yeah the notifications are platform in dependent and I wouldn't style them in this way. What's there looks good so far. Nice job :+1:

janbaykara:
@david-sabata Hrmm. 

> Button to "never scrobble this track" (would require uniqueID to be implemented by current connector)

A potential issue with "never scrobble this track" option is we'd then need to store the UIDs of the songs we don't want to scrobble. That could become a bit cumbersome... ?

david-sabata:
@janbaykara we'll be storing UIDs and some preferences (like custom labels) one day anyway... And functionality that depends on it can be enabled only for songs that have the UID set

fedorg:
There's now a difference in behavior when the user clicks the icon. It either does this popup thing or disables scrobbling. I think the icon should always display a popup, and there should be a button or something to skip scrobbling this track, forever or only this time. It would be nice to see some consolidation.

fedorg:
Also, clicking the notification just hides it and does nothing else. I think it would be appropriate to put skip options there, or maybe something else more functional. This may be a bad idea, though - KISS and everything, nobody likes visual clutter.

david-sabata:
I suspect pausing the scrobbling won't be something people would use multiple times a day. Therefore I wouldn't neccessarily put it into the notification and rather use the popup.

And I'm aware of the difference in popup behavior in v1 and v2 connectors. But that's kinda intentional. I don't want to put any more effort in supporting new features for v1 connectors. Rather update all connectors to v2.

fedorg:
Looks ugly enough
scrobbler_skip_concept1

janbaykara:
@fedorg This is how my branch janbaykara/web-scrobbler looks at the moment: https://www.dropbox.com/home/Screenshots?preview=Screenshot+2015-08-24+15.03.48.png

PR #705, opened 2015-08-21
Fix #621 adds myradio.ua

### LINKED ISSUE #621: Please add support myradio.ua
I like this online radio 
http://myradio.ua/player/7
But scrobler does not work with it
Thank you

galeksandrp:
Maybe

#621

Seems http://myradio.ua/player/playlists not works, only radio

### DIFF
--- a/connectors/v2/myradio.ua.js
+++ b/connectors/v2/myradio.ua.js
@@ -7,21 +7,21 @@ Connector.playerSelector = '.cur-blk';
 Connector.separators = [' -             '];
 
 Connector.getArtistTrack = function () {
-    var text = $('.name').text().substring(12);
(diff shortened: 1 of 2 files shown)


connectors/v2/myradio.ua.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '.cur-blk';
+
+Connector.separators = [' -             '];
```
bes-internal: Not sure that this is the best way. Maybe we need to apply .trim() to artist and track in the original function?

galeksandrp: Yes, your solution if more flexible. Thanks!

PR #710, opened 2015-08-23
PageAction file reorganisation

janbaykara:
Store away icons and popup files into dedicated directories, clearing up the root dir.


janbaykara:
Nix that, I'll redo this after David finishes his changes.

Issue #694, opened 2015-08-13
Connector request: radio.yandex.ru


TimO-n:
I recently started listening to Radio Yandex, but unfortunately they do not offer Last.fm scrobbling. Would it be possible to create a connector for their website. It would be awesome. URL: https://radio.yandex.ru/

roamn:
Yandex support lastfm in profile settings, you can link this service to yandex.
And scrobble yandex music and radio.

roamn:
Screenshot of settings page: http://i.imgur.com/avW71tJ.png

bes-internal:
Connector already in Version: 1.50 (d9e152783) at chrome web store.

david-sabata:
Is the Radio Yandex working using this extension working or not?
Using internal L.FM connection is not a real solution for us :)

TimO-n:
@david-sabata Yandex advertising inserts and extension stop reading after headers songs

galeksandrp:
Don't see any advertising even without AdBlock. It works for me.

We need to split artist names from 'Nancy Sinatra, Billy Strange' and catch first.

bes-internal:
@TimO-n can you copy page source with advertising? What we want:
1. on page with adv
2. right click on page on any empty place
3. above in code find root container <div class="page-root">
4. right click on it and "edit as html"
5. select and copy all (ctrl+a, ctrl+c) in textarea

TimO-n:
now in extension have problem with scrobbling. Beta LastFM can't authenticate with new service api

bes-internal:
we only need the code page, the extension does not matter here

1v:
Connecting account here https://music.yandex.ru/settings working fine on https://radio.yandex.ru/.

alexesprit:
The extension works fine with YandexRadio now.

PR #715, opened 2015-08-24
Fix artist name yandexradio

galeksandrp:
#694

### DIFF
--- a/connectors/v2/yandexradio.js
+++ b/connectors/v2/yandexradio.js
@@ -3,8 +3,8 @@
 /* global Connector */
 
 Connector.playerSelector = '.page-station__bar';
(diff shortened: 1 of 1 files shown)


bes-internal:
confirm

connectors/v2/yandexradio.js
```diff
@@ -3,8 +3,8 @@
 /* global Connector */
 
 Connector.playerSelector = '.page-station__bar';
-Connector.trackSelector = '.slider__items > div:nth-child(3) .track__title a';
```
galeksandrp:
```html
<div class="track__title">
<a href="">First author</a> & <a href="">Second author</a>
</div>

PR #718, opened 2015-08-24
Fixes some minor auth problems

janbaykara:
Should bring an end to `session.trim() not a func` errors.


bes-internal:
Yes, with this patch, after the re-authentication it works

david-sabata:
Thanks for quick fix. Although I don't seem to be able to reproduce the issue :-|

PR #746, opened 2015-08-30
Add wonder-fm

galeksandrp:


connectors/v2/wonder.fm.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '#menu';
+
+Connector.getArtistTrack = function () {
```
bes-internal: Why here has default implementation of this function?

galeksandrp: wonder.fm
```javascript
				track = text.substr(0, separator.index);
				artist = text.substr(separator.index + separator.length);
```
default
```javascript
				artist = text.substr(0, separator.index);
				track = text.substr(separator.index + separator.length);
```
Feel the difference.

bes-internal: ok

Issue #535, opened 2015-02-05
Youtube repeat button support


NukeTheMoon:
Pressing the repeat button after a track on youtube finishes playing should start scrobbling the track again.

david-sabata:
This functionality is included in v2 connectors. Unfortunately YouTube is not migrated to v2 yet.

PR #752, opened 2015-09-01
Upgrade Youtube connector to v2

janbaykara:
I'm not sure why this hasn't already been done. I'm also not sure why the v1 connector was calling the API, other than to check the category, but that cat info is now in the HTML.


janbaykara:
Fixes #535

connectors/v2/youtube.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+
+/* global Connector */
+
+var scrobbleMusicOnly = false;
+chrome.storage.local.get('Connectors', function(data) {
```
inverse: Should this be using the storage abstraction?

janbaykara: I believe not, as the connector is run content-side, so does not have access to the background environment objects like ChromeStorage.

david-sabata: @janbaykara you're right. The storage abstraction is not injected into page. I don't think it's necessary and we should probably inject as many scripts as possible (yet still convenient for connectors impl.)

I'm thinking about unified configuration getter for current connector somewhere in base connector, but this is totally ok now ;-)

david-sabata:
The API was there because we couldn't tell if the video is from music category. But I hated it :smile: . Awesome that you got rid of it!

PR #768, opened 2015-09-06
Fix #765

### LINKED ISSUE #765: ERR_CONNECTION_TIMED_OUT
Trying to re-authenticate scrobbler, getting "ERR_CONNECTION_TIMED_OUT".

galeksandrp:
#765


david-sabata:
As much as I don't like switching from HTTPS back to HTTP, it seems that L.FM really handles all API traffic on HTTP anyway :-1:

PR #719, opened 2015-08-24
Song-Info popup improvements

janbaykara:
Polished up the song-info popup, and added a few useful features:
- [X] Countdown to scrobble timer (mentioned in #702)
- [X] Track/artist edit mode (may satisfy #245, #371, #619) (replaces `manual_scrobble.html` popup with song info popup in edit-mode)
- [X] User track play-count
- [ ] Ban/unban song from scrobbling (referencing Unique ID)
- [ ] Pause/resume scrobbling ( #616, #1196 )

Potential discussion points: usefulness of displaying things like total track time (see above screenshot) or [album name](https://www.dropbox.com/s/6ipeknczic05y90/Screenshot%202015-08-24%2015.03.48.png?dl=0) (see commented out code in /popups/info.html)?

PR #776, opened 2015-09-10
Add timestamp to notification

janbaykara:
Another simple one. Adds the time of play to the notification. Useful in situations where you've been playing things all day, and have a big list such as:

screenshot 2015-09-10 14 11 38


janbaykara:
I might actually add the website source to this. E.g. 
> 12:59pm • BBC Radio Player

galeksandrp:
:+1: connector source

janbaykara:
Album is displayed in the popup, if it exists.

Paszt:
There's no reason the album can't be in both the notification and the popup.  Especially since the popup is a page action and to see it, the user has to go to the tab playing the music which most would most likely already have the album displayed without the need for displaying the popup.

It would be nice if instead of a page action it was a browser action, but that would be a pretty big restructuring of the extension.

janbaykara:
@Paszt I kinda feel like the album is peripheral information, relative to track/artist, and hence a bit more at home in the pageAction, which I see as a 'more info' kind of domain. 

Also, I do like the light-weight nature of the notifications, and we'd likely end up spilling into a new line with the album, thereby making it taller.

janbaykara:
screen shot 2015-09-13 at 16 50 07

Paszt:
Perhaps you didn't read it, so I'll say again. In order for the user to open the popup, they have to display the tab that is playing the music which already has the album information (if it's available in the popup, it's available on the page playing the music) - at that point there's no need to display the popup if the user is trying to find out the album.  I see the album as more important information than the source and time.  If I want to see the time, that information is available on last.fm.

fedorg:
@Paszt I think it's better to discuss the popup in the relevant pull request #719.
As for the album title, I think it's better to keep the notifications as clean as possible. I don't see any reason to include the connector name in them, for example. That said, I would love to see the album title AND time in the notification, too, but there are some issues:
 - What if it's not an album, but a playlist or a mixtape?
 - What if the time overlaps with anything?
Adding a fourth line to the notifications makes them feel like a wall of text, which may or may not appeal to some people. I think that the ideal solution would be to make a separate setting to control notification "thickness", if anything. There's no magic bullet.

janbaykara:
* Rationale for time: as in the screenshot in the OP, you might have been listening to music all day, and want to quickly scan through your notification list to find out when something was played, time-wise.
* Rationale for connector: you are probably on a different tab, or using a different app, when you scrobble the track. So when you see the notification, you recognise why it came up - i.e. what website made a notification about a song appear.

@Paszt I see your point about navigation redundancy etc., but I do feel that album information is, as @fedorg has pointed out, a very variably beast, and also one that is often a bit irrelevant. So what if the song has the Album Y - maybe it was originally in Album X, and included in Y as a B, and then years later also included in Album Z. Last.fm scrobbling is fundamentally about Artist+Track, and album plays very much second fiddle to that. Just my opinions though - we can of course include that information anywhere, but... idk... I value lightweight, highly relevant notifications to information-full, contextual notifications; rather leave that to the webpage or the popup.

Paszt:
I really don't know what to say.  Not sure how adding the album makes it heavy.  But I guess I can just run off my fork instead as showing the album makes perfect sense to me.  If there's not an album name, well then you just don't display anything - simple as that.

david-sabata:
There are really multiple cases for notifications, but I think we can quite easily have a new section in settings with checkboxes for values to be included in notifications. Same goes for buttons, if we ever add some, which may present a distraction for some people, while a must for others.

PR #785, opened 2015-09-14
Fix Bandcamp various artist scrobbling

tdhooper:
The Last.fm API no longer trims artist and track names, so ensure
they're always trimmed in the Bandcamp connector.


inverse:
oOo a fix from a last.fm staff member :smile:  Nice :+1:

Issue #650, opened 2015-06-26
New Icon suggestions


Paszt:
Here are some flat icon suggestion I was playing around with.  The current one is a bit smaller than all my other extension icons.

![chrome-web-scrobbler_128-icons](https://cloud.githubusercontent.com/assets/1742693/8368842/d78989a4-1b7d-11e5-8b37-35775f81ca01.png)

Let me know if there's interest.

inverse:
These are yummy!

janbaykara:
I quite like the simplicity of 2 and 5, although using an unmodified version of the last.fm logo might miscommunicate what the scrobbler is, and who it's made by.

I like the concept of 7, though I think it may not work well at small icon sizes. Could we try that without the blue square in the centre, @Paszt ?

Paszt:
![chrome-web-scrobbler_128-icons](https://cloud.githubusercontent.com/assets/1742693/8501851/7e8cf002-2177-11e5-8552-282f7c3f18fd.png)?

lacivert:
vote for 6 and/or 7

inverse:
I really like 2 and my partner likes 5.

NAUSHNIK52:
2 or 7

david-sabata:
Nice work :)

I'd prefer flat design. Also I'd like to keep the icon informative. Basically this should be the "startup" version, where no music is playing, but there should also be versions for "some music is playing" and "unrecognized track".

I probably like 7 the best, but I'm not sure how it would look in the address bar. The decorative border may take up some useful space.

What do you think?

Paszt:
I meant those above as the large 128px icon that's named icon128.png, though I did mess around with the smaller flat icons too.  Here's they are:  
    
![web-scrobbler_icons](https://cloud.githubusercontent.com/assets/1742693/8768955/f1ab0512-2e5e-11e5-8a68-f4cbc716587f.png)
    
Unfortunately the 7 from above doesn't scale down very well as you suggested.

inverse:
Awesome work @Paszt I'm really digging number 1 and 2, although 2 is still my favourite! :+1:

david-sabata:
My choices would be
* a3
* b - nothing; the cross icon turned out to be historical
* c2
* e - some flat tick on green background in the spirit of the row number 2
* g2, but maybe with white text color?

Also I'm considering removing states "d" and "f". I'd use the logo on gray (or maybe black?) background for scrobbling disabled by user.

What do you think?

[...]

Issue #197, opened 2013-06-27
Introduce new icon for validation-in-process


david-sabata:
So we know that a connector is working fine and it's just validation taking so long

janbaykara:
@david-sabata Were you thinking a 'scanning/searching...' kind of icon? 

E.g. ![](http://www.b2bsymposium.com/assets/Images/product-images/icons/binoculars-icon.png)

david-sabata:
@janbaykara not bad ... but it should be in the same style as whatever results from #650

janbaykara:
@david-sabata purely a google search for illustration. I completely agree.

PR #790, opened 2015-09-18
Flat Icons

Paszt:
Changed icons to flat style.  Added new "loading" icon to signify that
song info is being loaded from last.fm. Ref #650, #197.


inverse:
I dig these changes! :+1: looks good!

core/background/controller.js
```diff
@@ -294,6 +297,7 @@ define([
 
 				// re-send song to pipeline
 				if (data.artist || data.track) {
+					pageAction.setSongLoading(currentSong);
```
inverse: :+1: nice!

david-sabata:
Awesome, thanks! :+1:

inverse:
Damn, looks good!

PR #794, opened 2015-09-19
Require artist & track

Paszt:
don't process song if it doesn't contain both the artist and track.

### DIFF
--- a/core/background/controller.js
+++ b/core/background/controller.js
@@ -39,8 +39,8 @@ define([
 		 */
 		this.onStateChanged = function(newState) {
 
-			// empty state has same semantics as reset; even if isPlaying, we have no data to use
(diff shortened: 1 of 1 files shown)


core/background/controller.js
```diff
@@ -39,8 +39,8 @@ define([
 		 */
 		this.onStateChanged = function(newState) {
 
-			// empty state has same semantics as reset; even if isPlaying, we have no data to use
-			var isEmptyState = (!newState.artist && !newState.track && !newState.uniqueID && !newState.duration);
+			// empty state has same semantics as reset; even if isPlaying, we don't have enough data to use
+			var isEmptyState = (!(newState.artist && newState.track) && !newState.uniqueID && !newState.duration);
```
inverse: :+1: good catch

inverse:
:shipit:

PR #798, opened 2015-09-26
AOL Radio

artlung:
(aolradio.slacker.com updated and this change fixes scrobbling with the new DOM


artlung:
@david-sabata is this scheduled to get merged at some point?

PR #809, opened 2015-10-14
Fix scrobbling for V1 connectors

fedorg:
Workaround for missing song URLs and connector data.
BTW V1 are truly broken in 1.53


janbaykara:
That... makes sense... Fair enough.

Issue #801, opened 2015-10-07
Support for Cubic.fm


lef-est:
Will you please add support for Cubic.fm?
It's an aggregator for Soundcloud, Beatport, Youtube etc. (https://cubic.fm/integrations)

inverse:
I've requested an invite, will see what I can do when I get it. 

.. .alternatively if you're feeling adventurous you could take a look at one of the v2 connectors and adapt it to the site :wink:

lef-est:
Oops sorry this is my invite code!
https://cubic.fm/?ref=producthunt&i=H1i4Fv72Gg sorry forgot about that.

inverse:
:+1: cheers! Ill work on this for you

lef-est:
Thank you! :)

PR #814, opened 2015-10-18
Add cublicfm support

inverse:
- Add cublicfm support
- Tidy up connectors.js

What would be a better way to check to see if the app has booted? At the moment I just setup a MutationObserver to watch for the injected code which works. 

refs #801


david-sabata:
I prefer MutationObservers too, so this is totally ok ;-)

PR #826, opened 2015-10-28
Add cache artist track

leeyikjiun:
Use case:
Whenever the artist and track cannot be parsed, we can fix that by scrobbling manually. However, that cleaned artist and track is lost when we re-visit the page. This fixes the problem by saving user input for the song on chrome's local storage, and then loading it when the same song is re-visited later.

Technical issue:
After loading from the local storage, I wanted to make it seem as though it was a manual scrobble by calling setUserSongData. However, it does not seem to be working and I appreciate any inputs on getting it to work.

Right now, as a temporary fix, I'm just calling onProcessed directly. It works, but the notification, now playing, and album art won't be updated.

### DIFF
--- a/core/background/controller.js
+++ b/core/background/controller.js
@@ -205,7 +205,22 @@ define([
 					pageAction.setSiteSupported();
 				}
 			} else {
-				pageAction.setSongNotRecognized();
+				// try to load a processed artist track from local storage
+				var key = song.parsed.uniqueID;
+				chrome.storage.local.get(key, function(data) {
+					var processed = data[key];
+					if (processed) {
+						// setUserSongData(processed);
+						// alert("doesn't reach here");
+
+						// temp fix
+						song.processed = processed;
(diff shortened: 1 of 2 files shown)


david-sabata:
Hi @leeyikjiun, apologize for the long silence. While this probably works, it looks a little hacky.
There should be a separate pipeline stage, something like `LocalCache`, placed right after the `UserInput` stage. It will check if the unique ID is filled, present in the local storage and if so, fill the data into song structure.
Also please note that we don't access Chrome APIs directly. For local storage there's already a wrapper that supports namespaces, see `core/background/chromeStorage.js` and its usages

inverse:
Having this cached and remembered would be super cool. @leeyikjiun did you make any further progress?

leeyikjiun:
@inverse Have not made any further progress.

alexesprit:
Implemented in #1167.

PR #855, opened 2015-12-13
Ego.fm connector -

signaleleven:
Warning: as the player runs on a different domain and the playing information is displayed reliably (and with ajax update) only on www.egofm.de, this version scrobbles when the homepage is open, regardless of the status of the external player. For me it's kind of a feature, because sometimes I listen to this radio on plain old FM :)

If this is not acceptable, please reject this PR and advise on how to read the content of an iframe running a different domain (prevented by anti XSS measures in the browser)

### DIFF
--- a/connectors/v2/ego-fm.js
+++ b/connectors/v2/ego-fm.js
@@ -1,0 +1,24 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '#playerInfo';
+
+Connector.artistSelector = '#current > span.artist';
+
+Connector.trackSelector = '#current > span.song';
+
+Connector.isPlaying = function () {
+	var spanelement = document.querySelector('.now');
+	if(spanelement !== null)
+	{
+		if( spanelement.innerHTML == 'Jetzt:')
+		{
+			return true;
+		}
+	}
+	else
+	{
+		return false;
+	}
+};
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -860,6 +860,12 @@ define(function() {
 			js: ['connectors/v2/dreamfm.js'],
 			version: 2
 		},
+		{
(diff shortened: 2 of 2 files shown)


connectors/v2/ego-fm.js
```diff
@@ -0,0 +1,26 @@
...
+
+/* global Connector */
+
+Connector.playerSelector = '#playerInfo';
+
+Connector.artistSelector = '#current > span.artist';
+
+Connector.trackSelector = '#current > span.song';
+
+Connector.isPlaying = function () {
```
inverse: You could also use the `Connector.playButtonSelector` setting this to `#play`. Have a look at: https://github.com/david-sabata/web-scrobbler/blob/master/core/content/connector.js#L77

inverse:
The problem you can view with: 

https://travis-ci.org/david-sabata/web-scrobbler/builds/96563352

We have outlined that we should be using tabs and not spaces for indentation. Also there are a few errors regarding trailing spaces.

connectors/v2/ego-fm.js
signaleleven: Unfortunately clicking the play button opens a new window, but does not change anything else in the page...

inverse: Isn't that the ID for the play button on the window popup though?

signaleleven: The new window that opens is in a different domain, and I am not running the connector code there.
I am running it on the radio homepage (www.egofm.de) because that's where the current playing track information is. 
In the other window, the information is shown within a auto-refreshing iframe. 

Have a look if you want, www.egofm.de

inverse: Ahhh my bad - Didn't see that :+1:

PR #868, opened 2015-12-22
Rework PLEX connector in v2 style

mdomi:


Paszt:
There's no chance of jQuery being clobbered by another extension or page script because, from [Content Script Docs](https://developer.chrome.com/extensions/content_scripts#execution-environment): 

>the content scripts execute in a special environment called an isolated world. They have access to the DOM of the page they are injected into, but not to any JavaScript variables or functions created by the page. It looks to each content script as if there is no other JavaScript executing on the page it is running on. The same is true in reverse: JavaScript running on the page cannot call any functions or access any variables defined by content scripts.  
  
Also version 1 connectors are being phased out so this should be written as V2.  See [here](https://github.com/david-sabata/web-scrobbler/tree/master/connectors/v2) for some examples.

mdomi:
@Paszt thanks for the information, I'm a little new to the Chrome extension world. Would it be better for me to rework the Plex connector as a v2 connector?

Issue #727, opened 2015-08-27
Jamendo


FraYoshi:
![jamendo issue](https://cloud.githubusercontent.com/assets/14005514/9532398/c3e7204a-4d0b-11e5-987b-91e8be5cd85d.PNG)
Music is playing but not on scrobbling.. since all day long

signaleleven:
Jamendo had, since this issue was opened, a substantial overhaul of the interface. The connector must be updated.

alexesprit:
See #873. Works fine for me.

PR #873, opened 2016-01-03
Fix Jamendo connector

alexesprit:
Solves #727.

Hm, why base implementation of `getTrack/getArtist` function doesn't provide text trimming?


david-sabata:
@alexesprit it could, feel free to add it ;-)

PR #887, opened 2016-01-11
iHeartRadio v2 connector

ecogan:
Removed legacy connector and added v2 connector for iHeartRadio


Paszt:
Looking good, but there's a few improvements that should be made.

If the user switches to fullscreen (by clicking the arrow button in the bottom right corner) the elements in the bar along the bottom change. The track title disappears and the artist changes to the station description. For the getTrackArt, getArtist, getTrack, & getUniqueID functions you can include the full screen elements in the OR statements, putting them first. You also need to change the playerSelector property to accomodate the fullscreen layout.

You don't need both playerSelector and isPlaying so you can get rid of one.

A thank you message is displayed occasionally, you can use the isStateChangeAllowed function to ignore state changes when that message is displayed like this:

    	return !Connector.getTrack().startsWith('Thanks for listening to');
    }

They resize the track art using a query string parameter so you can get a bigger image by removing the query string from the image src:  

    src.split('?')[0]


where src is the parsed image src.

getCurrentTime is only used to detect replay.  Since this website doesn't allow the user to control song position it's not needed.

Issue #875, opened 2016-01-05
Youradio.cz and Czech radio stations


cermakjn:
Hi David,

I am from Lagardere Active. In the Czech Republic we provide music service Youradio and radio stations Evropa 2, Frekvence 1, Bonton, Dance and Zet.

We would like to implement our services to web-scrobbler. Could you please contact us directly: podpora@youradio.cz?

Thanks, Jan

PR #888, opened 2016-01-12
Connectors implementation

cermakjn:
Add Youradio.cz, Evropa2.cz, Europa2.sk, Frekvence1.cz, Radiobonton.cz and Danceradio.cz connectors.

Via issue #875


david-sabata:
I have to ask - what is the purpose of those copyright notes?

I'm afraid this would prevent anyone else to make changes to the code you contribute and that is something I definitely want to avoid.

cermakjn:
Hi David, there is internal rule to put copyright to our "public" scripts. 

For this situation, copyright is not necessary and could be removed, if you want.

david-sabata:
Would be better, thanks

cermakjn:
Copyrights has been removed.

PR #894, opened 2016-02-08
Add support for subphonic plugin in an owncloud installation

mschaaf:


connectors/v2/subphonic.js
```diff
@@ -0,0 +1,30 @@
...
+
+Connector.getAlbum = function() {
+	return $('#songdetails_artist').text().split(' - ')[1] || null;
+}
+
+Connector.currentTimeSelector = '#played';
+
+Connector.durationSelector = '#duration';
+
+Connector.playButtonSelector = '#PlayTrack'
```
Paszt: If you overwrite the isPlaying function, playButtonSelector isn't used. You can remove this line.
https://github.com/david-sabata/web-scrobbler/blob/master/core/content/connector.js#L221

mschaaf: is removed

PR #900, opened 2016-02-19
Notify of authentication problems

alexesprit:
`tradeTokenForSession` function callback wasn't called on token request error, and user wasn't noticed of authentication errors on startup. The notification was shown only one time after extenstion installation.

It fixes notifications supposed to be shown at extension startup. But if something goes wrong with authentication while using extension, user won't be noticed about that. It's possible to implement notification on 'now playing' and 'scrobble' auth errors, but I can't find a proper way to notify user. I have 2 options:
1. Show same notifications as ones are shown at startup. It might be annoying (so we can show one notification per user session).
2. Show an auth error icon in omnibar. It's good to attract user attention, but it's need to implement navigation to LastFM auth page by clicking on the icon. I have no idea how to do that.

### DIFF
--- a/core/background/services/lastfm.js
+++ b/core/background/services/lastfm.js
@@ -143,6 +143,7 @@ define([
 			})
 			.fail(function(jqxhr, textStatus, error) {
 				console.error('auth.getSession failed: ' + error + ', ' + textStatus);
+				cb(null);
 			});
 	}
 


david-sabata:
Good catch!

I agree with changing the icon. The implementation shouldn't be that hard. See `pageAction.js`, where icons are defined. You basically need to add another function like `setAuthFailed` which would set up the right icon and popup. The popup may then contain the message about the need to reauthenticate and maybe a link to settings page. I wouldn't authenticate users directly from the popup as it is unloaded and destroyed once it loses focus.

Paszt:
I can create a new svg icon if needed.Maybe something like this: 
page_action_error
or some other variation.

Issue #911, opened 2016-03-04
Add support for Amazon echo


puthali:
Currently playing songs on Amazon echo device are available at following URL - http://alexa.amazon.com/#player 

Using that and this Chrome extension one can easily scrobble all the songs playing on the Echo device.
 
I added support for it, creating this issue to track the pull request.

PR #912, opened 2016-03-04
Add support to scrobble from Amazon Echo

puthali:
This adds the basic support for #911 scrobbling from Amazon Echo. I tested with Live Radio from iHeartRadio, as well as stations & playing songs from your library. 

I changed the starter.js to support a delayLoad if specified. I think it's a good approach but I will be happy to change that if there are better suggestions on how to solve it. 

This is my first attempt from the brief look at code I had. So if I'm missing anything do let me know.  Btw I was really impressed & happy about how easy it was to add support!

### DIFF
--- a/connectors/v2/alexa.js
+++ b/connectors/v2/alexa.js
@@ -25,41 +25,41 @@
 })();
 
 var isPlayingLiveRadio = function() {
-  if($('#d-secondary-control-left .disabled').size() == 1 && $('#d-secondary-control-right .disabled').size() == 1) {
-    return 'true';
-  } else {
-    return 'false';
-  }
-}
+	if ($('#d-secondary-control-left .disabled').size() == 1 && $('#d-secondary-control-right .disabled').size() == 1) {
+		return 'true';
+	} else {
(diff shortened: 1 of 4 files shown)


alexesprit:
Why don't you use MutationObservers? You can use cubic.fm and kollekt.fm connectors as examples. In addition, you can take a look at my changes of Yandex radio connector in #903.

Delays are bad because the player element can be loaded later than your 1000 ms.

puthali:
I did give MutationObservers a try looking at cubic.fm but it didn't work, which is why I went with the delay. Looked like it needed everything to be handled manually in the connector? I'll look at the other two examples you mentioned and give it another go over the weekend.

puthali:
Fixed to use the MutationObserver instead of the delay.

alexesprit:
:+1: Good! And please fix linter errors to make Travis CI happy. You can see them by clicking on [Details](https://travis-ci.org/david-sabata/web-scrobbler/builds/113987863) link or by executing `grunt test` on your PC.

PR #913, opened 2016-03-06
Check for "Full Album" string in youtube parser

Mewster:


Mewster:
I think that the string "Full Album" can be removed from the string

PR #934, opened 2016-03-29
Add support to scrobble from Youtubify

Sacret:
Connector for http://youtubify.vebto.com/

### DIFF
--- a/connectors/v2/danceradiocz.js
+++ b/connectors/v2/danceradiocz.js
@@ -1,0 +1,11 @@
+'use strict';
+/* global Connector */
+Connector.playerSelector = 'div#root>div#player';
+
+Connector.artistSelector = '#artist';
+Connector.trackSelector = '#title';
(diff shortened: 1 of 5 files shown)


Paszt:
Testing this I've found that pausing the songs causes a reset since you're using the track.playing class. Also, you only need either playButtonSelector or isPlaying, not both. I recommend the following for youtubify.js:

	'use strict';

	/* global Connector */

	Connector.playerSelector = '#player-controls';

	Connector.artistSelector = '.current-track .info .artist-name';

	Connector.trackSelector = '.current-track .info .track-name';

	Connector.trackArtImageSelector = '.current-track img';

	Connector.playButtonSelector = '#player-controls .icon-play';

	Connector.durationSelector = '.track-length';

	Connector.currentTimeSelector = '.elapsed-time';

Sacret:
Thank you, @Paszt!

I've made changes according your suggestions.

PR #946, opened 2016-04-11
vk.com beta design support (new.vk.com)

umountdevbrain:


alexesprit:
It's better to write connectors using new v2 format. You wrote this connector in v1 format which is deprecated and not recommended to use.
There are a lot of examples of v2 connectors in `connectors\v2` directory; the documentation is placed in `core\content\connector.js` file.

PR #949, opened 2016-04-14
Amazon new URL

### LINKED ISSUE #948: Amazon Music in Chrome Not Working
The scrobbler is stopped working when streaming via Amazon Music while using it in the Chrome browser.  It was sketchy the past few weeks (having to choose an artist and then refresh the page to get the scrobbler to recognize it), but it seems like it is completely gone - the icon is gray, and does not even say that the site is enabled for scrobbling.

Paszt:
geolocation doesn't allow me to test all countries so I'm just guessing that they'll use the same sub domain in each country.

Resolves #948


TheDudeWithTheThing:
using these changes in unpacked extension mode and works perfectly, thanks @Paszt ❗ 

I'm in the US and using music.amazon.com as the domain

PR #978, opened 2016-05-31
Fix #935 add bandcamp

### LINKED ISSUE #935: Song titles aren't read when playing from Bandcamp user pages
When playing songs from one's own collection or another user page, the plugin icon shows a question mark and has a wrong title in the text field.

Example URL: https://bandcamp.com/donnerpartyofone. Then click play on any of the album covers.

There is a special case for songs which are released as stand-alone releases, such as "Den Lille Piges Død" by Myrkur, which might have to be treated differently when played (the HTML does not have the `<a class="fav-track-link" ...>` which denotes the song title for other albums).

galeksandrp:
#935

### DIFF
--- a/connectors/v2/bandcamp.js
+++ b/connectors/v2/bandcamp.js
@@ -35,7 +35,7 @@ function getArtist() {
 function getTrack() {
 	var track = $('.track_info .title').first().text() ||
 				$('.trackTitle').first().text() ||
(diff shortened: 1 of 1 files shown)


inverse:
This is already done https://github.com/david-sabata/web-scrobbler/pull/937

PR #987, opened 2016-06-15
Add vevo connector (#674)

### LINKED ISSUE #674: connector request: vevo.com
(They had their own scrobbling support in [2010](http://www.last.fm/forum/34905/_/647424#f13467332): 
vevoacc3221 1
but it was apparently dropped:
image)

Thanks either way for the great tool!!

clentner:
Add [vevo.com](http://www.vevo.com) connector (fixes #674)

### DIFF
--- a/connectors/v2/vevo.js
+++ b/connectors/v2/vevo.js
@@ -1,0 +1,11 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '#player';
+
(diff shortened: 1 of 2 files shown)


inverse:
Could we fix the conflict and then we can go with this :+1:

alexesprit:
Implemented in b26387298c7fee4e393ec2bdd5ab3ebd9aa3134a.

Issue #997, opened 2016-07-13
Notify when song can't be resolved


dufferzafar:
Sometimes the scrobbler can't detect which song is playing (any idea why this occurs?) and the icon changes to a question mark one. I then have to manually click on it, and submit the song.

I think the user should be notified that the song couldn't be detected.

ex47:
![screenshot_5](https://cloud.githubusercontent.com/assets/2480069/17136764/ff545fa4-533f-11e6-8d43-350d86983b74.png)

Looks ok?

dufferzafar:
Yeah, that looks neat!

alexesprit:
Added in #1002 with some changes.

Issue #185, opened 2013-06-16
Not valid - Adding tracks to last.fm


ghost:
Everytime I use omnibox to enter a song still not existing on last.fm, I get the "Not valid" message.

It would be great if we could scrobble songs for the first time to last.fm even when the scrobbler doesn't recognize it. 

Examples:
http://www.youtube.com/watch?v=t6MIE5hrZFk
http://www.youtube.com/watch?v=4_F7EgrmAdU

alexesprit:
Since the extension can save and load edited track info, this issue is resolved. Plus, we have an option to force scrobbling of unknown tracks.

PR #1002, opened 2016-07-26
Show notification for unrecognized song

ex47:
possible fix for #997 
- [x] remove the GA tracking for this kind of notifications, it's not intended
- [ ] users will probably expect to open the form with track info upon clicking the notification
- [x] the notification is something that should be opt-in, which means a checkbox in settings which is unchecked by default

ff545fa4-533f-11e6-8d43-350d86983b74
03a8fc64-5e2a-11e6-8891-60f178ca951f


david-sabata:
Few points to improve
* remove the GA tracking for this kind of notifications, it's not intended
* users will probably expect to open the form with track info upon clicking the notification
* the notification is something that should be opt-in, which means a checkbox in settings which is unchecked by default

ex47:
@david-sabata 
> users will probably expect to open the form with track info upon clicking the notification

Do you mean this popup?
screenshot_5

If so, i'm afraid it is not possible. [stackoverflow](https://stackoverflow.com/questions/10479679/how-can-i-open-my-extensions-pop-up-with-javascript?answertab=votes#tab-top)

alexesprit:
I have some suggestions.

Since `Unrecognized song notifications` option depends of  `Use desktop notifications` one, the first option should be disabled (be inactive for interaction) too  if the second option is unchecked:

image

The patch:
```patch
diff --git a/options/options.js b/options/options.js
index 03e36f9..3ace120 100644
--- a/options/options.js
+++ b/options/options.js
@@ -20,6 +20,7 @@ require([
 			.attr('checked', (localStorage.useNotifications == 1))
 			.click(function () {
 				localStorage.useNotifications = this.checked ? 1 : 0;
+				$('#use-unrecognized-song-notifications').attr('disabled', !this.checked);
 			});
 
 		$('#use-unrecognized-song-notifications')
@@ -108,6 +109,9 @@ require([
 		});
 
 		$('input#toggle').prop('checked', checkedState);
+
+		var showNotifications = $('#use-notifications').is(':checked');
+		$('#use-unrecognized-song-notifications').attr('disabled', !showNotifications);
 	}
 
 	function listConnectors() {
```

Add an offset of `Unrecognized song notifications` (because of its dependency):

image

I noticed that some programs use this way to show options dependency.

options.css
```css
#use-unrecognized-song-notifications {
	margin-left: 10px;
}
```

f1u77y:
> users will probably expect to open the form with track info upon clicking the notification

I also suggest to add option for forcing scrobbling(see #185); this way seems to be good for fixing that issue.

About form: pageAction does not allow to show popup programmically, so we need to use a separate form for that.

alexesprit:
I tried to use 'manual_scrobble' page as a separate window, and it works almost fine.
P.S. Opera ignores width and height params. :(

```patch
diff --git a/core/background/notifications.js b/core/background/notifications.js
index 673186d9..ddbde87a 100644
--- a/core/background/notifications.js
+++ b/core/background/notifications.js
@@ -127,7 +127,11 @@ define([
 					message: song.getArtist() + ' — ' + song.getTrack()
 				};
 
-				chrome.notifications.create('', options);
+				chrome.notifications.create('', options, (notificationId) => {
+					addOnClickedListener(notificationId, () => {
+						window.open('/popups/manual_scrobble.html', 'info', 'width=250,height=140');
+					});
+				});
 			}
 		};
 
diff --git a/popups/manual_scrobble.html b/popups/manual_scrobble.html
index 6a7ecce5..05829b69 100644
--- a/popups/manual_scrobble.html
+++ b/popups/manual_scrobble.html
@@ -2,7 +2,7 @@
 <html lang="en">
 	<head>
 		<meta charset="UTF-8">
-		<title></title>
+		<title>Submit the song info</title>
 		<script type="text/javascript" src="/vendor/jquery-2.1.0.min.js"></script>
 		<script type="text/javascript" src="manual_scrobble.js"></script>
 		<style>
```

alexesprit:
Rebased and merged into master manually.

PR #1001, opened 2016-07-26
add connector for apidog.ru

ex47:

PR #1003, opened 2016-07-27
Add TrtTurku Connector

TaeZStkyoht:
This connector scrobbles songs from http://www.trtturku.net


ex47:
I think you have to add your connector to the list of connectors `core/connectors.js`. Like here [connector for apidog.ru #1001](https://github.com/david-sabata/web-scrobbler/pull/1001/files).

TaeZStkyoht:
Thans ex47. I have been already wondering about how it finds the site link.

PR #1006, opened 2016-07-28
Fix #1004: add connector for groovemp3.com

### LINKED ISSUE #1004: Add support for groovemp3.com (new version of Grooveshark)

ex47:
#1004

Currently ignores other artists if there are more than one.

### DIFF
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -1038,6 +1038,13 @@ define(function() {
 			matches: ['*://youtubify.vebto.com/*'],
 			js: ['connectors/v2/youtubify.js'],
 			version: 2
+		},
+
+		{
+			label: 'GrooveMP3',
+			matches: ['*://groovemp3.com/*', '*://www.groovemp3.com/*'],
+			js: ['connectors/v2/groovemp3.js'],
+			version: 2
 		}
 
 	];
--- a/connectors/v2/groovemp3.js
+++ b/connectors/v2/groovemp3.js
@@ -1,0 +1,22 @@
+'use strict';
+
+/* global Connector */
+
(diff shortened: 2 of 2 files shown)


core/connectors.js
```diff
@@ -1020,6 +1020,13 @@ define(function() {
 			matches: ['*://*radiobonton.cz/*'],
 			js: ['connectors/v2/danceradiocz.js'],
 			version: 2
+		},
+
+		{
+			label: 'GrooveMP3',
+			matches: ['*://groovemp3.com/*'],
```
Paszt: _www_ subdomain is also valid, please include that: http://www.groovemp3.com/  

    matches: ['*://groovemp3.com/*', '*://www.groovemp3.com/*'],

alexesprit:
Just noticed that groovemp3 has the same layout as youtubify.

inverse:
Do you think they should just use the same connector?

alexesprit:
I think there's no reason to duplicate the same thing. But in this case the connector should have the common name for both websites (e.g. used UI template name).

alexesprit:
Looks like it's a [BeMusic music streaming engine](https://codecanyon.net/item/bemusic-music-streaming-engine/13616699). So, I think the `youtubify` connector should be renamed to `bemusic` and should be used for both websites.

ex47:
so what do we do?

alexesprit:
Well, I suggested an option already. :)

Or you can resolve conflicts and leave the connector as it is, but it should be fixed (see #1150 for details),

alexesprit:
Impelemented in 6da7be655ebcecf0b7cf7c799a70ad9e97253b37.

PR #1008, opened 2016-07-30
Moved pakartot.lt connector to API V2

2m:
Also added artist and title trimming which was missing in the old connector.


2m:
Fixed missing semicolon.

inverse:
Thanks!

connectors/v2/pakartot.js
```diff
@@ -0,0 +1,21 @@
...
+Connector.playerSelector = '#player';
+
+Connector.artistSelector = 'div.jp-player-artist > a';
+
+Connector.trackSelector = 'div.jp-player-title > a';
+
+Connector.playButtonSelector = '.jp-play';
+
+Connector.getArtist = function() {
+	var text = $(this.artistSelector).text().trim();
```
inverse: @alexesprit @david-sabata should we add trimming to the core?

alexesprit: I asked @david-sabata about adding `trim` to default `Connector.getArtist` implementation. He agreed.
I think we'll have no regressions, if we add trimming.

david-sabata: Agree, trimming the values before doing anything else should be totally fine

alexesprit: > Or move this logic: https://github.com/david-sabata/web-scrobbler/blob/master/connectors/v2/bandcamp.js#L17 to core? something like some base connector cleanup?

I think it's should be in the core if it's used by more than one connector.

What about implementation. I think that's all stuff should be placed into some `util` module, not in the `connector` module. Youtube shared metadata filter should be placed into that module too. Then, these all functions can be easily covered by unittests (the unittest core is almost done).

inverse: What do you think of an implementation like this:

https://github.com/david-sabata/web-scrobbler/compare/master...inverse:cleanText?expand=1

alexesprit: Looks nice for a start. But a lot of cleanup functions will cause 'flags hell'. I have some idea about configurable cleanups without flags in `connector` module (and it can be combined with Youtube shared metadata filter!), but I'm afraid it's will be too complicated. And I'm not sure if it is implementable since I'm not familiar with JavaScript world. I'll try to implement it laler, and we'll see how it's can be used.

P.S. I noticed we all forgot about Gitter and started to discuss things in diff comments :D

david-sabata: I'd keep only the very basic cleanup in the base connector, basically just trimming, which is useful everywhere.

I've had an idea long time ago for customizable filters on "core" level. Connectors would do the basic cleanup and send data to core, where a configurable set of filters (probably somewhere on the beginning of pipeline) would be applied. The filters could be further grouped and could be customizable connector-wide or for single specific connector.

I think moving the cleanup to core would make connectors even more simple and we'd avoid copy-pasted Youtube filters in most of them :grinning:

PR #1013, opened 2016-08-08
Hide token in 'auth.gettoken' response

alexesprit:
It's better to hide token automatically by the extension instead of explaining people how to hide it manually (https://github.com/david-sabata/web-scrobbler/issues/981#issuecomment-238158879).

### DIFF
--- a/core/background/services/lastfm.js
+++ b/core/background/services/lastfm.js
@@ -53,8 +53,6 @@ define([
 		http_request.setRequestHeader('Content-Type', 'application/xml');
 		http_request.send();
 
-		console.log('getToken response: %s', http_request.responseText);
-
 		var xmlDoc = $.parseXML(http_request.responseText);
 		var xml = $(xmlDoc);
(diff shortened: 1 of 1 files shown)


core/background/services/lastfm.js
```diff
@@ -71,6 +69,10 @@ define([
 				// set token and reset session so we will grab a new one
 				data.sessionID = null;
 				data.token = xml.find('token').text();
+
+				var response = http_request.responseText;
+				console.log('getToken response: %s', response.replace(data.token, data.token.substr(5)));
```
inverse: Why change it's position in the file?

alexesprit: To prevent logging of response twice in error is occurred.

inverse: In which case would it be better to use to use data.token as it's already parsed?

inverse:
:+1:

PR #1017, opened 2016-08-09
Add douban.fm Connector

fakelbst:


alexesprit:
If you override `Connector.isPlaying` function, there's no reason to define `Connector.playButtonSelector` field.

fakelbst:
@alexesprit All right, I'll fix it,  thanks.

connectors/v2/doubanfm.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '.fullplayer';
+
+Connector.trackArtImageSelector = '.playing-cover img';
+
+Connector.trackSelector =	'.fullplayer .titles .title a';
```
inverse: Could we fix this indentation please :+1:

fakelbst: Fixed, thanks.

fakelbst:
Douban Programme is not widely used, Douban FM is much more popular. Please merge this commit :), thanks.

alexesprit:
Hi, I just wanted to check you connector, and it doesn't work:
```
starter.js:35 Web Scrobbler: Setting up observer
starter.js:58 Web Scrobbler: Player element (.fullplayer) was not found in the page.
```
I've tried to listen to the music from this URL: https://douban.fm/songlist/429077. Scrobbling from the main page (https://douban.fm/) doesn't work too.

Is there something I'm doing wrong?

fakelbst:
@alexesprit I'm sorry, seems they do a refactory for the page. And i'm fixing my code now.
And there seems I can't get Connector.playerSelector with class in first time. The page rending dynamical, how can I set playerSelector after the page loaded?

alexesprit:
@fakelbst in this case you should use the `Two MutationObservers` pattern. You can use cubicfm connector as example of implementation of this pattern.

fakelbst:
@alexesprit Thanks very much. Fixed.😬

alexesprit:
Small improvements:
**Fix song recognition**  
```patch
@@ -26,9 +26,13 @@

 Connector.trackArtImageSelector = '.playing-cover img';

-Connector.trackSelector = '.player-wrapper .middle>div:nth-child(2) a';
+       return $('.player-wrapper .link.artist-name').first().text();
+};

-Connector.artistSelector = '.player-wrapper .link.artist-name';
+Connector.getTrack = function() {
+       return $('.player-wrapper .middle>div:nth-child(2) a').first().text();
+};

 Connector.getDuration = function () {
        var text = $('.time').text();
```
Resolves this issue:
image

**Fix playing state recognition**
```patch
@@ -36,6 +40,5 @@ Connector.getDuration = function () {
 };

-       var classString = $('.buttons .icon').eq(2).attr('class');
-       return /(^|\s)icon-pause(\s|$)/.test(classString);
+       return $('.player-wrapper .buttons .icon-pause').length > 0;
 };
```
The `isPlaying` function returns `false` to me, because the `classString` variable equals to `icon` if the song is actually playing.

Can you please check these changes and accept them if possible?

fakelbst:
@alexesprit Thanks for the improvement. Just update my codes.

PR #1024, opened 2016-08-10
Fix and improve AccuJazz connector

alexesprit:
#### Changelog:
* set proper URL;
* added duration processing;
* added album field;
* add coverart.

I replaced AccuJazz connector data in the list of connectors to make injector catch it before AccuRadio.


inverse:
:+1: Looks good to me

PR #1029, opened 2016-08-11
Delete Z-Music connector

alexesprit:
z-music.org is gone and redirects to muzebra.ru. The latter is still in development.


inverse:
:+1:

Issue #1022, opened 2016-08-10
redundant definition?


ex47:
europa 2 - [connectors.js](https://github.com/david-sabata/web-scrobbler/blob/a4944bca28edd535a16acbcbfc4c783cbff6b5be/core/connectors.js#L985-L997)
same js file

spotify - [connectors.js](https://github.com/david-sabata/web-scrobbler/blob/a4944bca28edd535a16acbcbfc4c783cbff6b5be/core/connectors.js#L237-L248)
https://player.spotify.com/ redirects to https://play.spotify.com/

alexesprit:
Yep, `spotify-play.js` is useless. Can't figure out why it's needed.

[onair.evropa2.sk](http://onair.evropa2.sk/) is unavailable for me. Maybe it's worth deleting it? Or is it unavailable just for me?

I can't listen to any music directly on [onair.evropa2.cz](http://onair.evropa2.cz/). But it has redirects to [youradio.cz](http://www.youradio.cz). Don't sure if Evropa2 connector is useful.

ex47:
> onair.evropa2.sk is unavailable for me.

http://www.downforeveryoneorjustme.com/http://onair.evropa2.sk/

> I can't listen to any music directly on onair.evropa2.cz. But it has redirects to youradio.cz. Don't sure if Evropa2 connector is useful.

works, but you have to disable ad blockers/social blockers

alexesprit:
> works, but you have to disable ad blockers/social blockers

Yes, after disabling adblock it started to work. In this case it's enough to drop onair.evropa2.sk from the list.

Paszt:
> Yep, spotify-play.js spotify.js is useless. Can't figure out why it's needed.  
When sites change urls, they often roll out the changes at different times in different locations.  At one point there was an overlap where both urls were valid.  If we can't be absolutely sure that it's not used in some local, there's really no issue in having an extra, unused domain remain.

alexesprit:
The issue should be closed. It's resolved in #1030 and #1085.

PR #1030, opened 2016-08-12
Delete v1 connectors

alexesprit:
More cleanup. All these connectors have v2 implementations which works fine ~~(except Hillydilly)~~.

**Update:**
Blinkbox Music and TTNET Muzik are no longer available. The first one was closed, the second one redirects to new site which has another layout.


ex47:
:+1:

ex47:
maybe you could also address #1022

alexesprit:
What you mean?

ex47:
Drop [this](https://github.com/david-sabata/web-scrobbler/blob/a4944bca28edd535a16acbcbfc4c783cbff6b5be/core/connectors.js#L985-L990), [this](https://github.com/david-sabata/web-scrobbler/blob/a4944bca28edd535a16acbcbfc4c783cbff6b5be/core/connectors.js#L237-L242) and this [file](https://github.com/david-sabata/web-scrobbler/blob/ff547dfdfd1b6682f61baf50ed4bdab0516106e8/connectors/v2/spotify.js). As a part of cleanup.

alexesprit:
Evropa's URLs are both valid (in #1022 I mistyped and went to wrong URL). These URLs need to be merged into one item in the connectors list. I'll make a new PR for this case.

Old Spotify connector is useless, but it's in v2 format. Will do it in another PR. In this PR I deleted connectors in v1 format.

manifest.json
```diff
@@ -15,7 +15,6 @@
    "web_accessible_resources": [
       "icon128.png",
       "connectors/jango-dom-inject.js",
```
inverse: Shouldn't this `jango` one be removed too?

inverse:
Awesome work :+1:

manifest.json
alexesprit: Yes, that need to be cleaned too.

inverse:
Thanks!

Issue #578, opened 2015-04-19
Paused music on vk.com still scrobbling


RadikMusabirov:
When i paused music on vk. com it's still scrobbling. On last.fm it's look like
![screenshot_3](https://cloud.githubusercontent.com/assets/12019447/7219753/bcff0942-e6b7-11e4-8985-def9299ce0cf.png)
![screenshot_4](https://cloud.githubusercontent.com/assets/12019447/7219754/beead1aa-e6b7-11e4-8ffb-8dfa851e1eda.png)

My profile on last.fm for details - http://www.lastfm.ru/user/M0safiR/tracks

inverse:
Could be related to #409

ex47:
This should be closed. Fixed long time ago #416

Issue #470, opened 2014-09-30
Scrobbler with vk.com  does't work


foobnix:
It works successful with Google Play Music
but with vk.com not works :(

Issue #409, opened 2014-06-20
Random and multiple scrobbles, VK.com


laiskvorst:
I use the Chrome extension and some songs tend to get scrobbled multiple times. I've noticed that it happens to songs with long titles more often than not.
Besides that, it often replaces the actual song with a random one from those scrobbled before or inserts it between other songs.

ex47:
This should be closed. Fixed long time ago #416

Issue #215, opened 2013-07-17
vk.com scrobbles only the first play


playbobrik:
Found a bug with vk.com scrobbling on Opera15. If the song is on repeat last.fm scrobbles it only on first play. Any subsequent plays become lost :( Scrobbling continues normally only after track change.

PR #1039, opened 2016-08-17
Fix VK scrobbling

### LINKED ISSUE #1038: New vk.com design fails scrobbling
Hi!  Seems like after vk.com made new design, scrobbling is not working anymore.

alexesprit:
VK has released new design of their site, so newvk.js should be mapped to 'vk.com', not 'new.vk.com'. Resolves #1038 and #1045.

And now there is no way to return back to old version of site, so old v1 connector is useless.

Also, #215, #409, #470, #578 should be closed. These issues are related to old v1 connector.


alexesprit:
Added decoding of HTML entities to prevent scrobbling of something like this:
image

And some cosmetic changes.

inverse:
That whole local storage thing seems so odd :/

ex47:
@inverse the vk uses localStorage to synchronize player status on multiple tabs resulting in DOM of the page sometimes lacking info about the song (e.g. when player is minimized)

kizhaeff:
When the update scrobbling to working VK?

kizhaeff:
When the update scrobbling to working VK?

alexesprit:
When @david-sabata will release new version of the extension.

PR #1044, opened 2016-08-21
Update for alexa.js

NotIT:
Amazon Alexa player changed, no longer has "-" between tracks in song history. Now always just return from the main player.


inverse:
We updated the core and removed the need for the mutationobserver in the connector level, could you fix the issues?

alexesprit:
Closed as outdated.

PR #1050, opened 2016-08-26
Add a v1 connector for Saavn

dufferzafar:
I pulled in the original source from [here](http://github.com/mayank127/web-scrobbler/), made some changes and tested it out.

But I can't get the connector to load.

Can someone please have a look and tell what's wrong?

### DIFF
--- a/connectors/saavn.js
+++ b/connectors/saavn.js
@@ -13,11 +13,9 @@
 var DOM_TRACK_NAME = '#player-track-name';
 var DOM_TRACK_DURATION = '#track-time';
 
-Connector.playerSelector = '#now-playing';
-
 // Prevent same track from being scrobbled continuously
 var previousTrack = '';
-var previousUrl = '';
+var previousURL = '';
 
 // Get track name
 function getTrack() {
@@ -63,9 +61,9 @@ function updateNowPlaying() {
 
         // Use jQuery to parse the HTML
         var trackDetailsDOM = $(trackDetailsHTML);
-        var artist = trackDetailsDOM.find('.page-meta-group > .meta-list').text()
+        var artist = trackDetailsDOM.find('.page-meta-group > .meta-list')[0].textContent
 
-        console.log('Validating: ' + artist + ' @#@ ' + track + ' @#@ ' + duration);
+        console.log('Validating: ' + artist + ' - ' + track + ' - ' + duration);
 
         if (!artist || !track || previousTrack == track) {
             return;
--- a/connectors/saavn.js
+++ b/connectors/saavn.js
@@ -1,0 +1,125 @@
+/**
+ * Chrome-Last.fm-Scrobbler - Saavn.com Connector
+ *
+ * Original Author: Mayank Meghwanshi [mayankmeghwanshi@gmail.com]
+ * Testing, Fixes, Refactor: Shadab Zafar [dufferzafar0@gmail.com]
+ *
+ * Saavn.com's now playing bar does not have Artist's name
+ * but there is a link to the Track's information - so we make an
+ * Ajax call and extract artist information.
+ */
+
+// DOM Nodes to keep track for song details
+var DOM_TRACK_NAME = '#player-track-name';
+var DOM_TRACK_DURATION = '#track-time';
+
+Connector.playerSelector = '#now-playing';
+
+// Prevent same track from being scrobbled continuously
+var previousTrack = '';
+var previousUrl = '';
+
+// Get track name
+function getTrack() {
+    var trackName = $(DOM_TRACK_NAME + ' a').text();
+    if (!trackName) {
+        return ;
+    }
+
+    return trackName;
+}
+
+// Get track length
+function getDuration() {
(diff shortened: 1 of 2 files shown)


alexesprit:
Is there a reason to write the connector in v1 format?

alexesprit:
> But I can't get the connector to load.
It's because of wrong URL pattern. It shold be '*://*.saavn.com/*'.

dufferzafar:
> Is there a reason to write the connector in v1 format?

Not really, no. The source I copied it from had it in v1. Once I get this to work, I'll port it to v2, I guess.

alexesprit:
Strange, I see no artist info in the player bar on the Saavn page, like in gaana.com.

connectors/saavn.js
```diff
@@ -0,0 +1,125 @@
...
+    return 60 * parseInt(duration[0]) + parseInt(duration[1]);
+}
+
+function updateNowPlaying() {
+
+    var track = getTrack();
+    var duration = getDuration();
+
+    // Extract Track's URL to find artist information
+    var trackURL = $(DOM_TRACK_NAME + ' a').attr("onclick");
```
dufferzafar: @alexesprit  Yes, there is no artist info in the now playing bar, but there is a link to track details, from which we can extract the info.

alexesprit: @dufferzafar ok, I got it. I think we could use the similar mechanism for gaana.com too.

connectors/saavn.js
```diff
@@ -0,0 +1,123 @@
...
+                });
+            }
+        });
+    });
+}
+
+console.log('Saavn.com connector loading');
+
+// Track enclosing <div>, id: 'now-playing' to observe track change
+$('#now-playing').on('DOMSubtreeModified', function () {
```
dufferzafar: The `.on` function is available in jquery > 1.7, but for v1 connectors we include jquery 1.6 

I got things to work by just using jquery 2 for this script.

But I think we should just port this connector to v2

@alexesprit Can you please have a look at tell me how can we achieve the ajax call etc. in v2 format?

connectors/saavn.js
dufferzafar: I don't think gaana.com has links for a single track, like saavn does.

connectors/saavn.js
alexesprit: @dufferzafar ok, I'll take a look at it. I have some thoughts about ajax in v2 connectors, but don't sure if it helps.

connectors/saavn.js
alexesprit: gaana.com has link to the album page, that contains now playing song info, including artist name.

connectors/saavn.js
alexesprit: > The .on function is available in jquery > 1.7, but for v1 connectors we include jquery 1.6

You could try to use MutationObserver to process DOM changes, though. That's how v2 core work (and some v1 connectors too).

But v2 format is preferred anyway.

connectors/saavn.js
dufferzafar: There's a link to the album page, yes, which includes the information but there's no indicator as to which song is now playing, so we'll have to find that from the list, right?

alexesprit: image
The `ul` element of now playing song has `currentSong` class, by which we can identify now playing song.
I also tried to play several songs on the same page, this `currentSong` class is set to proper song on every song change. Some websites are buggy and don't switch off indicators like this one, so they have a lot of marked as 'now playing' songs.

dufferzafar: Okay, but this is being dynamically added, right? We'll have to see if that `currentSong` indicator is available to us in the HTML fetched via Ajax.

alexesprit: Agree, it's need to implement AJAX fetching, and we'll see how we can use that track info.

connectors/saavn.js
inverse: V1 is definitely deprecated. I'll update the wiki

alexesprit:
I didn't test the connector too much, but it should work without problems.
```js
'use strict';

/* global Connector */

var songArtist = null;
var trackInfoUrl = null;

var asyncRequestSongInfo = function() {
    var url = $('#player-track-name a').attr('onclick').split(',')[1].slice(2, -3);
    if (trackInfoUrl != url) {
        $.ajax({url: url, context: document.body}).done(function(trackDetailsHTML) {
            var trackDetailsDOM = $(trackDetailsHTML);
            var artist = trackDetailsDOM.find('.page-meta-group > .meta-list')[0].textContent;
            songArtist = artist;
        });
        trackInfoUrl = url;
    }    
};

Connector.playerSelector = '#now-playing';

Connector.getArtist = function() {
    asyncRequestSongInfo();
    return songArtist;
};

Connector.trackSelector = '#player-track-name';

    return $('#controls #play').hasClass('hide');
};
```

alexesprit:
@dufferzafar I implemented the connector in v2 format some time ago. Can you test it?

alexesprit:
We dropped support for v1 format.

PR #1052, opened 2016-08-29
Fix Dash Radio connector

alexesprit:
Fix song recognition on startup.

### DIFF
--- a/connectors/v2/dashradio.js
+++ b/connectors/v2/dashradio.js
@@ -7,3 +7,6 @@ Connector.playButtonSelector = '#play';
 Connector.artistSelector = '#player-content > div.song-info > h3';
 Connector.trackSelector = '#player-content > div.song-info > h2';
 Connector.trackArtImageSelector = '#player-content .album-cover img';
+
+// Simulate DOM change on startup to force song recognition
+Connector.onStateChanged();


alexesprit:
Unfortunately, this hack doesn't fix the issue, because `Connector.reactorCallback` is null when `Connector.onStateChanged` is invoked in the connector body.

I think we should have some workaround for websites such as DashRadio, Novaplanet, WWOZ and some others, which don't change DOM on startup. Like this:
```diff
diff --git a/core/content/connector.js b/core/content/connector.js
index c63eeab..60f1060 100644
--- a/core/content/connector.js
+++ b/core/content/connector.js
@@ -243,6 +243,8 @@ var BaseConnector = window.BaseConnector || function () {
                        return true;
                };

+               this.forceDomChangeOnStartup = false;
+
                // --- state & api -------------------------------------------------------------------------------------------------


diff --git a/core/content/reactor.js b/core/content/reactor.js
index 9a9599b..cc60631 100644
--- a/core/content/reactor.js
+++ b/core/content/reactor.js
@@ -49,7 +49,9 @@ var Reactor = function(connector) {

        // setup listening for state changes on connector
        connector.reactorCallback = this.onStateChanged;
-
+       if (connector.forceDomChangeOnStartup) {
+               connector.onStateChanged();
+       }


        /**
```

And definetely we should get rid of things like this: [connectors/v2/evropa2cz.js#L20](https://github.com/david-sabata/web-scrobbler/blob/master/connectors/v2/evropa2cz.js#L20).

PR #1053, opened 2016-08-29
Fix and improve FluxFM connector

alexesprit:
#### Changelog:
* fixed FluxFM connector;
* added album field;
* added coverart.


inverse:
Thanks Alex!

PR #1055, opened 2016-08-31
Switch 22tracks to v2 connector

alexesprit:


connectors/v2/22tracks.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '.player';
+
+Connector.artistTrackSelector = '.player__progress__title';
+
+Connector.isPlaying = function () {
+	return $('.icon-pause').length;
```
inverse: Shouldn't this be more of a boolean expression? e.g. `$('.player).has('.icon-pause')`

alexesprit: @inverse it's should be like this:
```js
	return $('.icon-pause').length > 0;
};
```

alexesprit:
Updated `isPlaying` function.

PR #1064, opened 2016-09-07
Mail.Ru Music service support

ViRGiL175:
New V2 connector for Mail.ru Music service.

### DIFF
--- a/connectors/v2/mailrumusic.js
+++ b/connectors/v2/mailrumusic.js
@@ -17,3 +17,13 @@ Connector.getArtist = function() {
 Connector.isPlaying = function () {
 	return $('.l-music__player').hasClass('playing');
 };
+
+Connector.getTrackArt = function () {
+	var coverArtURL = $('.l-music__player__song__cover').css('background-image');
+	var regexp = /https?%3A\S*jpg/i;
+	var coverArtURLClean = regexp.exec(coverArtURL);
+	var coverArtURLstring = coverArtURLClean.toString();
+	coverArtURLstring = coverArtURLstring.replace(/%3a/gi, ':');
+	coverArtURLstring = coverArtURLstring.replace(/%2f/gi, '/');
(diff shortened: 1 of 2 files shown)


alexesprit:
Your connector has wrong playing state recognition. Web-scrobbler assumes that music is playing while it's paused on the website and vice versa. To get it worked properly use `Connector.isPlaying` function:
```js
	return $('.l-music__player').hasClass('playing');
};
```

Also, few suggestions:
* add current time/duration processing to support repeating tracks;
* add coverart. Mail.ru provides 100x100 px cover arts which are better than Last.fm's ones (64x64 px).

ViRGiL175:
Sorry but I can't reproduce wrong playing state recognition.

I compared it with Youtube. The same behaviour:

- When I start music Web-scrobbler icon changed to green play icon.
- If I pause track Last.fm still showing "Scrobbling now", icon still green.
- If I close music service page Last.fm still showing "Scrobbling now".

Suggestions are on the way :)

alexesprit:
Sorry, I didn't noticed 8431580 commit.

ViRGiL175:
I spent several hours searching for a bug, and learned a lot about CSS and JS! :D

alexesprit:
Coverart fetching doesn't work for me. The connector produces a lot of uncaught errors:
```
mailrumusic.js:25 Uncaught TypeError: Cannot read property 'toString' of null
```
Also, you can take a look at other connectors that get coverart URL from CSS property too, e.g. `spotify-play.js` or `tradiio.js`.

ViRGiL175:
Yes, I'm newbie in JS and work in Sublime Text.
Don't even know how to debug etc )

I will look into and try to move to some IDE or smth.
Just need some time.

alexesprit:
Implemented in a0759cf591fd4f369bd6bb711e430e8d29cf9298.

PR #1067, opened 2016-09-15
Youtube Topic Support

Paszt:
Add support for Youtube topics.  For example: https://www.youtube.com/watch?v=3A9AzTKKgPI


inverse:
Thanks @Paszt - this feels like a bit of an edge case, is this a common thing?

inverse:
This still valid? feels like it could do with some :heart:

alexesprit:
This info is not available for videos have long description until description is expanded. That's true for new Youtube desing.

Issue #560, opened 2015-03-23
Another Deezer issue


NAUSHNIK52:
Hello. I just want to share about Deezer issue. Scrobbler works OK, but I hate this issue.
Every little change at Deezer page starts song scrobbling over. Like, if I'm change volume from 50% to 51%, open a side menu, open another Deezer page - scrobble start over. HTML code not changes. Is it possible to fix it?
P.S. This is issue persist in all languages for me.
Sorry for my bad English, I hope you understand what I mean.

inverse:
This is really an issue. The scrobbler works by listening to DOM changes on the title tag.

Below is an example of the problem.

![bug](https://cloud.githubusercontent.com/assets/713525/6777189/6fc2a1f0-d147-11e4-9090-3a0a7fd8ccf3.gif)

Wouldn't a better solution to be to either check to see if this matched track matches what already is playing (checking duration to allow for replays of the same song).

or listening to events being triggered directly from `dzPlayer`.

NAUSHNIK52:
@inverse I am just edited the deezer.js

    else if ($('.player .control-pause').size() > 0) {
		
		if ($('.progress-time').text() == '00:00') {
		
            return {
                track: $('.player-track-title').text(),
                artist: $('.player-track-artist').text().replace($('.player-track-artist').children().first().text(), '').trim().replace(/,$/, ''),
                duration: parseDuration($('.player-progress .progress-length').text())
            }
		}	
    }

It is works. But I am just learning the coding and maybe you or someone another find a better way to fix the issue.

inverse:
I see - so your going to only return when the track time text looks reset. I'm not sure if this is the best thing to do since it may not always be correct. Also you may want to enable strict matching using triple equals (`===`).

I'm trying to modify the scrobbler to use the `dzPlayer` object and hook into the Event subscriber mechanism but I'm struggling to find out when this is bound to the page. 

@david-sabata is it bad to do something such as:

```JS
    function bindPlayer() {

        if (dzPlayer === null) {
            window.setTimeout(bindPlayer, 1000);
            return;
        }


     // Register events here

 
    }
```

david-sabata:
@inverse It aint pretty, but it's acceptable if you do this just to initially find the player :)

inverse:
should be fixed in next release

PR #1073, opened 2016-09-22
Fix Deezer connector

### LINKED ISSUE #536: problems with deezer
hi, I've got problems with scrobbling from deezer, last.fm scrobbler doesn't recognize any song/track. When used in other services (for ex. youtube) everything seems ok.

alexesprit:
Resolves #536 #541 #560.

Additional changes:
* added current time/duration processing;
* added track art.


connectors/v2/deezer.js
```diff
@@ -0,0 +1,18 @@
...
+
+/* global Connector */
+
+Connector.playerSelector = '#page_sidebar';
+
+Connector.getArtist = function() {
+	var artists = $.map($('.player-track-artist').children(), function(artist) {
+		return $.text(artist);
+	});
+	return artists.join(', ');
```
inverse: This doesn't look like it makes sense to me? Are things really that wacky on the page?

alexesprit: That's an example how an artists element looks:
```html
<h3 class="player-track-artist heading-2-sub ellipsis">
    <!-- react-text: 1666 -->from<!-- /react-text -->
    <!-- react-text: 1667 -->&nbsp;<!-- /react-text -->
    <!-- react-text: 1668 --><!-- /react-text -->
    <span class="player-track-link" data-action="link" data-href="/artist/282118">Major Lazer</span>
    <!-- react-text: 1670 -->, <!-- /react-text -->
    <span class="player-track-link" data-action="link" data-href="/artist/288166">Justin Bieber</span>
    <!-- react-text: 1672 -->, <!-- /react-text -->
    <span class="player-track-link" data-action="link" data-href="/artist/4333887">MO</span>
</h3>
```
If `player-track-artist` is used as selector, the actual artist is 'from Major Lazer, Justin Bieber, MO'. Of course, artists can be extracted by splitting the whole string by non-breaking space and taking the second part of the string. But I think it's not a good way. So I decided to get children of `.player-track-artist`, convert them to an array of strings and join by comma.

PR #1081, opened 2016-10-03
Listen to events generated by VK player

f1u77y:
`window.ap` send player events in vk, so we must inject a script that
will listen to events from this object


alexesprit:
Why do we need to listen to this event? I think it's easy enough to watch DOM changes.

f1u77y:
@alexesprit we are already listening events generated by player node but it does not generate any events on "progress", so we have to either listen to `window.ap` or emulate "progress" by executing timer. I think first is better

alexesprit:
Yes, you're right, events are better than timers.

But since we use VK events to watch playback state, watching DOM changes is useless. So, `Connector.playerSelector` should be undefined.

f1u77y:
@alexesprit got it, will update pr asap

alexesprit:
What about removing `updateTrackInfo` calls in every metadata getter and placing it before `Connector.onStateChanged`?

inverse:
@f1u77y could you sort out the conflicts with the manifest please?

f1u77y:
@inverse done

inverse:
That was quick, cheers!

Issue #1082, opened 2016-10-06
Connetor for driveplayer.com


saintego:
It would be great to support driveplayer.com. It's google drive music player which support id3 tags.

saintego:
Hi
I cannot create pull request. This worked for me. Can you accpet it? 
Thanks
[driveplayer.com.zip](https://github.com/david-sabata/web-scrobbler/files/516358/driveplayer.com.zip)

saintego:
@nicolasgarnier

alexesprit:
Should be closed

PR #1083, opened 2016-10-07
Connetor for driveplayer.com #1082

saintego:
Connetor for driveplayer.com

### DIFF
--- a/connectors/v2/driveplayer.com.js
+++ b/connectors/v2/driveplayer.com.js
@@ -13,9 +13,7 @@ Connector.getTrack = function () {
 	return text || null;
 };
 
-Connector.isPlaying = function () {
-	return $('.jp-play').css('display') === 'none';
-};
+Connector.playButtonSelector = '.jp-play';
 
 Connector.getCurrentTime = function() {
 	var currentTimeStr = $('.jp-current-time').text().trim();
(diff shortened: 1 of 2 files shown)


connectors/v2/driveplayer.com.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+/* global Connector */
+Connector.playerSelector = '.jp-play';
```
alexesprit: `Connector.playerSelector` should be defined as `.jp-audio`. Now it's just a play button selector.

connectors/v2/driveplayer.com.js
```diff
@@ -0,0 +1,28 @@
...
+	var text = $('.playing > .artist').text().trim();
+	return text || null;
+};
+
+Connector.getTrack = function () {
+	var text = $('.playing > .title').text().trim();
+	return text || null;
+};
+
+Connector.isPlaying = function () {
```
alexesprit: This `isPlaying` function can be replaced to `Connector.playButtonSelector = '.jp-play';` definition. But the function is correct too, though.

saintego: Thanks. I was inspired by bandzone.cz.js so I made it in their way

alexesprit:
Looks nice now! :+1:

PR #1092, opened 2016-10-11
Fix WDVX connector

alexesprit:


connectors/v2/wdvx.js
```diff
@@ -2,16 +2,12 @@
...
 
-Connector.getTrack = function () {
-	return $('.title').text().replace('Title: ', '');
-};
+Connector.trackSelector = '#now-playing .title';
 
-Connector.isPlaying = function () {
-	return $('.pauseW').is(':visible');
+Connector.isPlaying = function() {
+	return $('.audiojs').hasClass('playingW');
```
inverse: Is there really a class called `playingW` looks like a typeo ;D

alexesprit: No, it's not a typo:
image

PR #1093, opened 2016-10-12
Connector for genie.co.kr

nerogit:
http://genie.co.kr is one of famous Korean music streaming service

### DIFF
--- a/connectors/v2/genie.js
+++ b/connectors/v2/genie.js
@@ -3,24 +3,11 @@
 /* global Connector */
 
 Connector.playerSelector = '#music-player';
-
-Connector.getArtist = function() {
-	return $('#ArtistNameArea').text();
-};
-
-Connector.getTrack = function() {
-	return $('#SongTitleArea').text();
-};
-
-Connector.getDuration = function() {
-	var playTime = $('#playTime strong').text();
-	return Connector.stringToSeconds(playTime.trim());
-};
-
-Connector.getCurrentTime = function() {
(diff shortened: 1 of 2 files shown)


nerogit:
Thanks for advice:) and I modified it

connectors/v2/genie.js
```diff
@@ -0,0 +1,19 @@
...
+
+Connector.playerSelector = '#music-player';
+Connector.artistSelector = '#ArtistNameArea';
+Connector.trackSelector = '#SongTitleArea';
+Connector.albumSelector = '#AlbumTilteArea';
+Connector.currentTimeSelector = '#playTime span';
+Connector.durationSelector = '#playTime strong';
+
+Connector.isPlaying = function() {
+	var btn = $('#PlayBtnArea')[0];
```
inverse: I would simplify this to 

```
	return $('#PlayBtnArea').hasClass('active');
};
```

Or is there really more than one instance of `#PlayBtnArea`?

inverse: @alexesprit I hadn't visited the site - but I was more commenting on the use of jquery to simplify the function.

alexesprit: @inverse got it.

nerogit: It's just for using `classList` property. I changed it to use `hasClass` method. Thanks @inverse

alexesprit:
Works fine! 👍

inverse:
Thanks for doing this :+1:

PR #1095, opened 2016-10-18
Update dev dependencies

alexesprit:
#### Changelog:
* updated dev dependencies;
* added devDependencies badge;
* fixed linter errors.

Since updated linters produces errors I've also included fixes of those errors:
* 'Incorrect end of line character(s) found' error (lintspaces);
* 'Rule doesn't have all its properties in alphabetical order' error (csslint).


inverse:
Nice change 👍

PR #1100, opened 2016-10-20
Fix NRK radio

alexesprit:
Switch the connector to v2 format.


connectors/v2/nrkradio.js
```diff
@@ -0,0 +1,46 @@
...
+	if (separator !== null) {
+		artist = text.substr(0, separator.index).trim();
+		track = text.substr(separator.index + separator.length).trim();
+	}
+
+	return {artist: artist, track: track};
+};
+
+Connector.isPlaying = function () {
+	return true;
```
inverse: This doesn't seem right.

alexesprit: There's no way to check if song is paused, so it's better to assume it's always playing.
Although, the `Connector.isPlaying` function implementation can be removed, because core supposes the song is playing by default.

inverse:
So now the connector will be in continuous play mode once it's detected the 'player'?

alexesprit:
Yes, you're right.

hbhc:
Can someone in here make radio.nrk.no scrobble again? 
It seems to have stopped working.

PR #1105, opened 2016-10-22
Move the 'Two MutationObservers' pattern logic to core

alexesprit:
These changes only affects on websites which have 'delayed' player element.

I didn't remove MutationObservers usage from some connectors since their logic is more complex than in the original implementation.

### DIFF
--- a/connectors/v2/alexa.js
+++ b/connectors/v2/alexa.js
@@ -2,27 +2,7 @@
 
 /* global Connector */
 
-(function() {
-		var playerObserver = new MutationObserver(function() {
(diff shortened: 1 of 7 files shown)


inverse:
I think this is a nice approach @david-sabata do you have any input? 

What are the drawbacks of globally enabling this for all connectors? Should we have this as a connector level toggle? or just go this ay?

alexesprit:
Updated recently added connectors.

david-sabata:
I think the two-observer approach has worked fine so far, so I say it's a good idea to have it in core :+1:

Issue #1111, opened 2016-10-31
Repeating the notification of the same song


hilgi:
For some reason today, when a song stats and I receive the notification on my desktop, it keeps opening and closing the notification, I turned off notifications but I notice the Chrome widget if opening and closing constantly. You can see it in the vid, the notifications were doing the same. 

I see them in the last.fm library but after they are done, they look to be removed. I am playing the web player for amazon music. 

![video0029](https://cloud.githubusercontent.com/assets/16419101/19867980/4af776b6-9f63-11e6-9d7e-1e55ac7a6967.gif)

.

alexesprit:
Is this happening only with Amazon? Can you please scrobble the music from other websites (e.g. Youtube)?

It's also would be super cool if you provide the extension logs. You can use [this howto](https://github.com/david-sabata/web-scrobbler/wiki/How-to-debug-the-extension) to get proper logs. In this case it's enough to reproduce the extension icon blinking, collect logs and paste them to some pastebin service.

hilgi:
It did not do it with Youtube. 

I saved the log, where do I add it?

alexesprit:
You can use http://pastebin.com/index.php, for example, to paste all your logs and send the paste URL to this thread.

If you are not familiar with pastebin services, you still can provide logs by sending them in comment. But Github doesn't collapse long text, so pastebin is preferred.

hilgi:
Sorry for my lack of knowledge but how do I copy the log to send it to paste bin? Thanks,

hilgi:
I have the log, I just don't see how to copy and paste it.

alexesprit:
The log is just a text, select it all, then right-click and select "Copy". Now you can paste it here or on pastebin by right-clicking and choosing "Paste".

hilgi:
Is this it? http://pastebin.com/m2Vi7QfT

alexesprit:
Yes, it is.

But you said the issue is happening when you listen to music on Amazon, right? I see you listened to music on Youtube in the log. Is the extension icon blinking too in this case?

alexesprit:
Seems like I found a way to reproduce the issue.

hilgi:
The blinking is only Amazon, I not YouTube.

alexesprit:
Ok, I got it.

I've made an attempt to fix the issue. Can you please ckeck if it's resolved? https://github.com/alexesprit/web-scrobbler/archive/fix-new-state.zip
[...]

PR #1112, opened 2016-11-02
Fix new state recognition

alexesprit:
The issue is descripted in #1111.

How to reproduce:
1. Find any connector which grabs album from website.
2. Add `getAlbum` function that returns empty string (`""`) in this connector.
3. Try to scrobble the music from website.

The issue is caused by connectors which have overriden `getAlbum` function that can return empty string.
In this case the check in `Controller.onStateChanged` function is failed:
```js
var hasSongChanged = (currentSong === null || newState.artist !== currentSong.parsed.artist || newState.track !== currentSong.parsed.track ||
newState.album !== currentSong.parsed.album || newState.uniqueID !== currentSong.parsed.uniqueID);
```
Let's take a look at `newState.album` and `currentSong.parsed.album` values:
* `newState.album` is `""` (empty string);
* `currentSong.parsed.album` is `null` (it becomes null in `Song` object constructor).

So, on every DOM change `hasSongChanged` will be always equal `true` until `Connector.getAlbum !== ""`.
Instead of adding checks like `text || null` in return statement in every overriden getter function I decided to place these checks in to core.

### DIFF
--- a/core/content/connector.js
+++ b/core/content/connector.js
@@ -114,8 +114,7 @@ var BaseConnector = window.BaseConnector || function () {
 		 * @returns {string|null}
 		 */
 		this.getArtist = function () {
-			var text = $(this.artistSelector).text();
-			return text || null;
+			return $(this.artistSelector).text();
 		};
 
 		/**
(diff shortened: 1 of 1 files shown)


inverse:
Interesting research and good write up of the problem. @david-sabata what do you think of the proposed solution?

david-sabata:
Looks good and makes the base connector more robust, so :+1: from me

PR #1113, opened 2016-11-03
Add playmoss

2m:


inverse:
Thanks for contributing - could you fix the grunt issues then we should be good to go!

2m:
Fixed the spaces/tabs grunt issue.

f1u77y:
you can just define `Connector.artistTrackSelector` since artist and track are in the same element

inverse:
I'd have to agree with @f1u77y on that one. Would make things simpler

2m:
I have made the changes requested.

connectors/v2/playmoss.js
```diff
@@ -0,0 +1,19 @@
...
+
+/* global Connector */
+
+Connector.playerSelector = '#footer-player';
+
+Connector.artistTrackSelector = '#footer-player li.playing a';
+
+Connector.playButtonSelector = 'button.play';
+
+Connector.getArtist = function() {
```
inverse: You can remove this function

connectors/v2/playmoss.js
```diff
@@ -0,0 +1,19 @@
...
+	var text = $(this.artistTrackSelector).text().split(' - ')[0].trim();
+	return text || null;
+};
+
+Connector.getTrack = function() {
```
inverse: This is no longer required either since the work is done in the BaseConnector via the artistTrackSelector property

2m:
Ahh thanks. Did not know that there is split by hyphen functionality already build in.

inverse:
Super :+1: Thanks for the commit :)

PR #1121, opened 2016-11-05
Fix reading options

alexesprit:
Was broken in #1120.

All values in localStorage are stored as strings.


inverse:
Ahh good catch - seems rather limiting to only allow string values :/

PR #1122, opened 2016-11-05
Delete useless connectors

alexesprit:
* `music.sonyentertainmentnetwork.com` redirects to 404 page;
* `www.slacker.com/webplayer/index_embed.vm` is just a page with picture;
* `thedrop.club` is not available;
* `shuffler.fm` is not available too.


alexesprit:
I'd like to remove fizy connector too. They provide 30-sec tracks, so the connector is useless because minimum scrobble time is 30 seconds.

`weborama.fm` seems like not available too.

inverse:
I'd say get rid of the fizy one and weborama :) Awesome job!

PR #1125, opened 2016-11-05
Switch Sullen-Ural connector to v2 format

alexesprit:


connectors/v2/sullen-ural.js
```diff
@@ -0,0 +1,27 @@
...
+
+		var artist = null;
+		var track = null;
+
+		if (separator !== null) {
+			artist = text.substr(0, separator.index);
+			track = text.substr(separator.index + separator.length);
+		}
+
+		return {artist: artist, track: track};
```
inverse: I would change this to have a single point of return - Perhaps init the track/artist vars outside of the first If. What do you think?

connectors/v2/sullen-ural.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = $('#tracks-async').length ? '#tracks-async' : ($('.band-col-right').length ? '.band-col-right' : '.album-tracks');
```
inverse: Could we make this a bit more readable? Having multiple ternary operators in a single statement is pretty hard to read

connectors/v2/sullen-ural.js
alexesprit: Sorry, didn't understand what you mean.

connectors/v2/sullen-ural.js
alexesprit: You're totally right. I'll update the code soon.

alexesprit:
Simplified setup of player selector.

`#tracks-async` and `.band-col-right` selectors are parents of `.album-tracks` one, so it makes no sense to use them.

inverse:
Much nicer!

PR #1127, opened 2016-11-06
Reimplement SoundCloud connector

### LINKED ISSUE #1106: SoundCloud Scrobbling
I toggle the repeat playing track button but the add-on doesn't scrobble on repeat.

f1u77y:
- Uses only DOM events and data
- `getCurrentTime`() is implemented
- Listens to 'progress' events

Fixes #1106


alexesprit:
It doesn't get artist while listening some songs, e.g. https://soundcloud.com/austra_midi/utopia-ikonika-remix. I understand why it happens, it's just a note.

I see only two options to fix it:
* use mixed DOM listening and script injecting;
* get artist by retrieving it from a song page (https://github.com/david-sabata/web-scrobbler/pull/1050#issuecomment-243073660 as example).

I'm not sure about 2nd option, but, in theory, it should work.

f1u77y:
Ok, I can fetch username from the song page, but we cannot understand if it's artist name or something else.
Ideally we should do something like returning array of possible `{arist, track}` and trying to get track info for each of them from last.fm

f1u77y:
Closing this in favor of #1134

PR #1129, opened 2016-11-06
Add amazon.co.jp

mft:
amazon.co.jp also services prime music, so extend the patterns for Amazon to include it.


inverse:
Thanks!

PR #1130, opened 2016-11-06
Delete useless connectors, pt. 2

alexesprit:
In addition to #1122.

I also noticed Blitzr website doesn't allow to play songs directly on the website. Should it be removed too?


inverse:
Perhaps it did at one time. Lets clean it up!

PR #1133, opened 2016-11-06
Removed unused radioplayer.js

inverse:
Removed radioplayer.js as I couldn't see it referenced. 

@alexesprit could you review this please :+1:


inverse:
We can always bring it back in a better form when people ask :+1:

PR #1134, opened 2016-11-07
Fix soundcloud scrobbling of repeated track

### LINKED ISSUE #1106: SoundCloud Scrobbling

f1u77y:
Watch for DOM changes for triggering `onStateChanged()` but continue to
use internal SC API for fetching track metadata since it seems to be the
only way to fetch username which is sometimes the artist name

Fixes #1106

### DIFF
--- a/connectors/v2/soundcloud.js
+++ b/connectors/v2/soundcloud.js
@@ -8,6 +8,7 @@ var track;
 var duration;
 var uniqueID;
 var artwork_url;
+var isPrivate = false;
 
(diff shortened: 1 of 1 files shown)


alexesprit:
Works fine! 👍

david-sabata:
You removed the check for private track, was that intentional?

f1u77y:
@david-sabata whoops, of course no. Fixed

PR #1137, opened 2016-11-08
Fix EgoFM connector

alexesprit:
EgoFM's player isn't located at main website directly, so I've left a dummy URL pattern to make sure the user is noticed the EgoFM is supported by the extension. :) Also, see my question about this case in Gitter chat.

### DIFF
--- a/connectors/v2/ego-fm.js
+++ b/connectors/v2/ego-fm.js
@@ -2,23 +2,10 @@
 
 /* global Connector */
 
-Connector.playerSelector = '#playerInfo';
+Connector.playerSelector = '.radioplayer-head';
 
-Connector.artistSelector = '#current > span.artist';
+Connector.artistTrackSelector = '.scrolling-text';
 
 Connector.trackSelector = '#current > span.song';
 
-Connector.isPlaying = function () {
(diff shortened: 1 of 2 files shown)


inverse:
Not sure how many people use that as an indicator but it should be alright :+1:

signaleleven:
Thank you for the fix. I added the connector because I needed it, but I was a bit ashamed of it. 
Somehow, I could not have the extension reach into the iframe on addradio.de to read the current playing track....I don't remember the details. I am glad it's working now.

connectors/v2/ego-fm.js
```diff
@@ -2,23 +2,10 @@
 
 /* global Connector */
 
-Connector.playerSelector = '#playerInfo';
+Connector.playerSelector = '.radioplayer-head';
 
-Connector.artistSelector = '#current > span.artist';
+Connector.artistTrackSelector = '.scrolling-text';
 
 Connector.trackSelector = '#current > span.song';
```
signaleleven: This line is probably not needed anymore...

alexesprit: Yes, you're right. I forgot to delete the selector.

PR #1140, opened 2016-11-09
Youtube: fix repeated scrobbling

### LINKED ISSUE #918: Scrobble youtube repeat
After pressing repeat button which appears instead of play button after the video finishes, scrobbling starts again - which is perfectly fine.

However, when using repeat function added by Magic Actions extension 
or 
right-clicking video and selecting Loop - scrobbling won't start again. It just stays on Scrobbled (song name), no matter how many times video gets repeated.

Is it just me? Am I doing something wrong? Or is this "function" not added yet. If so, would you consider looking into it, perhaps working it out somehow?

Thanks for great extension though ;-)

f1u77y:
- Listen to timeupdate events generated by `<video>` element instead of
using MutationObserver
- Use `currentTime` and `duration` properties of `<video>` element
instead of text values in `.ytp-time-*`

Fixes #918

UPD: progress -> timeupdate

### DIFF
--- a/connectors/v2/youtube.js
+++ b/connectors/v2/youtube.js
@@ -14,9 +14,27 @@ chrome.storage.local.get('Connectors', function(data) {
 	}
 });
 
-Connector.playerSelector = '#page';
+Connector.videoSelector = '#player-api .html5-main-video';
+
+{
+	$(Connector.videoSelector).on('timeupdate', () => {
+		if (Connector.isStateChangeAllowed()) {
+			Connector.onStateChanged();
+		}
+	});
+}
 
 Connector.artistTrackSelector = '#eow-title';
+
+Connector.escapeBadTimeValues = function (time) {
+	if (typeof time !== 'number') {
+		return null;
+	}
+	if (isNaN(time) || !isFinite(time)) {
+		return null;
+	}
+	return time;
+};
 
 /**
  * Because player can be still present in the page, we need to detect that it's invisible
@@ -26,9 +44,8 @@ Connector.getCurrentTime = function() {
 	if (isPlayerOffscreen()) {
 		return null;
 	}
-
-	var $time = $('#player-api .ytp-time-current');
-	return this.stringToSeconds($time.text());
+	const currentTime = $(this.videoSelector).prop('currentTime');
+	return this.escapeBadTimeValues(currentTime);
 };
 
 /**
@@ -37,11 +54,10 @@ Connector.getCurrentTime = function() {
  */
 Connector.getDuration = function() {
 	if (isPlayerOffscreen()) {
-		return 0;
+		return null;
 	}
-
-	var $duration = $('#player-api .ytp-time-duration');
-	return this.stringToSeconds($duration.text());
+	const duration = $(this.videoSelector).prop('duration');
+	return this.escapeBadTimeValues(duration);
 };
 
 Connector.getUniqueID = function() {


alexesprit:
The updated connector doesn't recognize now playing song because `videoSelector` has never class `playing-mode`.

```diff
diff --git a/connectors/v2/youtube.js b/connectors/v2/youtube.js
index 5edd041..cb2f517 100644
--- a/connectors/v2/youtube.js
+++ b/connectors/v2/youtube.js
@@ -68,7 +68,7 @@ Connector.getUniqueID = function() {
 };

 Connector.isPlaying = function() {
-       return $(this.videoSelector).hasClass('playing-mode');
+       return $('#player-api .html5-video-player').hasClass('playing-mode');
 };

 Connector.isStateChangeAllowed = function() {
```

f1u77y:
Whoops, that was ".html5-video-player" class, haven't noticed that. Reverted

f1u77y:
(and I'm wary of the fact that progress events are generated very rarely(1 time per ~6 secs), so sometimes there can be no progress events at the last 5% of song and it won't be marked for repeat)

alexesprit:
@f1u77y I noticed you call the `Connector.onStateChanged` function directly. In this case the result of the `Connector.isStateChangeAllowed` function is ignored.

f1u77y:
> In this case the result of the `Connector.isStateChangeAllowed` function is ignored

OK but isn't it better just to use this function inside default implementation of `Connector.onStateChanged` rather then outside of it? (I see you're talking about the same)

> The video element generates 'progress' events about 8-10 times per song

Ahh, I misused this event(it's about *loading* progress but not *playing*). Will fix.

f1u77y:
> Is there a way to watch these events somewhere in dev console?

`monitorEvents`

alexesprit:
> monitorEvents

Ok, thanks.

Seems like the connector works fine.

> OK but isn't it better just to use this function inside default implementation of Connector.onStateChanged rather then outside of it?

Yep, I meant the same thing.

connectors/v2/youtube.js
```diff
@@ -14,10 +14,28 @@ chrome.storage.local.get('Connectors', function(data) {
 	}
 });
 
-Connector.playerSelector = '#page';
+Connector.videoSelector = '#player-api .html5-main-video';
+
+{
+	$(Connector.videoSelector).on('timeupdate', () => {
+		if (Connector.isStateChangeAllowed()) {
```
alexesprit: I'm just wondering why did you put event listener setup into curly braces? Is there a certain reason for that?

f1u77y: To show that's a separate block for setup(not sure if it's needed because the whole file is connector setup)

alexesprit:
@f1u77y I noticed some regression after merging this PR. Here are steps to reproduce:
1. Open Youtube front page.
2. Select any music video that should be recognized by the extension.

The selected music video won't be recognized, but it actually should.

Can you reproduce the issue? If so, have you any thoughts on how to fix it w/o reverting the commit?

f1u77y:
Yes, that's because the video loads in the same page. We should search for the video element with `MutationObserver`

alexesprit:
@f1u77y Ideally we should also stop search if video element is found and start search again if video element is missing on the page (e.g. user moves back on front page).

alexesprit:
@f1u77y https://github.com/alexesprit/web-scrobbler/commit/5e92958834ba3c021581f44b29b24a0c0d55980b perhaps looks too complicated but I don't know how to simplify the code/reimplement by using another approach . Have you a better solution?

f1u77y:
I wrote mostly the same code, and I don't see any way to do it in a better way.
(BTW, usage of `MutationObserver` with all flags on looks like overkill, doesn't it?)

alexesprit:
Ok, so I'll commit these changes, since I don't want to revert this PR.

> (BTW, usage of MutationObserver with all flags on looks like overkill, doesn't it?)

I enabled only two flags which are required to observe visibility state of the video element.

alexesprit:
Youtube changed something on their website, so the connector from this PR works fine again.

Issue #1128, opened 2016-11-06
Stripping garbage from track names


f1u77y:
Some internet services provide tracks with many garbage in the tags. This garbage seems to be at most common between services(for instance youtube and soundcloud connectors share (mostly) same function for stripping that garbage). I think it'd be better to implement one global anti-garbage function and put it to the core.

### Further discussion topics
1. ~~Can such function be incompatible to some services?~~
2. ~~If artist and title are in the same string(what a pain), is it better to remove such garbage before or after separating artist from title?~~
3. Should we provide ability to add more garabge-detecting regexes for users?
4. **UPD** Sometimes connector should be able to pass more than one variant of parsed {artist, track}(when artistTrack is ambigous, e.g. `"artist with- dashes -track title"` or some cases like username on soundcloud). last.fm should handle such cases better than we can do.

alexesprit:
https://github.com/alexesprit/web-scrobbler/tree/metafilter

inverse:
@alexesprit I think what you're doing is a good start. Perhaps opening up a PR on this topic to perhaps get some input from others... What do you think?

alexesprit:
@inverse I think we should get rid of v1 connectors first.

f1u77y:
Closing this because relevant PR merged

alexesprit:
> Should we provide ability to add more garabge-detecting regexes for users?

It's supported already:
1. Connector developers can add new rules (regexes) to `Metadata.YoutubeTrackFilters` object to update `MetadataFilter.youtube` behaviour.
2. They also can write custom filter functions with usage of regexes.

PR #1142, opened 2016-11-09
Implement shared metadata filter

alexesprit:
Possible solution of #1128.

All description is placed in `core/content/filter.js` module.

Usage examples (it's insanely easy for common cases):
* bandcamp.js, vk.js (common filter with multiple functions);
* bandzone.js (filter using custom filter function);
* youtube.js (Youtube-based filter).

By default, a trim filter is used.

I've checked it on Youtube and some other websites. The filter works fine, so I guess it works fine at all (just because it's a common mechanism for all connectors).

I also merged this branch into `grunt-test` branch and run tests. Didn't find any regressions too.

### DIFF
--- a/connectors/v2/alexa.js
+++ b/connectors/v2/alexa.js
@@ -20,7 +20,7 @@ Connector.getArtist = function() {
 			return null;
 		}
 		var results = songTitle.split('-');
-		return results[0].trim();
+		return results[0];
 	} else {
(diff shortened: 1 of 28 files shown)


david-sabata:
This looks really good :+1:

inverse:
Looks interesting - I think this approach is def the way to go - so flexible!

alexesprit:
Fixed conflicts with master branch.

PR #1146, opened 2016-11-11
Remove Virgin Radio Turkiye connector

alexesprit:
VRT website has broken player and cannot play the music.

Radio Eksen which is covered by this connector too always provides "REKLAM" as artist and title while actually the music is playing.


inverse:
:+1:

inverse:
Almost there ;)

PR #1148, opened 2016-11-11
Remove 22tracks connector

alexesprit:
22track contains only short previews of tracks. So it's completely useless, because short tracks won't be scrobbled by the core.


inverse:
Makes sense

PR #1150, opened 2016-11-11
Fix Youtubify playing state recognition

alexesprit:
Youtubify hides play button while loading a song on startup. It makes the connector think the song is playing.


inverse:
Thanks :)

PR #1152, opened 2016-11-12
VK: update track info on every event

f1u77y:
Track info could be undefined on early startup, so sometimes we need to update
it on progress


alexesprit:
👍 I noticed this issue too.

inverse:
Thanks :+1:

PR #1153, opened 2016-11-12
Merge RadioTunes and JazzRadio into one connector

alexesprit:
There is no reason to keep multiple connectors for identical websites.


inverse:
Makes sense :+1:

alexesprit:
I found no information about owners of these websites.

PR #1155, opened 2016-11-12
Add RockRadio support

alexesprit:
The 'radiotunes' connector is used because these websites have the same layout.


inverse:
:+1:

PR #1157, opened 2016-11-13
Don't process new state if song is paused/stopped

alexesprit:
This makes the controller to start pipeline processing, which looks strange for stopped songs.


alexesprit:
What these changes resolve:
1. open http://youtubify.vebto.com/ using the extension from latest master (where youtubify is fixed);
2. the extension icon changes to a magnifying glass icon (means the pipeline process is started), but the song actually isn't playing (youtubify doesn't auto-play songs on startup).

I noticed that on other websites (e.g. VK), youtubify is used just as example.

PR #1159, opened 2016-11-14
Switch Daytrotter connector to v2 format

alexesprit:


core/connectors.js
```diff
@@ -230,8 +230,9 @@ define(function() {
 
 		{
 			label: 'Daytrotter',
-			matches: ['*://www.daytrotter.com/*'],
-			js: ['connectors/daytrotter.js']
+			matches: ['*://www.daytrotter.com*'],
```
inverse: Should it have the trailing / before the * or does it not matter?

connectors/v2/daytrotter.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Example: http://www.daytrotter.com/radio.html?playlistid=1239601&trackid=4893828
+ */
+function isRadioPlayer() {
+	return $('body').hasClass('radio');
+}
+
+function setupConnector() {
```
inverse: Super clear :+1:

inverse:
Aside the comments looks good :+1:

core/connectors.js
alexesprit: It should be fixed, of course.

alexesprit:
Hmm, Daytrotter Radio and Paste Magazine Radio have the same layout.

inverse:
Interesting. But this one implements more functionality right?

alexesprit:
Yes, this connector covers Daytrotter Radio (copy of PasteMagazine Radio) and Daytrotter itself.

Of course, we can drop `Add Daytrotter Radio support` commit and use `pasteradio` connector for Daytrotter Radio, but it's such a mess.

So, the PR is ready for merge.

inverse:
:+1:

Issue #1143, opened 2016-11-10
Rerurning more than one value in Connector.getArtistTrack


f1u77y:
Sometimes connector should be able to return more than one variant of parsed `{artist, track}`(when artistTrack is ambigous, e.g. `"artist with- dashes -track title"` or some cases like username on soundcloud). last.fm should handle such cases better than we can do.

inverse:
I'm not sure I follow? Could you give more concrete examples?

f1u77y:
@inverse 
From Soundcloud:
* [example 1](https://soundcloud.com/dopedod/dopedodbongra) parsed incorrectly
* [example 2](https://soundcloud.com/william-ekkebus/fruitpochette-the-crest-of-evil-01-shakunetsu-drop?in=william-ekkebus/sets/fruitpochette-the-crest-of-evil) parsed incorrectly

rob9315:
example one seems to be parsing correctly nowadays

<img width="437" height="170" alt="Image" src="https://github.com/user-attachments/assets/8b2ff975-a3ce-4256-886e-c96af5a29cfe" />

I suggest this be issue be closed

PR #1163, opened 2016-11-17
Fix VK connector

### LINKED ISSUE #1162: Vk.com: suddenly stops scrobbling
Hello! Today I saw thats scrobbling stop working on Vk.com. I'm using Google Chrome 54.0.2840.99 m.

alexesprit:
Resolves #1162.

### DIFF
--- a/connectors/v2/vk.js
+++ b/connectors/v2/vk.js
@@ -2,13 +2,14 @@
 
 /* global Connector, MetadataFilter */
 
-var INFO_ID = 0;
-var INFO_TRACK = 3;
-var INFO_ARTIST = 4;
-var INFO_DURATION = 5;
+const INFO_ID = 0;
+const INFO_TRACK = 3;
+const INFO_ARTIST = 4;
+const INFO_DURATION = 5;
 
-var trackInfo = null;
-var isPlaying = false;
+let trackInfo = null;
+let progressInfo = null;
+let isPlaying = false;
 
 (function () {
 	$(document.documentElement).append($('<script>', {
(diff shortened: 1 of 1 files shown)


f1u77y:
@alexesprit that DOM node seems to be on every VK page. Maybe we should consider using it as primary metadata source?

alexesprit:
@f1u77y I'd prefer fetching the data from DOM node too, but it contains artist and track in one place, so the song metadata can be parsed incorrectly (although, such cases are pretty rare).

To be honest, I don't like this fallback; in this case we cannot see if localStorage keys are changed (that means the replays are broken).

f1u77y:
@alexesprit 
> it contains artist and track in one place, so the song metadata can be parsed incorrectly

See #1143, that should fix a lot of that cases

> we cannot see if localStorage keys are changed

I've also found the `audio` object that seems to be more reliable then `localStorage` keys

alexesprit:
> See #1143, that should fix a lot of that cases

Yes, but we should implement this feature first. I have no idea how to do that atm.

> I've also found the audio object that seems to be more reliable then localStorage keys

Do you know how to get the data from this object? Or is it impossible?

f1u77y:
> Do you know how to get the data from this object?

It has `id`, `performer`, `title` and `duration` keys(but we need to send it from vk-dom-inject script because it's not in DOM)

Progress could be got with `ap._impl._currentAudioEl` which is an `<audio>` node but I dunno if it's more stable than `localStorage`

alexesprit:
I didn't find out how to get progress percentage, but I implemented fetching metadata from the audio element.

f1u77y:
Something like [this](https://github.com/f1u77y/web-scrobbler/tree/vk-audio-elem-metadata-source)

alexesprit:
Looks neat!

Is it worth opening a new PR with your patch instead of this one?

alexesprit:
Closed in favor of #1164.

PR #1164, opened 2016-11-18
VK: fetch metadata and state from objects instead of storage

### LINKED ISSUE #1162: Vk.com: suddenly stops scrobbling

f1u77y:
These objects(`audio` and `ap._impl._currentAudioEl`) seem to be more stable
then 'audio_vXX_whatever' localStorage items

Fixes #1162

### DIFF
--- a/connectors/v2/vk.js
+++ b/connectors/v2/vk.js
@@ -2,62 +2,49 @@
 
 /* global Connector, MetadataFilter */
 
-var INFO_ID = 0;
-var INFO_TRACK = 3;
-var INFO_ARTIST = 4;
-var INFO_DURATION = 5;
-
-var trackInfo = null;
-var isPlaying = false;
+let trackInfo = {};
+let isPlaying = false;
 
 (function () {
 	$(document.documentElement).append($('<script>', {
 		src: chrome.extension.getURL('connectors/v2/vk-dom-inject.js'),
 	}));
-	$(window).on('message', (event) => {
-		let eventType = event.originalEvent.data.type;
-		if (typeof eventType !== 'string') {
+	$(window).on('message', ({originalEvent: event}) => {
+		if (event.data.sender !== 'web-scrobbler') {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


connectors/v2/vk.js
```diff
@@ -2,62 +2,47 @@
...
+		if (typeof event.data.type !== 'string') {
 			return;
 		}
-		if (eventType.startsWith('vk:player')) {
-			if (eventType.endsWith('start')) {
+		if (event.data.type.startsWith('vk:player')) {
+			if (event.data.type.endsWith('start')) {
 				isPlaying = true;
-			} else if (eventType.endsWith('stop') || eventType.endsWith('pause')) {
+			} else if (event.data.type.endsWith('stop') || event.type.endsWith('pause')) {
```
alexesprit: It should be `event.data.type` in the second condition.

f1u77y:
Whoops, tabs

alexesprit:
`lintspaces` is configured to disallow spaces as indentation.

f1u77y:
@alexesprit I'm talking about indenting with tabs and aligning with spaces

alexesprit:
@f1u77y I understood you, but lintspaces assumes that spaces is an indentation, not an alignment.

P.S. Mixed tabs and spaces is not that thing should be used. That's why spaces are better for indentation.

f1u77y:
> Mixed tabs and spaces is not that thing should be used.

But that's the only way to align code with tabs in tab-size-agnostic way.

Sigh...

alexesprit:
> But that's the only way to align code with tabs in tab-size-agnostic way.

No, because tab width is configurable.

f1u77y:
Anyway, I've updated the PR to don't use spaces for alignment

(and I don't think that PR discussion is suitable place for yet another tabs-vs-spaces arguement)

PR #1165, opened 2016-11-18
Add base support for Firefox

alexesprit:
A couple of compatibility patches for Firefox.


connectors/v2/soundcloud-dom-inject.js
```diff
@@ -17,12 +17,12 @@ window.SC_ATTACHED = window.SC_ATTACHED || false;
 	}
 	/*jslint unparam: true */
 	window.webpackJsonp([], {
-		0: function (e, t, require) {
-			var modules = require.c, moduleid, el;
+		0: function (e, t, _require) {
```
inverse: How does this change work?

core/background/main.js
```diff
@@ -80,7 +80,7 @@ require([
 
 			if (!isActiveSession) {
 				isActiveSession = true;
-				GA.send('pageview', '/background-injected?version=' + chrome.app.getDetails().version);
+				GA.send('pageview', '/background-injected?version=' + chrome.runtime.getManifest().version);
```
inverse: How does Firefox handle this? 

I was reading a bit about: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/getManifest

and in their examples they use:

```
browser.runtime.getManifest()
```

core/background/notifications.js
```diff
@@ -126,15 +137,15 @@ define([
 				GA.event('notification', 'authenticate', 'open-unavailable');
 
 				// fallback for browsers with no notifications support
-				window.open(authUrl, 'scrobbler-auth');
+				chrome.tabs.create({url: authUrl});
```
inverse: Same with this

inverse:
Just some comments really, haven't tested.

core/background/main.js
alexesprit: Don't know about the example, but `chrome.runtime.getManifest` works fine as well as other Chrome API functuions (those, that implemented in Firefox, of course).

connectors/v2/soundcloud-dom-inject.js
alexesprit: AMO linter doesn't like variable named as `require`:
```
1) Unexpected global passed as an argument
Passing a global as an argument is not recommended. Please make this a var instead.
connectors/v2/soundcloud-dom-inject.js line 25
```

f1u77y:
What's blocking the merge?

PR #1167, opened 2016-11-18
Implement storage for manually entered metadata

alexesprit:
A (probably) proper implementation of #826.

### DIFF
--- a/core/background/chromeStorage.js
+++ b/core/background/chromeStorage.js
@@ -83,21 +83,29 @@ define([
 		chrome.storage.local.set(data, innerCb);
 	}
 
+	function hideStringInText(str, text) {
+		if (str) {
+			return text.replace(str, 'xxxxx' + str.substr(5));
+		}
+		return text;
+	}
+
 	/**
 	 * Logs all storage data to console
 	 */
-	function debugLog() {
+	function debugLog(namespace) {
 		get(function(data) {
-			var text = JSON.stringify(data);
-			if (data !== null && Object.keys(data).length > 0 && data.LastFM) {
-				if (data.LastFM.token) {
-					text = text.replace(data.LastFM.token, 'xxxxx' + data.LastFM.token.substr(5));
-				}
-				if (data.LastFM.sessionID) {
-					text = text.replace(data.LastFM.sessionID, 'xxxxx' + data.LastFM.sessionID.substr(5));
-				}
+			let namespaceData = data[namespace];
+			if (!namespaceData) {
+				return;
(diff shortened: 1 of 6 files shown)


inverse:
Looks good to me @david-sabata what do you think?

alexesprit:
The `ChromeStorage.debugLog` function should be changed somehow, because a lot of saved song metadata can produce large console output.

The possible solution is to add new argument to `ChromeStorage.debugLog` called like `namespace` that means what namespace data should be used to log it to console. This changed `debugLog` function should be called separately in every module that wants to show its storage data instead of calling this function once in main module.

Have someone better ideas how to resolve this?

alexesprit:
I just thought about using sync storage to keep song metadata, but I'm not sure if it's won't bring regressions. Perhaps, SyncStorage class should have ability to fallback to `chrome.local.storage` StorageArea usage on write errors.

https://github.com/alexesprit/web-scrobbler/commits/sync-storage

f1u77y:
> regressions

* Sync storage has limit on number of performed operations(but as far as I know all operations are cause by user actions, so this shoud not be a problem)
* Sync storage has either limit on number of stored items and 5x less limit on number of bytes which could be harmful for song corrections(there could be many of them)
* Firefox has no sync storage, so if we want to support it, we shoud use local storage if sync is not defined.

> Perhaps, SyncStorage class should have ability to fallback to chrome.local.storage StorageArea usage on write errors.
And then we should collect options/corrections/cache from two sources. That's bad because **sync** will be useless in that case.

alexesprit:
@f1u77y hmm, you're right.

f1u77y:
On the other side, using sync storage for corrections is nice because re-entering all of them is somewhat annoying. So, now I see two ways of transferring them to other machine:
1. Use hacks for using taking maximum from sync storage. This could be done with using one item for several corrections(they could take most of the limit, right?). This way is ugly but the most user-friendly way I know. And it should work(102,400 bytes is enough for 1000-2000 corrections, I think). A user who enters more then 1000 corrections should be a monster (oh wtf am I writing now, I seriously think about having nice time with bytes while coding JS?!)
2. Add ability to export all the corrections from local storage in eg. a text file for importing it on other machine. That's ugly and not user-friendly but this works.
3. Do nothing and let users have fun re-entering their corrections again and again on different machines.

alexesprit:
@f1u77y I also noticed some extensions use cloud services to sync their data between different machines.
Anyway, it's another task to implement such storages.

PR #1168, opened 2016-11-18
Remove v1 core

alexesprit:
Also, few questions:
* Should we move existing connectors from `connectors/v2` directory to `connectors` one?
* Should we remove version property from connector definition in `connectors.js`?
* Should we remove 'v2' mark in the extension options? http://i.imgur.com/TQ5Z6Mh.png


f1u77y:
N/n/n in my opinion.

Or y/y/y if we are sure that there won't be future versions of connector API.

PR #1169, opened 2016-11-18
Remove 'Remastered...'-like strings from song metadata

### LINKED ISSUE #880: Automatic elimination of unwanted title extensions like "Remastered..."
Hi, when listening to the new Beatles catalogue on [Spotify](https://open.spotify.com/artist/3WrFJ7ztbogyGnTHbHJFl2) and [Deezer](http://www.deezer.com/artist/1), I noticed that the native scrobbler of Deezer's web player does no longer submit the unwanted title extensions like "Remastered 2009" etc. to Last.fm. They seem to have implemented an internal filter for that which is a very welcome feature of course, because Spotify's title extensions spoil your Last.fm statistics due to automatically creating a new page for every "new" track.

Long story short: is it possible to implement such an automatic filter in your web scrobbler, too?

alexesprit:
Applied to Spotify and Deezer connector.

Resolves #880.

### DIFF
--- a/core/content/filter.js
+++ b/core/content/filter.js
@@ -138,6 +138,15 @@ MetadataFilter.youtube = function(text) {
 };
 
 /**
+ * Remove "Remastered..."-like strings from the text.
+ * @param  {String} text String to be filtered
+ * @return {String} Filtered string
+ */
+MetadataFilter.removeRemastered = function(text) {
+	return MetadataFilter.filterWithMap(text, MetadataFilter.REMASTERED_FILTERS);
+};
+
+/**
  * Replace text according to given map object.
  * @param  {String} text String to be filtered
  * @param  {Object} map  Object contains rules of replace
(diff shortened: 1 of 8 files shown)


core/content/filter.js
```diff
@@ -134,13 +134,31 @@ MetadataFilter.decodeHtmlEntities = function(text) {
...
+ * @return {String} Filtered string
+ */
+MetadataFilter.filterWithMap = function(text, map) {
+	for (let source in map) {
+		let target = map[source];
 
 		let regexParts = source.match(new RegExp('^/(.*?)/([img]*)$'));
 		if (!regexParts) {
-			console.log('Invalid Youtube filter: ' + source);
+			console.log(`Invalid regex: ${source}`);
```
inverse: What's with the weird quotes being used here?

alexesprit: It's sort of ES6 magic called 'template strings'.

PR #1183, opened 2016-11-30
Add rbmaradio.com connector

joe-niland:
This is my first attempt at a connector. It handles the player on https://www.rbmaradio.com

I think the playing detection still needs work, but not sure.


connectors/v2/rbmaradio.js
```diff
@@ -34,14 +34,14 @@ Connector.getDuration = function () {
 	var $progressControl = $('.c-player__controls > .c-player__progress > progress');
 	var duration = '';
 	if ($progressControl) {
-		duration = $progressControl.attr('max'); 
+		duration = $progressControl.attr('max');
```
alexesprit: There's missing `parseInt` call at this line. The `getDuration` function should return numbers.

alexesprit:
image

joe-niland:
Thanks for the review @alexesprit !

The show I was testing with didn't have individual track listing. Have fixed this now.
Also removed the track art selector because they only have one image per Show.

joe-niland:
Thank you @alexesprit

PR #1187, opened 2016-12-04
Browser CI testing for connectors, take 2

alexesprit:
## Main changes compared with original PR

**Added:**
* Add more connectors tests;
* Add ability to run certain tests;
* Add script that prints missing tests;
* Add advanced debug mode.

**Changed:**
* Make connector spec more flexible (see `connector-spec` documentation);
* Retry tests up to 5 times;
* Run tests one time each;
* Clean up code and make it JSHint-valid.

**Fixed:**
* Fix communication between the extension and tests.

Of course, there are a lot of other changes, but they're mostly internal. In particular, all functions were promisificated; removed most hacks and strange recursions. Almost all code was documented for easier understanding of it.

## What changed in the core
* Added `TestReporter` module which helps to communicate the core with test suite;
* `Connector.currentState.url` is now equal `window.location.href`;
* Changed minimum required version of Node.js to 6.0.0.

## How to use
Just install new dev dependencies and run tests:
```
> npm install
> grunt test
```
It's not neccessary to add Chromedriver to `PATH` since `chromedriver` package does it automatically.

## How to run certain tests
Usage:
```
> grunt test:%filename%
```
where `%filename%` is a connector/test filename w/o extension.

Examples:
```
Run single test for 8tracks connector
> grunt test:8tracks
```
```
Run tests for Hype Machine and Dash Radio connectors.
> grunt test:hypem:dashradio
```
Note: tests still run in alphabetical order.

## How to use command arguments
* `debug`: enable or disable debug mode. Disabled by default.  
Use `true|on|1` value to enable debug mode and `false|off|0` to disable it.
* `quitOnEnd`: close browser when all tests are completed. Enabled by default.  
Use `true|on|1` value to enable this feature and `false|off|0` to disable it.

Examples:
```
Enable debug mode and run single test for 8tracks connector
> grunt test:debug=on:8tracks
```
```
Just enable debug mode and run all tests
> grunt test:debug=on
```

## How to print missing tests
It's quite easy to do that:
```
> node scripts\print-tests-info.js         
Missing tests [2]:
 * alexa.js
 * amazon.js
Unused tests [1]:
 * 163music.js
 ```

## How to use optional uBlock extension
The `ublock.zip` extension file should be placed into `tests` directory.
The 'Using uBlock extension: ...' message on startup means uBlock is loaded successfully.

## Examples of tests
* accujazz (uses custom `waitUntil` function to skip advertisment);
* baidu (checks if player element exists only);
* vk (checks if connector is injected into a page);
* wonderfm (checks multiple domains covered by a single connector).


alexesprit:
I'm not sure if all tests behave correctly in different locations; the test interface can be extended to allow or disallow running tests for geolocked websites if necessary.

alexesprit:
Added Grunt `build` task that preprocess code and create zip package w/o test code.

PR #1191, opened 2016-12-07
YouTube and iTunes...

Angelfirenze01:
I've logged into two separate profiles and am still finding that my Last.fm account isn't scrobbling properly.  Do I need to execute a new version?


alexesprit:
What is this?

Angelfirenze01:
My mistake, I'll open an issue.

PR #1199, opened 2016-12-18
Core: make MusicBrainz pipeline optional

### LINKED ISSUE #1156: Don't wait for searcing for cover
It's a minor functionality but it can slow down fetching track info(e.g. when MB is down)

f1u77y:
MusicBrainz pipeline will now run after 'isProcessed' flag is set, so it will
not delay song recognition.

Fixes #1156

### DIFF
--- a/core/background/pipeline/pipeline.js
+++ b/core/background/pipeline/pipeline.js
@@ -12,26 +12,31 @@ define([
 	return {
 		processSong: function(song) {
 			// list of processors is recreated for every processing call
-			var processors = [
+			const processors = [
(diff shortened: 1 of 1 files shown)


alexesprit:
Now playing notifications can be displayed w/o a coverart, because the coverart can be fetched by MusicBrainz pipeline stage after nofification is created.

There's an ability to listen changes of Song object fields, but I don't know how to mix it and wait timeout (I think it's should be done anyway) atm.

inverse:
This still valid?

f1u77y:
The issue is not reproducible, so i'll close the PR.

PR #1202, opened 2016-12-22
Implement ability to edit now playing songs

### LINKED ISSUE #245: Customize Artist/TItle
Just like you can edit them for unrecognized songs, I would like the option to edit the Artist/Title of anything I scrobble.
Instead of being forced to rely on an automatic system that sometimes makes the wrong choice (reversed artist and title, having album name for artist or just being plain wrong), and gives me no option to correct.

### LINKED ISSUE #371: What do we do about songs scrobled with incorrect info?
For example, https://www.youtube.com/watch?v=UoK8DaJRDaM
This is scrobled in my library as きゃりーぱみゅぱみゅ – "CANDY CANDY , Kyary Pamyu Pamyu - CANDY CANDY"
When it should simply be きゃりーぱみゅぱみゅ – "CANDY CANDY"
or Kyary Pamyu Pamyu - "CANDY CANDY"
In the current system, there's no way to fix titles showing up incorrectly.

### LINKED ISSUE #619: Editing scrobbled songs
Scrobbler's UI is too limited. You can manually edit song to be scrobbled only if scrobbler doesn't recognize it. Usually it thinks it does and too often results are garbage like Stevie Wonder - I Wish (Live in London 1995). Then I have to go remove the entry and scrobble it manually. I think user should be able to correct it straightly from the interface to Stevie Wonder - I Wish. YouTube video titles are seldom in a correct form. You could even add a feature which links corrections to YouTube video links if they got enough support.

### LINKED ISSUE #968: Scrobbling
You can't change the song/artist of a scrobble if it has found something (even if what it has found is wrong...) :( I'd like to be able to change the scrobble details even if it has found something...
And you need to refill every time you play something that isn't found by the extension itself... Maybe you can make it so that it remembers what link is for what song (this is at Youtube)....

danivenk

### LINKED ISSUE #982: Wrong Title Scrobbled
Some songs are already "recognized" but with a wrong Artist name, because a few previous users hit enter on the standard artist name that the scrobbler filled in,
So i.e.: I'm listening to a track "A1. Shlømo - The Rapture", the name of the artist is "Shlømo", but it's a playlist he shared so he changed the name to A1 for the first song, A2 for the second, ...
Thus the scrobbler automatically recognizes "A1. Shlømo" as the artist's name, people who get the grey question mark icon just hit enter without reviewing and now the song gets scrobbled wrong for all users. I can't find a way to correct this, not using the extension nor using the last.fm website.

So basically it would be nice if we can still correct a song's artist name to what it's supposed to be, even after x previous users hit enter on the standard name and the scrobbler sees it as a validated name.

I tried searching the issues but couldn't really find anything similar... So sorry if I'm re-requesting a feature here or if it already exists. If so: how do I access this feature?

Cheers!

link to the artist's Soundcloud (I just hit play on the first playlist): https://soundcloud.com/shl-mo

### LINKED ISSUE #1021: Allow supplying album for unrecognised tracks
When tracks are unrecognised we should provide the option to provide the album as an optional field. 

refs #1014

alexesprit:
This PR is possible replacement of #719. I know that #719 brings more features, but it's unmergeable, and seems like no one wants to update it.

The editing mode differs from how it's made in #719, so I'd be glad to get comments on current implementation.

Features:
* Show album in info popup;
* Edit now playing song (even if it's recognized by Last.fm);
* Revert user changes;
* Work side by side with #1167.

Resolves #245.
Resolves #371.
Resolves #619.
Resolves #968.
Resolves #982.
Resolves #1021.


f1u77y:
Looks nice.

Some proposals:

* I think something like "edited" should be added to the popup if song is edited by user
* There should be (optional?) button like "Revert" or "Resubmit original info"
* Saving these edits in storage would be really nice(OK, that's not about popup UI but anyway)

alexesprit:
> There should be (optional?) button like "Revert" or "Resubmit original info"

Ok, I'll look into that.

> I think something like "edited" should be added to the popup if song is edited by user

Well, I think, the 'edited' label is redundant, when the 'revert' button is visible. And the latter should be visible only for edited songs.

> Saving these edits in storage would be really nice(OK, that's not about popup UI but anyway)

It should works fine with #1167, in theory.

f1u77y:
> Well, I think, the 'edited' label is redundant, when the 'revert' button is visible. And the latter should be visible only for edited songs.

Yes, that's the best option for that.

alexesprit:
I have some questions:
1. Should we close the popup when user clicks on 'Submit' button?
2. Should we close the popup whne user clicks on 'Revert' button?

The easier option is 'yes' for both questions. In this case it's not necessary to update popup when the song is processed by pipeline after editing (tbh, I don't know how to do that).

alexesprit:
Revert is finally done + metadata info update on the fly.

PR #1210, opened 2017-01-01
Add anghami.com connector

ali-essam:
Connector for https://www.anghami.com/
anghami connector

### DIFF
--- a/connectors/v2/anghami.js
+++ b/connectors/v2/anghami.js
@@ -11,7 +11,7 @@ Connector.currentTimeSelector = '.position';
 
 Connector.durationSelector = '.duration';
 
-Connector.playButtonSelector = '.play-pause-button paused';
+Connector.playButtonSelector = '.play-pause-button.paused';
 
 Connector.playerSelector = '#player > ul';
 
--- a/connectors/v2/anghami.js
+++ b/connectors/v2/anghami.js
@@ -1,0 +1,21 @@
+'use strict';
+
+/* global Connector */
+
+
+Connector.artistSelector = '.track-artist';
+
+Connector.trackSelector = '.track-title';
+
+Connector.currentTimeSelector = '.position';
+
+Connector.durationSelector = '.duration';
+
+Connector.playButtonSelector = '.play-pause-button paused';
+
+Connector.playerSelector = '#player > ul';
+
+Connector.getTrackArt = function () {
+	// Remove image resize parameter
+	return $('.cover-art > img').attr('src').split('&')[0];
+};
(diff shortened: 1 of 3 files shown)


connectors/v2/anghami.js
```diff
@@ -0,0 +1,21 @@
...
+
+Connector.artistSelector = '.track-artist';
+
+Connector.trackSelector = '.track-title';
+
+Connector.currentTimeSelector = '.position';
+
+Connector.durationSelector = '.duration';
+
+Connector.playButtonSelector = '.play-pause-button paused';
```
alexesprit: Well, 'play button' is a button that means to be invisible when the song is playing. Moreover, the `.play-pause-button paused` selector is not valid. Scrobbling works fine in this case, the extension scrobbles paused songs too.

Also, '.play-pause-button' element has a class named 'paused' only if song is *paused*, so it's enough to check if 'paused' class is present, like this way:
```js
	return !$('.play-pause-button').hasClass('paused');
};
```

tests/connectors/anghami.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = function(driver, connectorSpec) {
+	connectorSpec.shouldBehaveLikeMusicSite(driver, {
+		url: 'https://www.anghami.com/explore',
+	});
```
alexesprit: The test should start to play the music by itself by specifying 'playButtonSelector' option as CSS selector of an element that will be clicked by WebDriver after URL load to play the music (if the music isn't started automatically). 
Since Anghami doesn't start music automatically, this test won't pass. Here is an example of correct test:
```js
'use strict';

module.exports = function(driver, connectorSpec) {
		url: 'https://www.anghami.com/tag/47',
		playButtonSelector: '.track-item .playicon'
	});
};
```
`.track-item .playicon` is just a play button of a first track on the page.

ali-essam: :+1: Yeah, you're right, seems like Anghami doesn't always play a song automatically on the explore page, but if you go to a specific song url, https://www.anghami.com/song/18025331 for example, the player will pop up and it will always play the song automatically, I tried it and it passes now.

connectors/v2/anghami.js
ali-essam: My bad, what I meant to do was `'.play-pause-button.paused'` instead ([selector chaining](http://stackoverflow.com/a/3772305/1499431)). I'm not exactly sure why it was working for me before, at least it didn't start scrobbling till I hit the play button even if the pause button was visible.

Anyway, I'm not sure if overriding `isPlaying` is better or going with chained classes selector is. I will go with the selector now, but let me know what you think.

**Paused**
anghami-paused

**Playing**
anghami-playing

alexesprit: Yes, you're right, the selector is preferred.

PR #1214, opened 2017-01-07
Implement dewtone.com connector

### LINKED ISSUE #1212: Scrobble for dewtone.com
They had file to download in .m3u once and i could scrobble it but sadly  possibility listen them is only by website now. Would be great if you can support dewtone.com and let us scrobble that site

ali-essam:
Scrobbling works fine in Chrome. But the test doesn't pass as the website does not get fully loaded in selenium webdriver, notice that there's no track info in selenium. I hope you can help with that issue, also let me know if you have any other comments

***Chrome***
dewtone connector

***Selenium Test***
dewtone selenium

resolves #1212

### DIFF
--- a/connectors/v2/dewtone.js
+++ b/connectors/v2/dewtone.js
@@ -1,0 +1,12 @@
+'use strict';
+
+/* global Connector */
+
+
+Connector.artistTrackSelector = '#footer-player-track';
+
+Connector.playerSelector = '#player';
+
+Connector.isPlaying = function () {
+	return $('#footer-player-play').css('display') !== 'block';
+};
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -1044,6 +1044,13 @@ define(function() {
 			matches: ['*://www.anghami.com/*'],
 			js: ['connectors/v2/anghami.js'],
 			version: 2
+		},
+
+		{
+			label: 'Dewtone',
+			matches: ['*://dewtone.com/*'],
+			js: ['connectors/v2/dewtone.js'],
+			version: 2
 		}
 	];
 });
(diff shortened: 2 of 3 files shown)


alexesprit:
The test passed for me: http://pastebin.com/raw/w3QJY537
But it's strange it doesn't pass for you.

Have you tried to run test in debug mode? Also, please check the developer console, there should be events sent by testReporter. Are they missing?

The connector works fine.

inverse:
Code wise looks fine, but the test isn't passing for me either.

I'm on Ubuntu also.

```
Running tasks: test:dewtone

Running "test:dewtone" (test) task

Running "exec:run_tests:dewtone" (exec) task
Verifying property exec.run_tests exists in config...OK
File: [no files]
Max stdout+stderr size is 204800 bytes

node tests/runner.js dewtone
Expecting exit code 0
1..1
```

alexesprit:
I reproduced the issue with test on Ubuntu too. Nothing happens even if I click on play button in Chrome, that run by chromedriver.

I don't know what should be done on the test core side, so I propose to check player selector only using `shouldContainPlayerElement` function.

inverse:
Other connectors are fine, so I wonder what the issue is with this one?

alexesprit:
@inverse yup. Can't even imagine why it works on Windows and doesn't work on Ubuntu.

ali-essam:
I tried debugging the site itself with the driver, and it seems like some of its init code doesn't get called when the page loads, but it gets called without the driver.

I guess I will go with `shouldContainPlayerElement` as @alexesprit suggested.

Awaku:
So is it working or not? I have windows and still does not work for me

alexesprit:
@Awaku it will be available in the next release. To get it worked right now you can install the extension as an unpacked one from the latest master.

PR #1224, opened 2017-01-14
amazon connector needs to decode HTML entity references in album title

mft:
album titles with punctuation or accent are encoded with HTML entity references.
artist names nor track titles are not.
I don't know why, but that's the way it is.

### DIFF
--- a/connectors/v2/amazon.js
+++ b/connectors/v2/amazon.js
@@ -1,6 +1,10 @@
 'use strict';
 
-/* global Connector */
+/* global Connector, MetadataFilter */
+
+Connector.filter = new MetadataFilter({
+	album: MetadataFilter.decodeHtmlEntities
+});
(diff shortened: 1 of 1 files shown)


alexesprit:
You omitted trim function in the filter configuration (that was used in default filter). Is it an intentional change?

mft:
No, it isn't.  What is the correct manner to add a filter with keeping the default?

alexesprit:
```js
Connector.filter = new MetadataFilter({
	album: MetadataFilter.decodeHtmlEntities,
	all: MetadataFilter.trim
 });
```
Artist and track will be filtered by 'trim' function only, whereas album will be filtered by both functions.

mft:
Thank you for the comment.
I've pushed the fix to include the trim filter.

PR #1233, opened 2017-01-26
Add Emby connector

kkopec:
Added connector for [Emby](https://emby.media/) media server.
Album names are recognized only on album view, because i dont think the necessary info is provided anywhere else.

Tested on Emby v3.1.2.0 and Chrome v55 / Opera v42

### DIFF
--- a/connectors/v2/emby.js
+++ b/connectors/v2/emby.js
@@ -9,7 +9,7 @@ Connector.artistSelector = '.nowPlayingBarText .textActionButton[data-type="Musi
 Connector.trackSelector = '.nowPlayingBarText .textActionButton[data-type="MusicAlbum"]';
 
 Connector.isPlaying = function () {
-	return $('.nowPlayingBarCenter .unpauseButton').hasClass('hide');
+	return $('.nowPlayingBarCenter .playPauseButton .md-icon').text() === 'pause';
 };
 
 Connector.getCurrentTime = function () {
--- a/connectors/v2/emby.js
+++ b/connectors/v2/emby.js
@@ -1,0 +1,42 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = '.nowPlayingBar';
+
+Connector.artistSelector = '.nowPlayingBarText .textActionButton[data-type="MusicArtist"]';
+
+Connector.trackSelector = '.nowPlayingBarText .textActionButton[data-type="MusicAlbum"]';
+
+Connector.isPlaying = function () {
+	return $('.nowPlayingBarCenter .unpauseButton').hasClass('hide');
+};
+
+Connector.getCurrentTime = function () {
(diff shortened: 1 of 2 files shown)


connectors/v2/emby.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+Connector.getUniqueID = function () {
+	var url = $('.nowPlayingBarInfoContainer .nowPlayingImage').css('background-image');
+	return /Items\/(\w+)/g.exec(url)[1];
+};
+
+Connector.isStateChangeAllowed = function () {
+	// avoids scrobble timer resetting on view change
+	return !this.getCurrentTime() > 0;
```
alexesprit: Is `not` operator supposed to be here? Looks like "don't change state if current time more that zero". Can't understand why this condition is used here.

kkopec: The condition was initially wrapped in parenthesis that i must have mistakenly deleted. I refactored it to be more clear.

alexesprit: I'm sorry, but I still can't understand the *reason* of use of this check. I guess the current time is mostly greater than zero, so the state of connector won't be changed.

kkopec: OK, let me explain the purpose of this condition.
In Emby you have several possible views while playing a song, one being the album wiev. The album information is available only on this particular one. Entering or leaving it while playing a song changes the album data in `currentState` object, therefore triggering change of state and forcing song recognition/scrobble timer resetting.
This condition prevents track state changes while it is playing, allowing it to be scrobbled regardless of view changes.

I refactored it again, allowing for state change on both the beginning and the end of the song. That way it is now possible to properly scrobble looped track.

alexesprit: Ok, now I got it.

alexesprit:
I'm testing right now the connector, and it didn't worked until I changed the `isPlaying` function:
```diff
diff --git a/connectors/v2/emby.js b/connectors/v2/emby.js
index 6fa3e210..e65421d5 100644
--- a/connectors/v2/emby.js
+++ b/connectors/v2/emby.js
@@ -9,7 +9,7 @@ Connector.artistSelector = '.nowPlayingBarText .textActionButton[data-type="Musi
 Connector.trackSelector = '.nowPlayingBarText .textActionButton[data-type="MusicAlbum"]';

 Connector.isPlaying = function () {
-       return $('.nowPlayingBarCenter .unpauseButton').hasClass('hide');
+       return $('.nowPlayingBarCenter .playPauseButton .md-icon').text() === 'pause';
 };

 Connector.getCurrentTime = function () {
```

Is this issue for me only? I'm using Emby 3.1.5.

kkopec:
No, it was the same for me after i have upgraded the server. The HTML structure changed a bit after the yesterday's update.

alexesprit:
👍 Thanks!

PR #1237, opened 2017-02-07
freegalmusic.com library service connector

reubot:
This adds a connector for the freegalmusic.com streaming site used by libraries.

### DIFF
--- a/connectors/v2/freegal.js
+++ b/connectors/v2/freegal.js
@@ -1,9 +1,17 @@
-'use strict';
+﻿'use strict';
 
 /* global Connector */
 
-Connector.playerSelector = '#custom-jw-player-wrapper';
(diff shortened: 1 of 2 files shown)


connectors/v2/freegal.js
```diff
@@ -0,0 +1,17 @@
...
+
+/* global Connector */
+
+Connector.playerSelector = '#custom-jw-player-wrapper';
+
+Connector.artistTrackSelector = '#jw-player-text';
+
+//Connector.playButtonSelector  = '#jw-play-button';
+Connector.isPlaying = function () {
+//    console.info($('#jw-play-button').text());
```
alexesprit: Is these a reason to keep commented lines of code?

reubot: No,  they can be removed.

alexesprit:
Thanks for contribution!

PR #1246, opened 2017-02-22
Pandora new website album tracking

JasonLucas:
I realized that the connector for the new website wasn't tracking albums and I tried fix it by my self, this looks like worked. I don't have much knowledge with coding stuff so, hopeful that's right.


alexesprit:
I intentionally didn't include album in the connector, because it's available on the main page only, whereas artist and track are on player bar, which is available everywhere.

alexesprit:
This change makes the controller to reset state on AJAX navigation since album info insn't available anywhere.

PR #1247, opened 2017-02-22
Adapt to Firefox WebExtensions API

AltarBeastiful:
Made a few adjustements to support Firefox browser.

Just renames except for `chrome.notifications.getPermissionLevel()` which isn't defined there.
I don't know if disabling an extension notification is possible in Firefox so I just stubbed the function.

After this PR we can create an account on addons.mozilla.org and submit the extension for review, see: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Publishing_your_WebExtension
## Test extension in Firefox
* open Firefox
* enter "about:debugging" in the URL bar
* click "Load Temporary Add-on"
* open the add-on's directory and select any file inside the add-on.

Source: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Temporary_Installation_in_Firefox

---
This change is [Reviewable](https://reviewable.io/reviews/david-sabata/web-scrobbler/1247)

### DIFF
--- a/core/background/main.js
+++ b/core/background/main.js
@@ -71,7 +71,7 @@ require([
 
 			if (!isActiveSession) {
 				isActiveSession = true;
-				GA.send('pageview', '/background-injected?version=' + chrome.app.getDetails().version);
+				GA.send('pageview', '/background-injected?version=' + chrome.runtime.getManifest().version);
 			}
 		}
 	};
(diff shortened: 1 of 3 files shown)


alexesprit:
This is already proposed in another OR, isn't it?

AltarBeastiful:
Oops, it appears it is. Closing the request then.

PR #1248, opened 2017-02-25
Fair Price Music v2 connector

mrweiss:
Connector for the www.fairpricemusic.com. 

Handles all three pages - playlist/band/album.

Formatting changes for the connectors.js unless anyone objects -> shouldn't we standardize on 2 tabs for all files anyway? :)

Cheers

### DIFF
--- a/connectors/v2/fairpricemusic.js
+++ b/connectors/v2/fairpricemusic.js
@@ -1,0 +1,83 @@
+'use strict';
+
+/* global Connector */
+
+function setupPlaylistPlayer() {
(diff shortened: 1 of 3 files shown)


alexesprit:
@mrweiss The connector code looks good. Could you extend test to cover album and band pages?

mrweiss:
@alexesprit  I almost did that, but then realized - won't the protractor/webdriver tests just increase the page/play hits for those bands/albums? I suppose some bands might want be keeping stats :)

PR #1272, opened 2017-04-03
Play detection to simple language-agnostic selector, times, picture.

### LINKED ISSUE #1270: open.spotify.com data is fetched but (still) won't scrobble
### Expected Behavior
Well, `testReporter.js` outputs data objects in the console and the same data should be scrobbled. I have installed the latest from the master branch, which includes 1b2b536.
### Current Behavior
`testReporter.js` outputs data objects in the console but no data is actually scrobbled to last.fm. This issue might be similar to #1267.

### Steps to Reproduce (for bugs)
1. Download latest master version from git
2. Activate extension.
3. Uncheck every site _except Spotify (open)_.
4. Re-authenticate with Last.FM.
5. Go to https://open.spotify.com and hit **Ctrl-Shift+R**.
6. Extension recognizes open.spotify.com but nothing happens.

Pastebins of with console output of:
 - Extension: https://pastebin.com/raw/XtF4CQeN
 - open.spotify.com: https://pastebin.com/raw/iV9ZBXqJ

### Context
I'm trying to scrobble my lovely music taste.

### Steps for resolving the issue
- [x] Reauthenticate on Last.fm
- [x] Try to use unpacked extension from latest master
- [x] Check other issues for duplicates

### Environment

* Version used: The unpacked version from master branch gives **1.62.0** but the extension from the web store gives **1.62.2**.
* Browser Name and version: Google Chrome **58.0.3029.31** (Official Build) dev (64-bit)

myfonj:
Fixes #1270 on My machine™. `open.spotify.com` is now localised so it is not possible to rely on `title` attribute values.

### DIFF
--- a/connectors/v2/spotify-open.js
+++ b/connectors/v2/spotify-open.js
@@ -1,14 +1,14 @@
 'use strict';
 
-/* global Connector, Util */
+/* global Connector, MetadataFilter, Util */
 
 /**
  * The connector for new version of Spotify (open.spotify.com).
  */
 
-Connector.playerSelector = '.nowPlayingBar-container';
+Connector.playerSelector = '.now-playing-bar';
 
-Connector.getArtist = function() {
+Connector.getArtist = function () {
 	let artists = $('.track-info__artists a').toArray();
 	return Util.joinArtists(artists);
 };
@@ -22,3 +22,5 @@ Connector.playButtonSelector = '.control-button[class*="spoticon-play-"]';
 Connector.currentTimeSelector = '.player-controls__progress-time:first-child';
 
 Connector.durationSelector = '.player-controls__progress-time:last-child';
+
+Connector.filter = MetadataFilter.getRemasteredFilter();
--- a/connectors/v2/spotify-open.js
+++ b/connectors/v2/spotify-open.js
@@ -6,7 +6,7 @@
  * The connector for new version of Spotify (open.spotify.com).
  */
 
-Connector.playerSelector = '.now-playing-bar';
+Connector.playerSelector = '.nowPlayingBar-container';
 
 Connector.getArtist = function() {
 	let artists = $('.track-info__artists a').toArray();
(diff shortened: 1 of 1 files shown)


alexesprit:
Why did you remove the filter?

connectors/v2/spotify-open.js
```diff
@@ -9,16 +9,16 @@
...
-Connector.trackSelector = '.now-playing-bar div > div > div > a';
+Connector.trackSelector = '.track-info__name a';
 
-Connector.trackArtImageSelector = '.now-playing-bar .cover-art-image-loaded';
+Connector.trackArtImageSelector = '.now-playing__cover-art .cover-art-image-loaded';
 
-Connector.isPlaying = function() {
-	return $('.control-button[title="Pause"]').length > 0;
-};
+Connector.playButtonSelector = '.control-button[class^="spoticon-pause-"]';
```
alexesprit: This change makes the connector think the song is playing while it is actually paused and vice versa. Please use `spoticon-play-*` selector to avoid this behaviour.

myfonj: You are right, totally messed up in the end and tested just scrobbling and track detection, sorry. Pushed corrected version, but not yet thoroughly tested.

connectors/v2/spotify-open.js
```diff
@@ -1,24 +1,24 @@
...
+Connector.trackArtImageSelector = '.now-playing__cover-art .cover-art-image-loaded';
 
-Connector.isPlaying = function() {
-	return $('.control-button[title="Pause"]').length > 0;
-};
+Connector.playButtonSelector = '.control-button[class*="spoticon-play-"]';
+
+Connector.currentTimeSelector = '.player-controls__progress-time:first-child';
 
-Connector.filter = MetadataFilter.getRemasteredFilter();
```
alexesprit: Is there a reason to remove the metadata filter?

myfonj:
As so "remastered filter": in fact it seemed quite redundant to me at first and I'm not quite convinced it is a good thing to do generally. Year and "remastered" information are still data one could care about, and FMoPW should be post-processed (if at all) in the data centre. 

Who knows how many ELP fans will carefully distinguish between various remasters https://www.last.fm/music/Emerson,+Lake+&+Palmer/+tracks?date_preset=LAST_7_DAYS

PR #1274, opened 2017-04-05
updated IDs

andrewklimek:
I wrote a track info script today so I could do away with centova casts's script and its jQuery requirement.  No longer using the centova cast ID names


alexesprit:
Thanks! 👍

PR #1294, opened 2017-04-26
Listenonrepeat

### LINKED ISSUE #1293: [REQUEST] Add support for https://listenonrepeat.com
Add scrobbling support for the website:
https://listenonrepeat.com

inverse:
Resolves #1293

### DIFF
--- a/connectors/listenonrepeat.js
+++ b/connectors/listenonrepeat.js
@@ -1,12 +1,19 @@
 'use strict';
 
-/* global Connector */
+/* global Connector Util */
 
 Connector.playerSelector = '.player-controls';
 
-Connector.artistTrackSelector = '.player-controls .video-title';
+Connector.artistTrackSelector = '.player-controls > .video-title:first';
 
 Connector.playButtonSelector = '.controls > button:nth-child(4)';
+
+Connector.getArtistTrack = function () {
+	var text = $(Connector.artistTrackSelector).text();
+
+	console.log(text);
+	return Util.processYoutubeVideoTitle(text);
+};
 
 Connector.isPlaying = function() {
 	return $('button:contains(Pause Video)').length > 0;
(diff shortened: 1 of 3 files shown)


connectors/listenonrepeat.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+/* global Connector Util */
+
+Connector.playerSelector = '.player-controls';
+
+Connector.artistTrackSelector = '.player-controls > .video-title:first';
+
+Connector.playButtonSelector = '.controls > button:nth-child(4)';
```
alexesprit: This is redundant, since `isPlaying` function is used.

alexesprit:
I suggest to watch for changes of repeat counter and reset state using `Connector.resetState` function on every change. This approach will allow to support repeating videos.

inverse:
Good call - as you can see it's been a while since I touched this ;) Will update tomorrow or feel free to push to the branch if you wanna get this merged today.

inverse:
@alexesprit what would you use for the selector for the repeat counter? I could use the one generated by chrome but it feels very strict 

```
#page-home > div:nth-child(2) > div:nth-child(3) > div > div:nth-child(3) > div > div:nth-child(4) > div > div:nth-child(1)
```

Also would you attach this with a mutation observer?

alexesprit:
@inverse Looks like the repeat counter is not available when the connector is injected. We should wait until the counter element is visible using another MutationObserver. I'll try to implement something like that later.

> #page-home > div:nth-child(2) > div:nth-child(3) > div > div:nth-child(3) > div > div:nth-child(4) > div > div:nth-child(1)

That's no good. I think we could use `.hide-on-small-only svg + span` or parent of parent of `.hide-on-small-only #Icon` but they're no much better than generated one by Chrome.

alexesprit:
I had no luck with adding of repeat support, so I think the connector works as much as it can.

f1u77y:
@alexesprit 
> repeat support

Couldn't we just inject code into the iframe and use existing YT code that should support repeat?

alexesprit:
@f1u77y AFAIK we are not able to do it due to same-origin security policy.

inverse:
That sucks :( @alexesprit what issue did you have? getting access to the listen counter? 

I wish they had nicer semantics in their DOM.

f1u77y:
@alexesprit we can inject code that will interact with main frame via `window.postMessage()`

PR #1298, opened 2017-04-29
Youtube: check for Music category in new design

f1u77y:
What's with YT API key? I'm using mine in this PR now but ideally we should use key which belongs to someone who contributes more often or to the repo owner.

### DIFF
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -227,12 +227,15 @@ function setupBasePlayer() {
 			return false;
 		}
 
+		if (!scrobbleMusicOnly) {
+			return true;
+		}
+
 		const params = new URLSearchParams(document.location.search.substring(1));
 		const videoId = params.get('v');
 		let videoCategory = getVideoCategory(videoId);
-		if (videoCategory) {
-			let result = !scrobbleMusicOnly || (scrobbleMusicOnly && videoCategory === CATEGORY_MUSIC);
-			return result;
+		if (videoCategory !== null) {
+			return videoCategory === CATEGORY_MUSIC;
 		}
 
 		return false;
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -88,7 +88,7 @@ function setupDefaultPlayer() {
 }
 
 let categoryCache = new Map();
-const CATEGORY_MUSIC = 10;
+const CATEGORY_MUSIC = '10';
 const YT_API_KEY = 'AIzaSyApDdf5_bvYEq0CjiwKJ_VRo3CrAT3HscQ';
 
 function getVideoCategory(videoId) {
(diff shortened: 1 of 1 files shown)


alexesprit:
I suggest to use this approach in both players.

connectors/youtube.js
```diff
@@ -124,7 +144,19 @@ function setupMaterialPlayer() {
 	};
 
 	Connector.isStateChangeAllowed = function() {
-		return $('.videoAdUi').length === 0;
+		if ($('.videoAdUi').length > 0) {
+			return false;
+		}
+
+		const params = new URLSearchParams(document.location.search.substring(1));
+		const videoId = params.get('v');
```
alexesprit: Video ID is also accessible via `Connector.getUniqueID` function. I think it should be used instead of this.

f1u77y: Yes, it is. Will change now.

(BTW, `Util.getYoutubeVideoIdFromUrl` is using regex for extracting ID. Isn't this way of getting `v` parameter a bit more reliable?)

f1u77y:
> I suggest to use this approach in both players

Yes, I think it's better too.

connectors/youtube.js
alexesprit: @f1u77y yes, using `URLSearchParams` in that function is fine. Feel free to create another PR for this.

f1u77y:
@alexesprit done

alexesprit:
@f1u77y awesome, thanks. I'll test changes later.

Btw, do you know if there're some Youtube API limitations, e.g. number of calls per day etc?

f1u77y:
> Youtube API limitations

Approximately:

* 333333 queries per day
* 1e6 queries per 100 seconds(I don't know wtf is that but my "Quotas" page says that)
* 1e5 queries per 100 seconds per user

I think that's more than enough.

alexesprit:
@f1u77y I found a minor issue:
1. Enable "Scrobble only Music category" option.
2. Play some video from "Music category" and wait until the extension status in the extension bar is "now playing"
3. Select another video from non-music category.

Expected: the status is reset.
Actual: the status is still "now playing".

This is not a regression, the connector from master has the same behaviour. Have you any ideas how to avoid this?

alexesprit:
@Vankog JSON object can contain several data types according to JSON specification.

f1u77y:
@alexesprit this bug is caused by not changing state when navigating to non-music video. The only solution I see requires changing of `Connector` API. We should use `Connector.isScrobblingAllowed()` instead of `Connector.isStateChangeAllowed()` and stop scrobbling immidiately when scrobbling becomes not allowed. Not sure if this change is the best what we can.

alexesprit:
@f1u77y I noticed the music isn't scrobbled if an error is occurred during retrieving the video category.

alexesprit:
image
That's really strange. @f1u77y do you know why does it happen?

PR #1302, opened 2017-05-02
Add radio-mb connector

AhmedKamal20:
add a connector for https://beta.radio-mb.com/

### DIFF
--- a/connectors/radio-mb.js
+++ b/connectors/radio-mb.js
@@ -1,0 +1,14 @@
+'use strict';
+
+/* global Connector */
+
+
+Connector.playerSelector = '#radiomb';
+Connector.artistSelector = '#radiomb-np-artist';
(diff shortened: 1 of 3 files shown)


alexesprit:
I changed the test this way:
```js
'use strict';

module.exports = function(driver, connectorSpec) {
		url: 'http://beta.radio-mb.com/',
		playButtonSelector: '.fa-play-circle'
	});
};
```
and it's passed for me. Could you check it too and confirm if it's passed for you?

The connector works fine, but on website "Unknown Artist - Unknown Track" is always playing (perhaps, it's due to beta status of web service).

AhmedKamal20:
yeah , it passed for me too, thank you 

this work only in http://beta.radio-mb.com/ after pressing  'Play Radio' button 
will not work on http://www.radio-mb.com/ for , until we update it

PR #1307, opened 2017-05-07
Added lofi-radio.co connector

jakeoid:
Adding a basic connector for the lofi-radio website!


jakeoid:
👌

jakeoid:
That should fix things. 🤔

alexesprit:
Yes, now test is passed for me.

> Also, please add an URL pattern w/o "www".

Well, I meant to add the pattern and keep the one with "www", like:
```js
	{
		label: 'Lofi Radio',
		matches: ['*://lofi-radio.co/*', '*://www.lofi-radio.co/*'],
		js: ['connectors/lofi-radio.js'],
	},
```

The extension doesn't recognize the website in certain case without this change (I type "lofi-radio.co" in the adress bar and Chrome redirects to http://www.lofi-radio.co, but the website is also available at http://lofi-radio.co).

jakeoid:
oh I was certain I could do that, two moments.

alexesprit:
Thanks 👍

PR #1310, opened 2017-05-10
New connectors: "radioeins" and "Fritz"

carpet-berlin:
The player used by "radioeins" and "Fritz" is flash based, therefore it
is a bit complicated to get its state from inside our content script.
The function retrieveJwplayerState() injects code which gets the player
state and stores it in a document attribute (body.tmp_state), which then
can be read by our content script.

### DIFF
--- a/connectors/fritz.js
+++ b/connectors/fritz.js
@@ -1,0 +1,32 @@
+'use strict';
+
+/* global Connector */
+
+Connector.playerSelector = 'div.layoutnowonair_reload_active';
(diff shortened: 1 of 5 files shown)


alexesprit:
I suggest to change injected function to:
```js
if (typeof jwplayer === "function") $("body").attr("tmp_state", typeof jwplayer().getState === "function" ? jwplayer().getState() : "");
```
and remove `result` variable from `Connector.isPlaying` function.

PR #1322, opened 2017-05-30
Fixing Spotify Open's functionality for retrieving album titles.

StephenDonaldson:
Fixing Spotify Open's lack of sending album titles along with scrobbles as per [issue 1320](https://github.com/david-sabata/web-scrobbler/issues/1320) that I opened.

This fix uses a workaround to grab the album title of the currently playing track from an album page, single page, artist page, or playlist via sending an AJAX request to the hidden album url in the new Now Playing section and returning the title from the AJAX request's metadata. It would be more straightforward to use Spotify's API, but they've just turned off all un-authenticated API requests as of 2 days ago. This fix works on all areas of the site and scrobbles properly to Last.FM as well.

### DIFF
--- a/connectors/spotify-open.js
+++ b/connectors/spotify-open.js
@@ -5,6 +5,8 @@
 /**
  * The connector for new version of Spotify (open.spotify.com).
  */
+ 
+var albumTitle;
 
 Connector.playerSelector = '.now-playing-bar';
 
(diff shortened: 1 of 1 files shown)


alexesprit:
1. The connector send AJAX request on every `getAlbum` function call (approx. every 0,5 sec).
2. The changed code produces an error:
```
spotify-open.js:20 Uncaught TypeError: Cannot read property 'href' of undefined
    at BaseConnector.Connector.getAlbum (spotify-open.js:20)
    at BaseConnector.stateChangedWorker (connector.js:386)
    at MutationObserver.onStateChanged (connector.js:482)
```

StephenDonaldson:
The first one I'll look into a fix for, there's not much that I can really think of though since it needs to be able to update. I'll try to switch so that it only makes the ajax call when there's a player state change.

As for the second I'll add a fallback for that; it's throwing that error when the page first loads since nothing is playing and those divs have not been written yet.

PR #1331, opened 2017-06-14
Add QQ Video support

usdivad:
This adds support for scrobbling music videos on [QQ Video](https://v.qq.com/).

Examples:
https://v.qq.com/x/cover/zz4s3deuelkwbdk/t0023et2neq.html
https://v.qq.com/x/page/y0142cniol3.html

(More examples can be found in QQ Video's [Music](http://v.qq.com/music/) section.)

Things to review:
- [ ] I renamed `connectors/qq.js` to [connectors/qq-music.js](connectors/qq-music.js) and `tests/connectors/qq.js` to [tests/connectors/qq-music.js](tests/connectors/qq-music.js), as it seemed the best way to distinguish between QQ Music and QQ Video. I'm definitely open to other ways of naming it though.
- [ ] I wasn't sure whether or not to add `version: 2` for the connector definition in [core/connectors.js](core/connectors.js); right now it's not in the definition.
- [ ] In line 34 of [connectors/qq-video.js](connectors/qq-video.js) I define a `Connector.splitArtistTrack` function to parse the title string. Perhaps there's a better place to do this?


alexesprit:
Thanks for contribution!

1. It's totally fine.
2. The 'version' property is not used, because v1 core was removed. All connectors supposed to not have this property.
3. It's valid, but I'd define this function in the global scope. There's no sense to define it as the `Connector` object member.

I also noticed you used RegEx to get track name and `String.split` to get artist name. How do you think is it better to combine these things by using RegEx only? For example:
```js
const artistTrackRe = /(.+?)\s《(.+?)》/

function extractArtistTrack(artistTrackStr) {
	let match = artistTrackStr.match(artistTrackRe);
	if (match) {
		return { artist: match[1], track: match[2] };
	}

	return Util.emptyArtistTrack;
}
```

usdivad:
Thanks for the feedback, Alexey!

I think using RegEx for both artist and track is a great idea, as is defining the `splitArtistTrack` function in the global scope. I'll add those changes to this branch.

PR #1351, opened 2017-07-02
Enhance YouTube options

### LINKED ISSUE #1350: Misleading YouTube option names
If both options are disabled, we scrobble any category but user could misunderstand option titles (see [comments](https://github.com/david-sabata/web-scrobbler/issues/1349#issuecomment-312450318)). 

I'd like to suggest two options:
- One
  - Scrobble Music
  - Scrobble Entertainmaent
  - Scrobble other
- Two
  - Scrobble only these categories
    - Music
    - Entertainment
  - Scrobble all

f1u77y:
Don't use misleading option names.

Closes #1350.

### DIFF
--- a/connectors/youtube.js
+++ b/connectors/youtube.js
@@ -307,6 +307,9 @@ function readConnectorOptions() {
 			if (options.scrobbleEntertainmentOnly) {
 				allowedCategories.push(CATEGORY_ENTERTAINMENT);
 			}
+			if (options.scrobbleAll) {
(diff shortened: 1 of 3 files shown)


alexesprit:
@f1u77y If I disable all three checkboxes, that means the Youtube connector is turned off, right? I suppose it should be turned off in this case, but the connector still works.

inverse:
Should we cater for this? having a disabled state for connectors. Would make this interaction nicer.

Issue #1348, opened 2017-06-27
Ability to view list of saved edited tracks


fishstik:
I love the feature that saves edited track info and autofills it later. Very useful! Would like the ability to see (and maybe even edit?) a list of all my edited tracks. Could even just be a .txt file that could be opened from the Options page.

alexesprit:
@fishstik In the next release you can view edited tracks. I don't think editing the data should be implemented; this action can be done using the now playing popup. Removing tracks is not supported too, but I think it's a good feature should be implemented (possibly, later).

fishstik:
Awesome work, thanks!

PR #1352, opened 2017-07-02
View edited tracks in options

inverse:
Still WIP but wanted some input.

refs #1348

### DIFF
--- a/options/options.js
+++ b/options/options.js
@@ -226,7 +226,8 @@ require([
 					cache.append($('<li>').text('No items in the cache.'));
 				} else {
 					$.each(data, function(cachedKey, cachedData) {
-						cache.append($('<li>').text(cachedData.artist + ' - ' + cachedData.track));
+						let { artist, track } = cachedData;
+						cache.append($('<li>').text(`${artist} - ${track}`));
 					});
 				}
 			});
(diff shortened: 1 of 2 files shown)


options/options.js
```diff
@@ -214,6 +215,29 @@ require([
...
+
+			const localCache = ChromeStorage.getNamespace('LocalCache');
+
+			let cache = $('<ul class="list-unstyled"></ul>');
+			localCache.get((data) => {
+				if (Object.keys(data).length === 0) {
+					cache.append($('<li>').text('No items in the cache.'));
+				} else {
+					$.each(data, function(cachedKey, cachedData) {
+						cache.append($('<li>').text(cachedData.artist + ' - ' + cachedData.track));
```
alexesprit: Can we use template literals in this code block?
```js
let { artist, track } = cachedData;
cache.append($('<li>').text(`${artist} - ${track});
```

alexesprit:
What about removing items from the cache? Don't know exactly how to make this feature to look better; I suggest to add an "X" icon right after an item text.

options/options.js
inverse: We could try it :+1:

inverse:
We could always iterate on this :+1:

inverse:
@alexesprit should we merge this and iterate?

PR #1356, opened 2017-07-06
Add Nightcore connector

### LINKED ISSUE #1355: Connector Request: Nightcore.com
After searching around I can't find any scrobbler that supports this site.
https://www.nightcore.com

Been using this addon for a long time so figured I would come  here to see if it is possible to add this site to the list of supported sites.

Not sure if it is possible since it uses embedded Youtube videos.

f1u77y:
Resolves #1355

### DIFF
--- a/connectors/nightcore.js
+++ b/connectors/nightcore.js
@@ -1,0 +1,51 @@
+'use strict';
+
+/* global Connector */
+/* global MetadataFilter */
+
+if (window.frames.length === 0) {
+	// We are in YouTube frame
+	// Overriding artist and track getters because the website contains
+	// better metadata
+
+	let artist = null;
+	let track = null;
+
+	Connector.getArtist = () => artist;
+	Connector.getTrack = () => track;
+	Connector.getArtistTrack = () => {};
+
+	$('video').on('timeupdate', () => {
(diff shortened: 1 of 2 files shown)


alexesprit:
I'm not sure the Youtube connector should be used for Nightcore. It always uses "Nightcore" as an artist name, whereas the artist name is in another element.

f1u77y:
So I think we should take artist and title from the website but leave everything else to the YT connector, shouldn't we?

(And I couldn't find any other info on the website)

connectors/nightcore.js
```diff
@@ -0,0 +1,51 @@
...
+			return;
+		}
+		artist = event.data.artist;
+		track = event.data.track;
+	});
+
+	Connector.filter = new MetadataFilter({
+		artist: text => text.replace(/\s*original artist:/i, ''),
+		track: text => text.replace(/^nightcore - /i, ''),
+	});
```
alexesprit: The filter doesn't trim song fields, but it should. You can use default `TrimFilter` with appended custom functions (see `MetadataFilter.append` function).

inverse:
@f1u77y any chance for an update on this?

alexesprit:
Closed due to inactivity.

PR #1362, opened 2017-07-27
Separate Hype Machine connectors.

jaccarmac:
This allows one to use Hype Machine's built-in scrobbling while still being
able to scrobble premieres.


inverse:
Could we add a test?

jaccarmac:
Gotcha.

PR #1370, opened 2017-08-03
Add a connector for beta TuneIn (@beta.tunein.com)

eugenesvk:
Hey, thanks for the great addon! [TuneIn radio](https://beta.tunein.com) changed their interface in the new (beta) version, so the old connector stopped working. I've looked through the code on the new page and made a new connector, which does work with the new version.

I have just one question, though. Given that `div classes` have strange names like '.player__playerContainer___JEJ2U', which I'm not sure will remain unchanged relative to the more generic looking `div ids`, is there a way to refer to th

At the moment this only affects the following piece of code (not sure how important it is) since it has a `data-testid` instead of the regular `div id` (like other items) and I don't know how to write a proper condition referencing such `data` elements.

```
Connector.playerSelector = '.player__playerContainer___JEJ2U';
// Connector.playerSelector = '[data-testid]="player"'; // Alternative with div id. Not working due to data class, div id #player is another element
```

### DIFF
--- a/connectors/tunein-beta.js
+++ b/connectors/tunein-beta.js
@@ -2,17 +2,9 @@
 
 /* global Connector */
 
-// Full player container (bar at the bottom)	: [data-testid]="player"#player	.player__playerContainer___JEJ2U
-//   'Currently playing' container          	: #nowPlayingContainer         	.player__nowPlayingContainer___2FBjw
-
 Connector.playerSelector = '.player__playerContainer___JEJ2U';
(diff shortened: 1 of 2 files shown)


alexesprit:
> I have just one question, though. Given that div classes have strange names like '.player__playerContainer___JEJ2U', which I'm not sure will remain unchanged relative to the more generic looking div ids, is there a way to refer to th

Looks like we have no better solution for this, so it's totally ok.

Can you remove commented code from the connector file? Everything else looks good to me.

eugenesvk:
@alexesprit ok, understood, hope they don't change JEJ2U to JEJ3U too soon :)
Also, removed the commented code

PR #1380, opened 2017-08-24
Add cache clear

inverse:
- Add the ability to clear all the cache of stored tracks
- Make labels consistent


alexesprit:
I suggest to change style of the clear button to make it look like this one:
image

inverse:
Good call!

PR #1404, opened 2017-09-19
Add JetSetRadio.live connector

rrix:
It's a little .. awkward since there's not a player in the strictest sense of the word here, just a next button. I'm relying on the text of the track not changing except between tracks here, which I think is the correct behavior from my watching last.fm. 

I tested this in Firefox


alexesprit:
You can omit `isPlaying` implementation. The extension always assumes the song is playing by default.

alexesprit:
Nice, thanks!

PR #1410, opened 2017-09-24
Add feature to swap title and artist

Azenet:
I mainly use this extension with YouTube, which contains many videos that have their title and artist reverted. This change adds a "Swap" button to the info popup when editing the metadata. The button swaps the title and the artist, then submits the correction.


alexesprit:
Awesome, thanks!

PR #1413, opened 2017-09-25
Update yandex-music.js

WhyamIdointhat:
Just want to add album's name in scrobble.

### DIFF
--- a/connectors/yandex-music.js
+++ b/connectors/yandex-music.js
@@ -4,6 +4,8 @@ Connector.playerSelector = '.player-controls';
 
 Connector.artistSelector = '.track_type_player .track__artists';
 
+Connector.albumSelector = '.album_type_player .track__album';
+
 Connector.trackSelector = '.track_type_player .track__title';
 
 Connector.currentTimeSelector = '.progress__left';


alexesprit:
I found no elements by such CSS selector. Have you any example of URL where album info is available?

WhyamIdointhat:
So extension itself already scrobble album name, but time to time.
On this page album name is correct
https://music.yandex.ua/album/1611922
But here no album name at all
https://music.yandex.ua/album/2970031

alexesprit:
> So extension itself already scrobble album name, but time to time.

I suppose the extension just grabs album name using Last.fm API. I still see no elements having `.album_type_player .track__album` selector.

xyauhideto:
I had a try and would conclude that the general way to get the album title of the now-playing track is using  `XHR`.

``` javascript
Connector.getAlbum = () => {
	let url = $('track_type_player .track__cover').attr('href'),
		album = null;
	$.ajax({
		url,
		type: 'get',
		async: false,
		success: (data) => album = $(data).find('h1').text(),
    });
	return album;
}
```

But it will keep fetching even after getting the result. I wonder if it is worth doing.

Btw I found the `trackArt`.
```javascript
    let text = $('.track_type_player .track-cover').attr('srcset');
    return 'http:' + text.split(' ').shift();
};
```

alexesprit:
> I had a try and would conclude that the general way to get the album title of the now-playing track is using XHR.

Well, the extension will execute this code too often (about every 0,5...1 sec), so this is not acceptable.

PR #1417, opened 2017-10-03
Add WFMU connector

surfinzap:
I'd like to propose to add the support for WFMU, independent freeform radio from New York. 
Code-wise, I'm not sure, whether I have written tests the right way. Thanks for any help!


alexesprit:
Pretty interesting way to extract track info from a text string.

Could you rebase your branch to resolve merge conflicts?

PR #1432, opened 2017-10-10
163, QQ, Baidu, xiami connectors: Get unique ID

xyauhideto:
_Visits outside mainland China are (partialy) blocked by the sites, you may need VPN extensions to unlock them._

### DIFF
--- a/connectors/163-music.js
+++ b/connectors/163-music.js
@@ -4,7 +4,10 @@ const trackRegEx = new RegExp(String.fromCharCode(160), 'g');
 
 Connector.playerSelector = '.m-playbar';
 
-Connector.trackArtSelector = '.head.j-flag img';
+Connector.getTrackArt = () => {
+	let src = $('.head.j-flag img').attr('src');
+	return src.split('?').shift();
+};
 
 Connector.getTrack = () => {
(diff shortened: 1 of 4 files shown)


connectors/163-music.js
```diff
@@ -4,7 +4,10 @@ const trackRegEx = new RegExp(String.fromCharCode(160), 'g');
 
 Connector.playerSelector = '.m-playbar';
 
-Connector.trackArtSelector = '.head.j-flag img';
+Connector.getTrackArt = () => {
+	let src = $('.head.j-flag img').attr('src');
+	return src.substr(0, src.length - 12);
```
alexesprit: What does original `src` string contain?

xyauhideto: Something like this:
`http://p1.music.126.net/Gd-HAk9hKC85L0wNtfRs1g==/7946170535396804.jpg?param=34y34`
with a parameter to modify the size of the image

alexesprit: Could you remove the parameter in a way you did it in 39fd9295010e202838257703ad743dec205d4b0e commit? I think it's more clear.

PR #1435, opened 2017-10-14
updated youtubify and added Ma5onic Music Player *correction*

Ma5onic:
I added *://bemusic.vebto.com/* as one of the matches to youtubify since it is now hosted on a different subdomain.
I also added my website's music player (Ma5onic Music Player) which uses the same "connectors/" as Youtubify.

### DIFF
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -737,10 +737,17 @@ define(function() {
 
 	{
 		label: 'Youtubify',
-		matches: ['*://youtubify.vebto.com/*'],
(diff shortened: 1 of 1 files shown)


alexesprit:
Could you also update test file located at `tests/connectors/bemusic.js`?

alexesprit:
Updated tests.

I also noticed BeMusic connector doesn't work on new Youtubify website and on your one. Looks like they changed something inside the engine.

PR #1440, opened 2017-10-18
Use data() and val() APIs

xyauhideto:
For di.fm, use the selector instead

---
The `MusicPlayer` setup for [obozrevatel](http://radio.obozrevatel.com/) also uses `attr('data-*')`, but I can't find an instance to test the `data()` API, so it is not included.

### DIFF
--- a/connectors/1ting.js
+++ b/connectors/1ting.js
@@ -21,7 +21,7 @@ function isH5Player() {
 function setupDefaultPlayer() {
 	Connector.playerSelector = '.playact';
 
-	Connector.getUniqueID = () => $('.playing').attr('id').substr(3);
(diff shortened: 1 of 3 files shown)


connectors/di.js
```diff
@@ -27,6 +27,4 @@ Connector.getTrackArt = () => {
 	return null;
 };
 
-Connector.isPlaying = () => {
-	return $('#webplayer-region').attr('data-state') === 'playing';
-};
+Connector.playButtonSelector = '.icon-play';
```
alexesprit: Could you revert this change? Now the extension recognizes now playing track as unknown one while loading.

PR #1444, opened 2017-10-21
PLEX connector: add new web pattern; add track art

### LINKED ISSUE #1442: Plex Scrobbling not working
### Expected behavior
I've activated Scrobbling for PLEX on the Plugin settings... but nothing happens.
And this Problem is present on app.plex.tv and on the local server IP.

### Current behavior
Maybe there is a Problem with the Regex... because the Title is shown in the Browser-Tab!
If so you may can help me... maybe I can fix it by my self with some help.

### Possible solution
Maybe it will help when the Regex of the Plex detection will be overdone some time soon...

### Steps to reproduce (for bugs)
Can't really reproduce this... but 
image
will maybe describe it better....

### Environment

Browser Version: Chrome: Version 61.0.3163.100 (Offizieller Build) (64-Bit)
Plex-Server: Version 1.9.2.4285 (Running on a Freenas Jail)
Web-Scrobbler: 2.1.0

xyauhideto:
Might fixes #1442 

Also add video title parser.
Seemly the remote UI and the local UI are unfied, and the remote URL has
changed. Anyways, I just change the condition and add some functions.

### DIFF
--- a/connectors/plex.js
+++ b/connectors/plex.js
@@ -18,7 +18,7 @@ function setupConnector() {
  * @return {Boolean} Check result
  */
 function isRemoteUI() {
-	return location.hostname === 'app.plex.tv';
+	return $('[class^=AudioVideoPlayerView-container]').length > 0;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


connectors/plex.js
```diff
@@ -46,14 +46,27 @@ function setupLocalUI() {
...
 	Connector.trackSelector = `${Connector.playerSelector} [class*=titlesContainer] > a`;
 
+	// for watch-it-later videos
+	Connector.artistTrackSelector = `${Connector.playerSelector} [class*=MetadataPosterTitle-title]`;
+
+	Connector.getTrack = () => {
+		if (Connector.getArtist()) {
+			return $(Connector.trackSelector).text();
+		}
+		Connector.filter = MetadataFilter.getYoutubeFilter();
```
alexesprit: Is there a way to not create new filter object on every `getTrack` call?

inverse: Can't this be done outside the getTrack scope?

xyauhideto:
I considered in the way that the filter should only be applied when there's
no track art. I will edit it.

2017年10月21日 12:34，"Malachi Soord" <notifications@github.com>写道：

core/connectors.js
```diff
@@ -608,6 +608,12 @@ define(function() {
 	},
 
 	{
+		label: 'post-player',
+		matches: ['*://post-player.org/*'],
+		js: ['connectors/post-player.js'],
+	},
```
alexesprit: post-player is removed, isn't it?

connectors/plex.js
```diff
@@ -46,14 +46,27 @@ function setupLocalUI() {
...
+	// for watch-it-later videos
+	Connector.artistTrackSelector = `${Connector.playerSelector} [class*=MetadataPosterTitle-title]`;
+
+	Connector.filter = MetadataFilter.getYoutubeFilter();
+
+	Connector.getTrack = () => {
+		if (Connector.getArtist()) {
+			return $(Connector.trackSelector).text();
+		}
+	};
```
alexesprit: It's should be `return null` in the end of this block.

Issue #1445, opened 2017-10-21
Connection Request: scrobble from ccmixter.org


douglasawh:
I'd like to be able to scrobble from http://ccmixter.org/

### Expected behavior

Just an additional site request

### Current behavior

It just doesn't do it right now.

### Context

When i got to a page such http://ccmixter.org/people/Reiswerk I'd like to be able to scrobble those tracks.
There are also pages like http://ccmixter.org/view/media/remix/picks that have a slightly different format

alexesprit:
The extension now supports ccMixter scrobbling, thanks to @xyauhideto.

PR #1446, opened 2017-10-22
ccmixter connector

xyauhideto:
#1445

Supported sites: [ccMixter](http://ccmixter.org), [TuneTrack](http://tunetrack.net/) 

Know issues:
- Unrecognised song
It's better to switch on the notification

- Beta UI
  - The player will be hidden when and after visting the home page

### DIFF
--- a/connectors/ccmixter.js
+++ b/connectors/ccmixter.js
@@ -3,6 +3,8 @@
 setupConnector();
 
 function setupConnector() {
+	setupCommonProperties();
+
 	if (isCcPlayer()) {
 		setupCcPlayer();
 	} else if (isBetaPlayer()) {
@@ -12,8 +14,20 @@ function setupConnector() {
 	}
 }
 
+function setupCommonProperties() {
+	// Track item containers
+	Connector.getCurrentPlayer = () => {
+		let containers = Array.from($('.upload_info, .box, .trr, .tile, .tracklist-item, .tree-head'));
+		return containers.find((container) =>
+			Connector.testCurrentPlayer(container));
+	};
+
+	// eslint-disable-next-line no-unused-vars
+	Connector.testCurrentPlayer = (container) => false;
+}
+
 function isCcPlayer() {
-	return $('[id^=_ep_], .playerdiv').length > 0;
+	return $('[id^=_ep_], #tabs').length > 0;
 }
 
 function setupCcPlayer() {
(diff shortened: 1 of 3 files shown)


connectors/ccmixter.js
```diff
@@ -0,0 +1,111 @@
...
+
+setupConnector();
+
+function setupConnector() {
+	if (isBetaPlayer()) {
+		setupBetaPlayer();
+	} else if (isTuneTrackPlayer()) {
+		setupTuneTrackPlayer();
+	} else {
+		setupH5Player();
```
inverse: Should we add a check for the H5 page so we can console log if we fail to match or is this generic enough to handle all cases?

tests/connectors/ccmixter.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = (driver, spec) => {
+	spec.shouldLoadWebsite(driver, {
```
inverse: Is it worth adding a test for each type of player?

connectors/ccmixter.js
```diff
@@ -0,0 +1,111 @@
...
+	};
+}
+
+function setupH5Player() {
+	$('audio').bind('playing pause timeupdate', Connector.onStateChanged);
+
+	Connector.getCurrentPlayer = () => {
+		let containers = document.querySelectorAll('.upload_info, .box, .trr');
+		return Array.from(containers).find((e) => {
+			let a = e.querySelector('audio');
```
inverse: Could we use more descriptive naming of variables, this isn't clear.

xyauhideto:
Now I manage to get the flashplayer work, so I include it as well.

tests/connectors/ccmixter.js
```diff
@@ -0,0 +1,22 @@
...
+	// Beta player
+	spec.shouldContainPlayerElement(driver, {
+		url: 'http://beta.ccmixter.org/tree',
+		playButtonSelector: '.play-button'
+	});
+
+	// TuneTrack player
+	spec.shouldContainPlayerElement(driver, {
+		url: 'http://tunetrack.net',
+		playButtonSelector: '.play-btn button:contains(Pause)'
```
alexesprit: WebDriver throws an error when clicks on button by this selector.

tests/connectors/ccmixter.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+module.exports = (driver, spec) => {
+
+	// Beta player
+	spec.shouldContainPlayerElement(driver, {
```
alexesprit: Is there a reason to use this function to check the connector? `shouldBehaveLikeMusicSite` works fine too for me.

alexesprit:
Nice, thanks! 👍

PR #1453, opened 2017-10-24
Updated connectors "fritz" and "radioeins"

carpet-berlin:
### Simplified "fritz" connector
The "is playing" status can simply be determined from a certain tag's class names.

### Two types of "radioeins" players exist
The player page randomly provides one player (jwplayer) or another (lsplayer). The old (rather complicated) player is still supported. The new player has an <audio> element.


alexesprit:
@carpet-berlin Could you rebase master branch on latest master from upstream?

PR #1454, opened 2017-10-25
Updated "fritz" and "radioeins" connectors

carpet-berlin:
### Simplified "fritz" connector

The "is playing" status can simply be determined from a certain tag's class names.

### Two types of "radioeins" players exist

The player page randomly provides one player (jwplayer) or another (lsplayer). The old (rather complicated) player is still supported. The new player has an element.


alexesprit:
Thanks! 👍 

You can force push after rebase instead of opening new PR. ;)

PR #1475, opened 2017-11-12
luooMusic:  Add support for essay, musician, search

xyauhideto:


connectors/luoo.js
```diff
@@ -1,9 +1,109 @@
 'use strict';
+(() => {
```
alexesprit: Is there a reason to pull all code in this self-invoking function?

connectors/luoo.js
```diff
@@ -1,9 +1,109 @@
...
+	let setupConnector = () => {
```
alexesprit: Could you use function declaration instead of function expression?

xyauhideto: Okay

connectors/luoo.js
```diff
@@ -1,9 +1,109 @@
...
+			let text = Connector.getCurrentItem().find('.content').text();
+			let [album, artist] = text.trim().split(', ').map((text) => text.replace(/^\S+\s/g, '')) || Array(2).fill(null);
+			return { album, artist };
+		};
+
+		Connector.getAlbum = () => Connector.getAlbumArtist().album;
+
+		Connector.getArtist = () => Connector.getAlbumArtist().artist;
+
+		Connector.getUniqueID = () => Connector.getCurrentItem().data('id') || null;
```
alexesprit: `|| null` can be omitted.

tests/connectors/luoo.js
```diff
@@ -1,7 +1,21 @@
...
+		url: 'http://www.luoo.net/essay/915'
+	});
+	// musician
+	connectorSpec.shouldBehaveLikeMusicSite(driver, {
+		url: 'http://www.luoo.net/musician/'
+	});
+	// search
+	connectorSpec.shouldBehaveLikeMusicSite(driver, {
+		url: 'http://www.luoo.net/search/?q=love&t=single',
+		playButtonSelector: '.icon-search-play:eq(-1)'
```
alexesprit: WebDriver throws an error when clicks on an element with such selector. Do you have the same issue?

I tried to use `.icon-search-play` selector and it works fine for me, but there's another thing that first three songs from search results are unplayable for me. Clicking on play icon does nothing. I updated the test like this way and it passed for me:
```js
	// search
		url: 'http://www.luoo.net/search/?q=Spinning&t=single',
		playButtonSelector: '.icon-search-play'
	});
```

Issue #1503, opened 2017-11-20
Scrobbling from letsloop


Seryc:
### Expected behavior
I was wandering if would be any possibility to scrobble from letsloop.com one day? Thanks

inverse:
Sure - let me implement this :)

PR #1504, opened 2017-11-20
Added letsloop

inverse:
Refs #1503


Seryc:
Still saying "Last.fm Scrobbler does not support scrobbling from this website" to me. I reauthenticated, tried different browser and nothing happened.

inverse:
It'll be in the next release.

PR #1532, opened 2017-12-11
Add Český Rozhlas

balaziks:
Adds scrobbling from all stations of Český Rozhlas (Czech national radio).

### DIFF
--- a/connectors/rozhlas.js
+++ b/connectors/rozhlas.js
@@ -1,0 +1,7 @@
+'use strict';
+
+Connector.playerSelector = '#block-live-player';
+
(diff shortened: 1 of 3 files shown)


inverse:
Could we add a test case?

balaziks:
I added a test case by editing the `radionomy.js` file (I did the same thing for the connector itself). But I don't have deep understanding of how the tests here work and if for example the fact that the radio stations sometimes don't play music (there are also news and other shows played on them).

balaziks:
It would not really make any difference since `.fp-play` matches the play/pause button no matter what current player state is while `.is-paused` matches the same button in paused state only.

But I agree that it is clearer now.

PR #1533, opened 2017-12-11
add placeholder texts to titles

myfonj:
Problem: while editing track info it is not possible to tell the title from artist without temporarily deleting one value; fix: duplicating the placeholder text to title could help a little bit.


alexesprit:
👍

PR #1540, opened 2017-12-19
implement connector for perisonic frontend to subsonic

mschaaf:
It is a frontend for the Subsonic music server and can be found under http://robinbakker.nl/perisonic/

### DIFF
--- a/connectors/perisonic.js
+++ b/connectors/perisonic.js
@@ -1,0 +1,12 @@
+'use strict';
+
+Connector.artistSelector = '#title h3';
+
+Connector.trackSelector = '#title p';
+
+Connector.trackArtSelector = '#coverart img';
+
+Connector.isPlaying = () => $('#playpause').hasClass('pause');
+
+$('#title').bind("DOMSubtreeModified", Connector.onStateChanged);
+
--- a/core/connectors.js
+++ b/core/connectors.js
@@ -262,6 +262,12 @@ define(function() {
 	},
 
 	{
+		label: 'Perisonic',
+		matches: ['*://robinbakker.nl/perisonic/*'],
+		js: ['connectors/perisonic.js'],
+	},
+
+	{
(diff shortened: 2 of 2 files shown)


connectors/perisonic.js
```diff
@@ -0,0 +1,12 @@
...
+
+Connector.artistSelector = '#title h3';
+
+Connector.trackSelector = '#title p';
+
+Connector.trackArtSelector = '#coverart img';
+
+Connector.isPlaying = () => $('#playpause').hasClass('pause');
+
+$('#title').bind("DOMSubtreeModified", Connector.onStateChanged);
```
alexesprit: I guess the `#title` element contains track info only. If I am correct, the extension won't process play/pause state change.

Is there any element which contains both track info and play/pause button? If so, this element can be used as a player selector (`Connector.playerSelector`).

mschaaf: > I guess the #title element contains track info only. If I am correct, the extension won't process play/pause state change.

Yes, it is not able to process play/pause. Currently, it doesn't feel like a limitation.

> Is there any element which contains both track info and play/pause button? If so, this element can be used as a player selector (Connector.playerSelector).

I tried that element and it is not working.

If you want to give me a better advice or point me to a more detailed docu for a such a player situation I would be satisfied.

alexesprit: I cannot test the connector, because Perisonic doesn't work for me (Subsonic web UI works fine, though). But I looked into page source and noticed a `#player` element which contains player elements and track info. Can this element be used as a player selector? If not, I'll accept this PR anyway, since it works for you fine.

mschaaf: this element didn't work

alexesprit: That's sad.

mschaaf:
Thanks for accepting and merging.

PR #1549, opened 2018-01-01
Add duration parsing to Amazon Alexa connector

benscobie:
The Amazon Alexa SPA app has a Now Playing tab which includes additional metadata about the currently playing song. We can use this to parse the duration time of the song. The user will need to navigate and stay on the Now Playing tab in order for this new functionality to work. This appears to fallback gracefully if the user is not on the Now Playing tab though.

I've had to include a workaround to get around the fact that the UI does not update some of the metadata straight away when changing song which lead to incorrect duration time calculations.

**Note**: This **was** working on the v2.3.x branch but does not appear to work in master, even without my changes. It doesn't detect that a song is playing.

### DIFF
--- a/connectors/amazon-alexa.js
+++ b/connectors/amazon-alexa.js
@@ -1,7 +1,8 @@
 'use strict';
 
 Connector.playerSelector = '#d-content';
-Connector.remainingTimeSelector  = '.d-np-time-display.remaining-time';
+Connector.albumSelector = '#d-info-text .d-sub-text-2';
+Connector.remainingTimeSelector = '.d-np-time-display.remaining-time';
 Connector.currentTimeSelector = '.d-np-time-display.elapsed-time';
 
 Connector.getDuration = () => {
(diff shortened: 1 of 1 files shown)


alexesprit:
> Note: This was working on the v2.3.x branch but does not appear to work in master, even without my changes. It doesn't detect that a song is playing.

Is this true for Amazon Alexa only?

benscobie:
I've tried the YouTube connector and that seems to work okay.

alexesprit:
@benscobie Could you try to use `length` property instead of `size` function when checking element count?

benscobie:
Done. Do you want me to take a look as to why the connector isn't working in master?

alexesprit:
@benscobie I supposed it didn't work because of `size` function (I updated jQuery recently). Yes that would be great, if you could figure out why the connector doesn't work on master.

Probably, it's something wrong in content scripts.

[How to debug the extension.][1]

[1]: https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension#logs-of-content-scripts

benscobie:
I can confirm it's working in master after the size() replacement.

alexesprit:
Nice, thanks!

PR #1559, opened 2018-01-20
Added support for Resonate.is

### LINKED ISSUE #1558: Support Resonate.is
### Expected behavior

Support scrobbling for the Resonate streaming service.

Sample URLs
   https://beta.resonate.is/  (Unified player)
   https://resonate.is/album/138/ugress-it_came_from_beyond_eternal_september/

I will happily fund a Resonate account for any developer that wants to take this on.

inverse:
Resolves #1558

Should we add a test? It's a private site so how will that be handled?

### DIFF
--- a/src/connectors/resonate.js
+++ b/src/connectors/resonate.js
@@ -1,6 +1,5 @@
 'use strict';
 
-
 if (location.host.startsWith('beta3')) {
 	Connector.playerSelector = '#player';
 
@@ -11,7 +10,7 @@ if (location.host.startsWith('beta3')) {
 	Connector.durationSelector = '.playlist-item--playing .duration';
 
 	Connector.getTrackArt = () => {
-		return $('.player-artwork > div').css('background-image').replace('url(','').replace(')','').replace(/\"/gi, "");
+		return $('.player-artwork > div').css('background-image').replace('url(', '').replace(')', '').replace(/"/gi, '');
 	};
 
 	Connector.trackArtSelector = '.current-cover';
--- a/src/connectors/resonate.js
+++ b/src/connectors/resonate.js
@@ -10,13 +10,10 @@ if (location.host.startsWith('beta3')) {
 	Connector.durationSelector = '.playlist-item--playing .duration';
 
 	Connector.getTrackArt = () => {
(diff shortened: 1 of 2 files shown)


inverse:
Not sure why the build is failing - I didnt touch that file :(

src/connectors/resonate.js
```diff
@@ -0,0 +1,36 @@
...
+	Connector.playerSelector = '#player';
+
+	Connector.artistSelector = '#player .metas a';
+
+	Connector.trackSelector = '#player .name';
+
+	Connector.durationSelector = '.playlist-item--playing .duration';
+
+	Connector.getTrackArt = () => {
+		return $('.player-artwork > div').css('background-image').replace('url(', '').replace(')', '').replace(/"/gi, '');
```
alexesprit: This code could probably replaced by `Util.extractUrlFromCssProperty` function.

src/connectors/resonate.js
```diff
@@ -0,0 +1,36 @@
...
+	};
+
+	Connector.trackArtSelector = '.current-cover';
```
alexesprit: This property and another one below won't be used because of overriden `Connector.getTrackArt` function.

inverse: good catch!

lindner:
Looks like beta3 has been promoted to beta.  (That was fast...)

@peteratomic - can you confirm that this is going to stay this way?

lindner:
Also it would be nice if the Album Title was available in the DOM for use.

peteratomic:
Looks like Augustin is a few days ahead of schedule! Yes I think the address will stay the same until we go out of beta status.

inverse:
So shall I remove the beta3 routing and make this the normal flow? 

- Having album title in the dOM would a be a nice thing
- Also having the current duration would make scrobbling nicer, although not required

lindner:
Hi folks, the resonate devs have added the album data into the DOM.  Here's an example:

<div class="flex items-end w-100">
<span class="track-name truncate b pv1">Lung Breathers</span>
<span class="track-album truncate pv1 pl1 f7">Coolboy</span></div>

inverse:
@lindner  Updated

lindner:
Changes LGTM

lindner:
LGTM == Looks Good to Me.

PR #1561, opened 2018-01-23
Remove genre with 【 】 from YouTube

MajorWorkAlot:
Old xKito videos still have those weird brackets. I have been using this change locally for about a year now.


alexesprit:
Could you add a test in `tests/core/util.js` for this case?

MajorWorkAlot:
I am not sure how things like this should be handled. I didn't think about it doing the change.
[【genre】] Artist - Track
【[genre]】 Artist - Track

PR #1570, opened 2018-02-05
Adapted "deutschlandfunknova" connector after site changes

carpet-berlin:


src/connectors/deutschlandfunknova.js
```diff
@@ -1,18 +1,22 @@
...
-Connector.playerSelector = '.player__main';
+Connector.playerSelector = 'div.player.desktop';
 
-Connector.artistTrackSelector = 'span.title--title';
+Connector.artistTrackSelector = '.playing-title';
 
 Connector.getArtistTrack = () => {
 	let text = $(Connector.artistTrackSelector).text();
-	let m = text.match(/"(.*)" von (.*)/);
+	let m = text.match(/"(.*?)" von (.*?)\u2003\u00B7\u2003/);
```
inverse: You really need those chars exactly? Wouldn't `.*` be better?

carpet-berlin: The title info is available as ticker text only, like that:
"Title" von Artist · "Title" von Artist · "Title" von Artist
where the separator (' · ') is formed by the special unicode chars (em space, middle dot, em space).
As everything is contained more than once in one line, I think the easiest solution is to use the ungreedy modifier.

inverse: Ahhh - Okays that makes sense :+1:

PR #1594, opened 2018-03-08
Add connector for KEXP Radio (Seattle 90.3FM)

goldenapples:
Adds a basic connector configuration for KEXP web radio player
(https://kexp.org).

### DIFF
--- a/src/connectors/kexp.js
+++ b/src/connectors/kexp.js
@@ -1,0 +1,7 @@
+'use strict';
+
+Connector.isPlaying = () => $( '.Player--streamPlaying' ).hasClass( 'Player--playing' );
+
+Connector.playerSelector = '.Player-meta';
+Connector.artistTrackSelector = '.Player-title';
+Connector.albumSelector = '.Player-album';
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -1177,5 +1177,9 @@ define(function() {
 		label: 'HQ Radio',
 		matches: ['*://hqradio.ru/*'],
 		js: ['connectors/hqradio.js'],
+	}, {
+		label: 'KEXP Radio',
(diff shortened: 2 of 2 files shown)


src/connectors/kexp.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+Connector.isPlaying = () => $( '.Player--streamPlaying' ).hasClass( 'Player--playing' );
+
+Connector.playerSelector = '.Player-meta';
```
inverse: Would't `.Player` make more sense? I couldn't see this element...

src/connectors/kexp.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+Connector.isPlaying = () => $( '.Player--streamPlaying' ).hasClass( 'Player--playing' );
```
inverse: Could we remove redundant spaces around selector?

goldenapples:
Thanks for the quick feedback! I've updated the PR. I wasn't able to run the tests locally, but I think the spec I added should pass...

inverse:
@alexesprit are you able to run tests? Did the new project structure break the functional tests? I get errors too from this end :cry:

alexesprit:
@goldenapples please add the `*://www.kexp.org/` pattern to the URL match list. The connector works fine, though.

goldenapples:
Thanks @alexesprit I've added the www. subdomain to the pattern, as well as added patterns without a trailing slash after the domain name, to catch any possible URLs on the website that can play the radio. Thanks again for the quick review!

jtojnar:
Could you publish a new release including this, please?

jtojnar:
Will the Firefox version be released as well?

PR #1626, opened 2018-05-15
Add Radio Rocks connector

dmmat:
url: https://www.radioroks.ua/player/


dmmat:
Song change not auto detect, but triggered if change volume

src/core/connectors.js
```diff
@@ -1181,5 +1181,17 @@ define(function() {
 		label: 'KEXP Radio',
 		matches: ['*://kexp.org', '*://www.kexp.org', '*://kexp.org/*', '*://www.kexp.org/*'],
 		js: ['connectors/kexp.js'],
+	}, {
+		label: 'Radio Rocks',
+		matches: [
```
inverse: Would `www.radioroks.ua/player/*` be enough? For me I am always seeing www redirection and would cover further channels.

dmmat: I try, no effect

Issue #1641, opened 2018-06-02
Support for music.youtube.com


redsx16:
Music.youtube.com is not available as a connector. I tried adding https://music.youtube.com/* and *.youtube.com/ as url patterns. The it now shows it as supported but does not detect any music playing so nothing is scrobbling. FYI: This is the new Youtube Music that will eventually replace Google Play Music.

### Expected behavior
Songs scrobble to last.fm with proper artist and song title.

### Current behavior
None of the tracks playing are recognized so nothing scrobbles.

### Possible solution
Since the site is new most likely a new connector is needed to match formatting.

### Steps to reproduce (for bugs)

https://pastebin.mozilla.org/9086861

1. Load up music.youtube.com
2. Play any song
3. Check Last.fm or the extension for status
4. Try pressing bottom right corner arrow and notification pops indicating track not recognized

### Steps for resolving the issue

- [ ] Reauthenticate on Last.fm

- [ ] Try to use unpacked extension from latest master

- [ ] Check other issues for duplicates

### Environment

* Version used: 2.5 from Chrome store
* Browser name and version: Chrome 67.0.3396.62 (Official Build) beta (64-bit)

jawshoeah:
I can confirm music.youtube.com is not scrobbling.

jaccarmac:
Just got my invite so I will attempt to whip up a connector tonight.

jaccarmac:
Implemented, please give the PR a try if you feel comfortable running from source. I could use tests from a non-American locale, and the code is very rough but functional on my end!

redsx16:
So far it's working for me. Only issue I've noticed is that songs that have a "feat" artist are also found in the artist field when it's scrobbling with Force Track Recognition enabled. Each case the "feat" artist is after the main artist, ex: BlocBoy JB & Drake — Look Alive (feat. Drake) should be BlocBoy JB — Look Alive (feat. Drake)

jaccarmac:
Thanks for giving it a go. I noticed that. Gonna wait on making a change, as personally I tend to like the Beets style of scrobbling (features go in the artist field instead). Additionally check out a track that's supposed to have multiple artists on it, like https://music.youtube.com/watch?v=6qwwuKIj8BA.

redsx16:
Your example makes sense because there is no featured artist so none of the artists are found in both the artist field and the title field. At least for me it's unnecessary to have it duplicated. I don't know what the standards are.

jaccarmac:
[...]

PR #1644, opened 2018-06-05
Implement basic YT Music connector.

jaccarmac:
See #1641 

I've deliberately waited to add tests because this is incomplete. The way YT has set up this new page is a bit strange. On the upside, the elements don't move around as you navigate so we can lock onto the bar. However, now-playing status and artist/album splitting are based on hardcoding strings at the moment. Now-playing especially may not work in non-English locales. Also the way artists are handled is a little weird, features get put in the artist field on the page but are also included in track titles. Since artists are hyperlinked it is possible to pull apart the main artist from their features, but I have not done so at this point.

TL;DR: No tests yet, C+C accepted and encouraged!

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -22,8 +22,8 @@ Connector.isPlaying = () => {
 	return $('.ytmusic-player-bar.play-pause-button').attr('title') === 'Pause';
 };
 
-let separateArtist = byline => byline.split('•')[0];
-let separateAlbum = byline => byline.split('•')[1];
+let separateArtist = (byline) => byline.split('•')[0];
+let separateAlbum = (byline) => byline.split('•')[1];
 
 const filter = new MetadataFilter({
 	artist: separateArtist,
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -1,0 +1,32 @@
+'use strict';
+
+Connector.playerSelector = 'ytmusic-player-bar';
+
+Connector.getTrackArt = () => {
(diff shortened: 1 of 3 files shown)


jaccarmac:
The unrecognized track may be buggy. I noticed a not recognized popup appear but on switching to the window the track info had been parsed successfully. I'll get back to it when I have time but anyone else who wants to try it make sure to enable that setting to smoke test.

jaccarmac:
That check failure smelled like an eslint version conflict or something. Real weird, everything linted on my side but apparently Grunt changed the formatting.

inverse:
A good start? how are you finding using it in practice? I think we need to also cover cases. e.g. 

- When it's just playing youtube videos, it didn't recognise things for me
- Skipping ads

inverse:
Anymore work on this?

jaccarmac:
Not yet. I need to come back to it but haven't had a chance as it's mostly working for me so far.

jaccarmac:
Follow the contribution guide in this repo but check out my branch instead of master.

inverse:
@jaccarmac if you could:

- Resolve the merge conflicts
- Figure out the normal music detection
- Ad playback prevention scrobbling 

then we should be good to go!

jaccarmac:
Won't have time this weekend, will try to carve some out next week.

jaccarmac:
All those things should be addressed. Let me know what things look like on your end, I can't easily test the advertisement experience.

jaccarmac:
@MajorWorkAlot This version should deal with your link more sanely. Still not sure exactly how to determine when to apply the YT filter set to a title.

@inverse Your Boards of Canada video scrobbles properly for me now.

PR #1649, opened 2018-06-10
Implement IDAGIO connector

r8:
[IDAGIO](https://www.idagio.com/) is a streaming service for classical music.

This pull request adds connector for scrobbling tracks played with IDAGIO web app.


r8:
No rush, but is it something wrong with my pull request, or you're just too busy at the moment to look at it?

Tanami:
I would also really like this.

PR #1666, opened 2018-07-02
Add support for Hotmixradio.fr

calvinturbo:
Just quickly wrote a connector for Hotmixradio.fr, works with all the different radiostations on their website

### DIFF
--- a/src/connectors/hotmixradio.js
+++ b/src/connectors/hotmixradio.js
@@ -1,0 +1,22 @@
+'use strict';
+
+Connector.playerSelector = '#new_ajax-player';
+
+Connector.artistSelector = '#hotmixPlayerSongArtist .ajax';
+
+Connector.getArtist = () => getScrollableText('#hotmixPlayerSongArtist');
+
+Connector.getTrack = () => getScrollableText('#hotmixPlayerSongTitle');
+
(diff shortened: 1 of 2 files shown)


calvinturbo:
I see there's extra space or something and that's why the check has failed. Can someone correct this since I wrote this in Notepad++? Don't have other tools available atm

src/connectors/hotmixradio.js
```diff
@@ -0,0 +1,22 @@
...
+
+Connector.playerSelector = '#new_ajax-player';
+
+Connector.artistSelector = '#hotmixPlayerSongArtist .ajax';
+
+Connector.getArtist = () => getScrollableText('#hotmixPlayerSongArtist');
+
+Connector.getTrack = () => getScrollableText('#hotmixPlayerSongTitle');
+
+//Connector.getAlbum = () => getScrollableText('#songalbum');
```
antonin-lebrard: Actually you only have to add a space here, after the `//` to resolve the travis error. This should be possible with Notepad++

calvinturbo: Thanks, I will just remove the line

calvinturbo:
Have been using this script to scrobble from Hotmixradio.fr for a couple of hours ago, seems to work perfectly.

Issue #1400, opened 2017-09-13
Amazon Echo/Alexa scrobbles on Firefox are sending playlist name instead of album name


tchrischan:
If I play a playlist (either from alexa voice command, or Spotify Connect), the web scrobbler sends the playlist name as the album name.  It sends the album name correctly if I play a song or album by name, or use Your Daily Mix.

### Expected behavior

Album name of song should be scrobbled.

### Current behavior

Playlist name of song scrobbled instead.

### Possible solution

Get the album name from Spotify and use that instead (assuming it's accessible).

### Steps to reproduce (for bugs)

1.  Load Alexa Player website in firefox with web scrobbler add-in.
2.  Ask Alexa to play a playlist from Spotify. 

### Context

I have a bunch of dynamic playlists on my computer that rely on last.fm scrobbles, so bad data is a nuisance for me.

### Steps for resolving the issue

- [ y] Reauthenticate on Last.fm

- [n/a ] Try to use unpacked extension from latest master

- [ y] Check other issues for duplicates

### Environment

* Version used: 1.68.1
* Browser name and version: Firefox 55.0.3 (32-bit)

alexesprit:
@puthali could you check this issue and fix it?

thedrinkeno:
This is also reproducible by playing something from Pandora on Alexa via Echo voice commands. Instead of the album of the track, the Radio name is scrobbled instead.

There is a musicbrainz api that can possibly be leveraged

https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search

matelich:
Also happens on Amazon Music webplayer on Chrome - playlist name is being inserted as album name. I always use playlists, so I'd be happy with just an option to turn off populating the album field

inverse:
Will be in fixed in the next release :+1:

alexesprit:
Closing this, because #1694 is merged.

PR #1694, opened 2018-08-27
Fix amazon album

inverse:
Fix for #1400

### DIFF
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -11,9 +11,17 @@ Connector.getArtist = () => {
 Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
 Connector.getAlbum = () => {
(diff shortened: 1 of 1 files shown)


src/connectors/amazon.js
```diff
@@ -11,9 +11,17 @@ Connector.getArtist = () => {
...
-		$('.trackSourceLink a').attr('title');
+	if ($('tr.selectable.currentlyPlaying td.albumCell')) {
+		return $('tr.selectable.currentlyPlaying td.albumCell').attr('title');
+	}
+
+	if ($('.nowPlayingDetail img.albumImage')) {
+		return $('.nowPlayingDetail img.albumImage').att('title');
+	}
+
+	if ($('.trackSourceLink a').data('ui-click-action') === 'selectAlbum') {
```
inverse: When playing from playlist this data attr does not exist

PR #1698, opened 2018-09-02
Account for TIDAL redesign.

jaccarmac:
Redesign seemed to roll in late last week.

data-test-id attributes are probably less than ideal but until I absolutely
have to use string matches on class attributes I will not do so.

I tried to follow the old structure as much as possible; Not totally sure why
the duration and album art are not restricted to the player selector.

### DIFF
--- a/src/connectors/tidal.js
+++ b/src/connectors/tidal.js
@@ -1,15 +1,19 @@
 'use strict';
 
-Connector.playerSelector = '#player';
+Connector.playerSelector = '[class^="nowPlaying"]';
 
-Connector.trackArtSelector = '.js-footer-player-image';
+Connector.playButtonSelector = `${Connector.playerSelector} [class^="playbackToggle"]`;
 
-Connector.trackSelector = `${Connector.playerSelector} [data-bind="title"]`;
+Connector.isScrobblingAllowed = () => !!$(Connector.playButtonSelector);
 
-Connector.artistSelector = `${Connector.playerSelector} [data-bind="artist"] a:first`;
+Connector.isPlaying = () => $(Connector.playButtonSelector).attr('data-test-id') === 'pause';
 
-Connector.playButtonSelector = `${Connector.playerSelector} .play-controls__play`;
+Connector.trackSelector = `${Connector.playerSelector} [class^="mediaInformation"] span:eq(0)`;
 
-Connector.currentTimeSelector = '.js-progress';
+Connector.artistSelector = `${Connector.playerSelector} [class^="mediaArtists"]`;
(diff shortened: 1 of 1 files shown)


jaccarmac:
Clearing the queue seems to do weird things to scrobble detection. Less badly-behaved than the last scrobbler at least but less than ideal.

inverse:
Thanks for working on this. I don't have a Tidal account so I am unable to test. 

What do you mean with comm's between their name? Normally if there needs to be additional parsing on any getter we overload the function if it becomes too complex with a selector.

jaccarmac:
Yep, that's what I mean. The DOM looks like the snippet below, and I'm matching on the `data-test-id`

```html
<span class="mediaArtists--2L7Wb">
  <a href="/artist/7628878" title="House of Feelings" data-test-id="grid-item-detail-text-title-artist" draggable="false">House of Feelings</a>
  , 
  <a href="/artist/6118403" title="Denitia" data-test-id="grid-item-detail-text-title-artist" draggable="false">Denitia</a>
</span>
```

inverse:
So a track can have multiple artists attributed?

jaccarmac:
Give it a think for sure, since it's not only a problem with Tidal. Weird artist combos are present on YouTube, Spotify, and elsewhere as well. IMO it's really a problem with Last.fm itself.

In any case, my thinking w.r.t. proper solutions is to make an option between two scrobbling styles.

1. Last.fm style: Primary artist per track, additional artists in a feature list at the end of the track name.
2. Musicbrainz style: Artists and features in artist field, track title clean.

Personally my preference is for 2, Last.fm works better with 1, and most scrobblers use a mix of both.

inverse:
Totally - For example here's one that kinda goes the way you said with 1.

https://www.last.fm/music/Modeselektor/_/Berlin+feat.+Miss+Platnum

jaccarmac:
Alright, I cleaned this up and made it play a little nicer with the queue, should be easy to merge. There's one more issue that I think may be a wider bug since I also see it on Google Play Music, so I may open another ticket for that.

Play/pause state changes aren't getting picked up. It seems that the MutationObserver is not getting triggered by clicking the pause button, which doesn't actually add or remove anything from the DOM. It simply changes the value of some attributes. Not sure what the best way of dealing with that is given all the MutationObserving is being done outside of individual connectors.

jaccarmac:
Not going to ask why but the most recent change made that problem go away. I'll take it. This is ready to go now.

inverse:
Changes look reasonable - I am unable to test though

PR #1709, opened 2018-09-11
Fixes FIP title and album

4-Eyes:
The track and album selectors for FIP returned multiple elements, which meant the track and album were always wrong. This change ensures that only a single element, containing the title/album is selected.


inverse:
Thanks :+1:

PR #1712, opened 2018-09-12
new deezer layout support added

markon1:


src/connectors/deezer.js
```diff
@@ -1,20 +1,44 @@
 'use strict';
 
-Connector.playerSelector = '#page_sidebar';
+if($('.page-player').length){
```
inverse: Although this most likely works I think separating all the logic for new and old cleaner. My reasoning is that when we need to eventually clean up the old layout we can do so easily.

markon1: @inverse so instead of having IF for each selector, I should have only one main IF which decides if it's new or old layout? Is that the only change required?

lbrueckner: This updated connector did not find the track and artist data when the queuelist is opened in the new deezer.

I've done a more complete fix for myself last night, should I do another pull request for this?

inverse: Yes - one big for old and new would be cleaner.

PR #1721, opened 2018-09-18
Implement Naver Connector :)

minimal1:
[Naver](https://music.naver.com/) is the top of Korean search engines.
This pull request adds connector for scrobbling musics played with naver music web app.


inverse:
Thnx!

PR #1722, opened 2018-09-19
Bugs connector encoding change

JeonHyungJin:
Bugs connector which i added has an encoding problem, because it contains Korean.
So i changed bugs.js file`s encoding in connector repository.

### DIFF
--- a/src/connectors/bugs.js
+++ b/src/connectors/bugs.js
@@ -16,7 +16,7 @@ Connector.durationSelector = '.finish';
 
 Connector.isPlaying = () => {
 	let btn = $('.btnPlay').first();
-	return btn.html() === 'ÀÏ½ÃÁ¤Áö'; // if text 'ÀÏ½ÃÁ¤Áö' in button, it means 'pause' because music is playing.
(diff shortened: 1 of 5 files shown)


JeonHyungJin:
I missed an encoding problem.... Sorry.. haha

src/connectors/bugs.js
```diff
@@ -16,7 +16,7 @@ Connector.durationSelector = '.finish';
 
 Connector.isPlaying = () => {
 	let btn = $('.btnPlay').first();
-	return btn.html() === '�Ͻ����'; // if text '�Ͻ����' in button, it means 'pause' because music is playing.
+	return btn.html() === '일시정지'; // if text 'ÀÏ½ÃÁ¤Áö' in button, it means 'pause' because music is playing.
```
inverse: Can we remove the comment since I dont think it brings any value.

JeonHyungJin: Yes. you can remove that comment!

JeonHyungJin:
I removed that comment!

JeonHyungJin:
Thanks!!!!

PR #1723, opened 2018-09-19
Implement Soribada Connector

NamJinWoo:


NamJinWoo:
This soribada site is a music webapp in South Korea. This is almost the first music web site in our country. Please take a look and leave me a comment. This would help your site have more datas.

JeonHyungJin:
Maybe your code missed " , " in core/connectors.js file.
:)

NamJinWoo:
core/connector part has been modified!

src/connectors/soribada.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+
```
inverse: Spacing seems a bit off in this file.

src/core/connectors.js
```diff
@@ -967,11 +967,14 @@ define(function() {
 	},
 	{
 		label: 'Naver',
-		matches: ['*://playerui.music.naver.com*'],
+		matches: ['*://playerui.music.naver.com/*'],
```
inverse: Looks like I missed that one :/

src/connectors/soribada.js
NamJinWoo: okay, I've done it! Sorry about it!

src/core/connectors.js
NamJinWoo: Never Mind, I just looked around and it was a simple error!

NamJinWoo:
I've  done all the changes requested!

src/connectors/soribada.js
```diff
@@ -0,0 +1,23 @@
...
+Connector.artistSelector = '.info .pado_tit_wrap span:nth-child(2)';
+
+Connector.trackSelector = '.pado_tit_wrap strong:first';
+
+Connector.currentTimeSelector = '#progress .now';
+
+Connector.durationSelector = '#progress .total';
+
+Connector.isPlaying = () => {
+
```
inverse: Could we also remove this extra spacing too?

src/connectors/soribada.js
```diff
@@ -0,0 +1,23 @@
...
+	return $('#controller .pause').length !== 0;
+
+};
+
+Connector.getTrackArt = () => {
+
```
inverse: And here

inverse:
After the final bits we should be good :+1:

NamJinWoo:
I fixed the extra spacing. Please check ! It is so nice to have a feedback with you. I'll try my best to solve many issues and contribute to this project. Let's keep in touch!

PR #1724, opened 2018-09-19
bus fix about the album selector

JeonHyungJin:
After i tested my code many times, i found some problem about the album selector...
So i fixed the bug about my code and i made a new PR.


JeonHyungJin:
And how can i be a contributor about this project? :)

inverse:
Thnx for the fix - You already are ;)

PR #1725, opened 2018-09-19
[BugFix] isPlaying in naver.js wasn't working

minimal1:
It's kind of jquery selector.
I fixed the selector adapted jquery grammer.


minimal1:
oops...I'm sorry. 
I found why I could see my name.
There are only 100 contributors in 'Contributors'.

Issue #1718, opened 2018-09-18
Update Radioparadise.com because of web site update


TokeRider:
Please update scrobbling for new.radioparadise.com.  They recently updated their web player(and it's awesome) to a new format that doesn't get songs scrobbled by Web Scrobbler.  Thank you.

JeonHyungJin:
I updated [Radioparadise](https://www.radioparadise.com) `s source code.
So now this program will support this web site :)
I made [new PR](https://github.com/web-scrobbler/web-scrobbler/pull/1729) for this.

alexesprit:
I think it was resolved already as the extension supports two player types (btw, the old one is gone, right?).

TokeRider:
Yes the old one is gone.
    On Tuesday, April 23, 2019, 3:55:10 PM EDT, Alexey <notifications@github.com> wrote:  
 
 
I think it was resolved already as the extension supports two player types (btw, the old one is gone, right?).

alexesprit:
Legacy player is available on legacy.radioparadise.com, but I broke support for that addresses. :D Not sure if we need to keep support.

PR #1727, opened 2018-09-21
Update Radioparadise.js for support that web

JeonHyungJin:
Radioparadise.com was updated. So now this program does not support [Radioparadise](https://www.radioparadise.com) web player.
Issue #1718 talked about this. So i updated this web site`s js file.
Thanks :)


JeonHyungJin:
I will do again.... !!

PR #1730, opened 2018-09-21
Added one regex to the removeRemastered-filter

lbrueckner:
I noticed the other day that the deezer connector is using the removeRemastered filter and now I was wondering why on some very common occurrences the Remastered-string is not removed. To fix this I've added a regex that matches (Remastered) and [Remastered]  at the end of the string.

The deezer search for 'Remastered' is showing me 300 albums from which >100 have just (Remastered) or [Remastered] at the end of their titles. So I think this is a worthwhile addition.

First I'd added a regex that matches Remastered with any additional text inside of rounded or square braces. This would match even ~70 more albums out of this 300 that are unmatched by the existing filter. However, there are a couple of examples (~15), like "Urban Hymns (Super Deluxe / Remastered 2016)", where Super Deluxe indicates a special edition with additional songs. And I'm unsure if it is desirable to remove this.

### DIFF
--- a/src/core/content/filter.js
+++ b/src/core/content/filter.js
@@ -141,7 +141,7 @@ class MetadataFilter {
 	}
 
 	/**
- 	 * Predefined filter functions.
+	 * Predefined filter functions.
 	 */
 
 	/**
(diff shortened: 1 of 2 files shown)


inverse:
That additional case seems like a more complex example which I am not sure how is best to handle. If it's consistent between various tracks there could be an additional filter to remove this.

Are there any reasons where we would want such meta information scrobbled though? e.g. indicating that it was a "remastered" addition listened?

PR #1729, opened 2018-09-21
Update Radioparadise.js for support that new web player.

JeonHyungJin:
Radioparadise.com was updated. So now this program does not support [Radioparadise](https://www.radioparadise.com) web player.
Issue #1718  talked about this. So i updated this web site`s js file.

I made 2 PR for this, but at that time my code has bugs. haha..
So i fixed them all!
Now [Radioparadise](https://www.radioparadise.com) support this program!

Thanks :)

PR #1734, opened 2018-09-24
RadioParadise.js modify to new Wep player

JeonHyungJin:
I modifired RadioParadise.js file to new web player!
about this [PR](https://github.com/web-scrobbler/web-scrobbler/pull/1729)
And at that PR i modified connectors.js file for running this program in "earbits.com" site when user use without "www".
Is that ok to commit again about that part?

### DIFF
--- a/src/connectors/radioparadise.js
+++ b/src/connectors/radioparadise.js
@@ -1,9 +1,11 @@
 'use strict';
 
-Connector.playerSelector = '#header';
+Connector.playerSelector = '#controls';
 
 Connector.artistTrackSelector = '#nowplaying_title > b';
 
 Connector.trackArtSelector = '#nowplaying_title > img';
 
-Connector.isPlaying = () => $('#play_button').hasClass('button_active');
+Connector.isPlaying = () => {
+	return $('#play_button').hasClass('button_active');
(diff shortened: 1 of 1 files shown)


inverse:
What id the new web player? From what I can see that site has:

- header player
- slideshow player
- Audio player
- Beta player

I think changing just for the new one isn't the right way to go. Supporting multiple player elements would be the best way to do this. 

To do this you should check for the existence of an element and setup the behaviour if that's found for that player.

JeonHyungJin:
Web site which you are talking about is RadioParadise.com?!

inverse:
radio paradise eclectic commercial free internet radio


See the screenshot

JeonHyungJin:
okay i see. I will do again😊

inverse:
Have a look at how we do it for the Deezer connector.

https://github.com/web-scrobbler/web-scrobbler/blob/f627a7076f2a17d3989f8a3ad231d49065158b73/src/connectors/deezer.js

Basically loading the right logic based on a DOM element existence.

inverse:
Closing as resolved in another PR. Feel free to extend to add support for other sites as required

PR #1735, opened 2018-09-27
Use a DOM-injcted file for deezer.com for better data 

lbrueckner:
After I'd adapted the deezer connector for the new layout some days ago I was not very happy with the result, so I have build this new connector which is using the internal (unofficial) deezer api. I believe this is better for various reasons:

- Not DOM-dependent, it is working out of the box with the old and the new (... and future) page layout.

- The internal api is stable since three years, probably longer.

- More complete data as the albumname and the songId (used as uniqueId) are not available in the DOM but through the api.

- The data provided by the api distinguishes between main artists and featured artists. When the old connector is submitting  an ugly list of comma-seperated artistnames, the new connecter is building a more lastm-conform representation by adding the featured artists to the title. Examples:

  old: https://www.last.fm/music/Tyler,+The+Creator,+Rex+Orange+County,+Anna+Of+The+North/_/Boredom (3899 scrobbles)
  new: https://www.last.fm/music/Tyler,+the+Creator/_/Boredom+(feat.+Rex+Orange+County+&+Anna+Of+The+North) (12.9K scrobbles)

  old: https://www.last.fm/music/Underworld,+Iggy+Pop/_/Bells+&+Circles  (317 scrobbles)
  new: https://www.last.fm/music/Underworld+&+Iggy+Pop/_/Bells+&+Circles (3664 scrobbles)

I'm using this for a week now and I'm very pleased with the results :-)

I'm not sure if the retry-loop in initConnector() is necessary, it was never executed so far. Maybe the core source is taking care that the page is ready before injecting?

### DIFF
--- a/src/connectors/deezer-dom-inject.js
+++ b/src/connectors/deezer-dom-inject.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const DEEZER_MAIN_ARTIST = '0';
-const DEEZER_FEATURED_ARTIST = '5';
-
 initConnector();
 
 function initConnector() {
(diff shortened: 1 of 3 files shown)


src/connectors/deezer-dom-inject.js
```diff
@@ -0,0 +1,106 @@
...
+		artists: item.ARTISTS
+	};
+}
+
+function getTrackArt(pic) {
+	return `https://e-cdns-images.dzcdn.net/images/cover/${pic}/264x264-000000-80-0-0.jpg`;
+}
+
+function lastfmifyArtists(item) {
+
```
inverse: Extra space

src/connectors/deezer-dom-inject.js
```diff
@@ -0,0 +1,106 @@
...
+	let artists = getArtistsnamesByRole(item, '5'); // featured artists only
+	if (artists.length === 0) {
+		return title;
+	}
+
+	return `${title} (feat. ${getArtistsList(artists)})`;
+}
+
+function getArtistsnamesByRole(item, roleId) {
+
```
inverse: Extra spacing

src/connectors/deezer-dom-inject.js
```diff
@@ -0,0 +1,106 @@
...
+		title = `${title} ${item.VERSION}`;
+	}
+
+	// don't add featured artists from the item object to the title
+	// if there is already feature information included
+	if (title.search(/\s[([]?[f|F](eat\.|eaturing)\s([^ ]+)/i) !== -1) {
+		return title;
+	}
+
+	let artists = getArtistsnamesByRole(item, '5'); // featured artists only
```
inverse: Could we define 0 and 5 as variables to give more context.

src/connectors/deezer-dom-inject.js
```diff
@@ -0,0 +1,106 @@
...
+
+function getArtistsnamesByRole(item, roleId) {
+
+	return item.ARTISTS.filter((a) => a.ROLE_ID === roleId) // 0: main artist; 5: featured artist
+		.sort((a, b) => a.ARTISTS_SONG_ORDER < b.ARTISTS_SONG_ORDER) // respect ordering
+		.map((a) => a.ART_NAME) // array of artistnames from arry of items
+		.filter(removeDublicateArtists);
+}
+
+function removeDublicateArtists(artist, index, artists) {
```
inverse: Typo

src/connectors/deezer-dom-inject.js
```diff
@@ -0,0 +1,106 @@
...
+
+function lastfmifyArtists(item) {
+
+	let artists = item.ARTISTS;
+	artists = getArtistsnamesByRole(item, '0'); // main artists only
+	if (artists.length <= 1) {
+		return item.ART_NAME;
+	}
+
+	return getArtistsList(artists);
```
inverse: Why call this? this returns a list right?

lbrueckner:
Thanks a lot for your feedback!

I've removed the extra newlines, fixed the typo and added two constants for the values used by deezer to seperate main artists from featured artists.


`return getArtistsList(artists);`

The parameter is an array of artistnames and the return value is a string containing the artistnames in a list, example:

['Kilroy', 'John Doe', 'The Ramblers']  ->  'Kilroy, John Doe & The Ramblers'


And, of course, it would be good to get agreement of other deezer users.

Paszt:
Last FM gets a lot of data from Musicbrainz and the correct style is to have the featured artist be part of the artist name.  
https://www.last.fm/music/Tyler,+The+Creator+feat.+Rex+Orange+County+&+Anna+Of+the+North/_/Boredom  
https://musicbrainz.org/doc/Style/Artist_Credits#Featured_artists

inverse:
Any other Deezer uses able to comment?

lbrueckner:
I'm using the injected connector for several thousand scrobbles now with great success. But recently I've found some inconsistencies in the deezer API responses. First I implemented fixes/workarounds, but it was starting to get quite complicated and in the end I think it is not worth the hassle.

My updated code is not longer trying to make sense of the featured artists information in the API response. It is now using the ART_NAME property which is a simple string.

inverse:
We can already iterate and your solution already looks much more elegant :)

lbrueckner:
Thanks!

PR #1737, opened 2018-10-03
Added a new connector for Joox, Tencent's music site for Myanmar

mmhan:
It's pretty straightforward, apart from Album name data.

The player currently doesn't display album name. Hence, an ajax request was needed to figure out the album name. It's done asynchronously and cached to avoid loading the page repeatedly.


src/connectors/joox.js
```diff
@@ -0,0 +1,80 @@
...
+	if (!isAlbumDataLoaded(_albumData)) {
+		_albumData.loaded = loadAlbumData((album_name) => _albumData.name = album_name);
+	}
+
+	return _albumData.name;
+};
+
+Connector.applyFilter(filter);
+
+// minified zg2uni of Rabbit converter
```
inverse: To me this feels like it should be included as a 3rd party dependency and available not just here.

src/connectors/joox.js
```diff
@@ -0,0 +1,80 @@
...
+
+
+
+const defaultAlbumData = {
+	path: undefined,
+	loaded: false,
+	name: undefined
+};
+
+const playerSelector = '#jooxPlayerBar';
```
alexesprit: I'd get rid of these constants and assign selector strings to `Connector.%someProperty%` directly.

src/connectors/joox.js
```diff
@@ -0,0 +1,80 @@
...
+const artistSelector = '#lblSinger';
+const trackSelector = '#lblSongName';
+const playBtnSelector = '#btnPlay';
+
+let _albumData = $.extend({}, defaultAlbumData);
```
alexesprit: Please, use Object.assign instead of this.

mmhan:
> Hi, sorry for the late comment — I had my PC broken.
> 
> It's ok, but:
> 
> * That "minified" part should be non-minified because of store reviews; and it should be placed after all functions for better readabilily (main code first, then code from libs etc);
> * Also, we declare functions using `function` keyword; we don't use Function Expression;
> * `$.get` can be replaced with `fetch`.
> 
> **Edit**: the connector test should be removed, as there're no connector tests.

I'm trying to implement this by using `import`, but I'm hitting a bump. Can you give me some pointers?

alexesprit:
`import` is not required to fix the points I listed.

You can use `fetch` w/o `import` keyword (see `gaana.js` for example).

Function declaration is a way to define functions like this:
```js
function zg2uni (output) {
    ...
}
```
Your way is fine too, but we should keep the unified code style.

To unminify the lib part you can use either source code of library, or any online service.

mmhan:
> You can use `fetch` w/o `import` keyword (see `gaana.js` for example).

> Function declaration is a way to define functions like this:

Already fixed those two. I'll see if there's an online service that I can use for zg2uni.

inverse:
Any updates on this one?

mmhan:
Not quite sure what this error is about.

```
Running "unused_files" task
Fatal error: Found 1 unused files: connectors/joox.js
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
##[error]Process completed with exit code 1.
```

PR #1738, opened 2018-10-04
Implement 1001tracklists Connector

### LINKED ISSUE #1695: 1001tracklists.com connector request
https://www.1001tracklists.com/tracklist/1uuuhf59/san-holo-autolaser-bitbird-radio-020-2018-08-10.html
Example one

WhiteAls:
Resolves #1695
It seems that I used a few ugly tricks. Or it only just seems...

### DIFF
--- a/src/connectors/1001tracklists.js
+++ b/src/connectors/1001tracklists.js
@@ -2,15 +2,16 @@
 
 let nextCue;
 let cue;
-let mixDuration = Util.stringToSeconds($('.mediaTabItem:first span').text().split('[')[1].trim().slice(0, -1));
+let currentTime;
+const mixDuration = Util.stringToSeconds($('.mediaTabItem:first span').text().split('[')[1].trim().slice(0, -1));
 
-Connector.playerSelector = '.playerWidgetFields';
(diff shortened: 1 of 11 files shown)


src/connectors/1001tracklists.js
```diff
@@ -0,0 +1,48 @@
...
+	let text = $('.cPlay:first meta[itemprop="name"]').attr('content');
+	return Util.splitArtistTrack(text);
+};
+
+Connector.getCurrentTime = () => {
+	return currentTime - cue;
+};
+
+Connector.getDuration = () => {
+	switch (true) {
```
inverse: Def some magic going on here to work around things ;)

WhiteAls:
Sadly it can't work with every mix, because cues needs to be filled. It shall be done by 1001tl users manually. There a plenty mixes that have a tracklist without cues, like your example. Cues is needed to calculate track duration. Well, I can do some dirty hack for tracklists without cues. Like, estimate duration by dividing mix time to number of songs in it. But it doesn't seem like a good idea.

inverse:
It's a step forward - perhaps a lot of magic going on and doesn't cover all cases.

PR #1741, opened 2018-10-05
Connector Soribada.js fixed

NamJinWoo:
I tested on my local laptop after compiling and building it, but it didn't work properly.

Therefore, I fixed properly and tested it. Now it works properly!

### DIFF
--- a/src/connectors/radioparadise.js
+++ b/src/connectors/radioparadise.js
@@ -1,9 +1,13 @@
 'use strict';
 
-Connector.playerSelector = '#header';
+Connector.playerSelector = '.player-controls-wrapper';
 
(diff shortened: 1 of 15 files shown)


NamJinWoo:
it didn't work because i didn't put "www" in front of the url.
Also, it couldn't get the image because i wrote "http:" in front of the image source even the img url has http already.

please check!!

src/connectors/radioparadise.js
```diff
@@ -1,9 +1,13 @@
 'use strict';
 
-Connector.playerSelector = '#header';
+Connector.playerSelector = '.player-controls-wrapper';
```
inverse: This file should not be touched in this PR

PR #1742, opened 2018-10-05
Automated deployment for extension on tag

inverse:
This change should enable us to automatically deploy via travis to the chrome store when we tag a release.

### Changes
Will require setting up some environment variables on Travis to contain the secrets

#### Chrome
`CHROME_CLIENT_ID`
`CHROME_CLIENT_SECRET`
`CHROME_REFRESH_TOKEN`

#### Github
 `GITHUB_TOKEN`

#### Amo
 `AMO_ISSUER`
 `AMO_SECRET`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -346,12 +346,6 @@ module.exports = function(grunt) {
 		}
 
 		grunt.task.run(`bump-only:${releaseType}`);
-		// Patch releases are in vX.X.X branch, so there's no reason
-		// to make changelogs for them.
-		if (releaseType !== 'patch') {
-			grunt.task.run('publish:add0n');
-		}
-
 		grunt.task.run('bump-commit');
 	});
 
@@ -362,6 +356,16 @@ module.exports = function(grunt) {
 	 * @param {String} releaseType Release type that 'grunt-bump' supports
 	 */
 	grunt.registerTask('release-local', (releaseType) => {
+		if (!releaseType) {
+			grunt.fail.fatal('You should specify release type!');
+		}
+
+		// Patch releases are in vX.X.X branch, so there's no reason
+		// to make changelogs for them.
+		if (releaseType !== 'patch') {
+			grunt.task.run('publish:add0n');
+		}
+
 		grunt.task.run(`release:${releaseType}`);
 
 		grunt.task.run(['publish:chrome', 'publish:firefox']);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -39,9 +39,9 @@ module.exports = function(grunt) {
 	const htmlFiles = [`${srcDir}/options/*.html`, `${srcDir}/popups/*.html`];
 	const cssFiles = [`${srcDir}options/*.css`, `${srcDir}/popups/*.css`];
 
-	const webStoreConfig = loadConfig('./.publish/web-store.json');
-	const githubConfig = loadConfig('./.publish/github.json');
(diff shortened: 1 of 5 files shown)


deploy.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+./node_modules/.bin/grunt publish:chrome
```
inverse: @alex-volga can you please add the commands we do for a release. e.g. bump/dump etc

Gruntfile.js
```diff
@@ -427,4 +427,46 @@ module.exports = function(grunt) {
 
 		return {};
 	}
+
+	function loadWebStoreConfig() {
+		if (isTravisCi) {
```
inverse: This is so we can use the "old way" but if we are on travis do it travis style ;)

stephanebruckert:
This is awesome! ping @alexesprit

alexesprit:
The `release` task should not run publish tasks. In my vision with automatic deployment you want to run `release` task to make a release. The task will create a "bump" commit, a new tag contains a version number, and will push changes (IIRC, there's no push now, but it can be added). And Travis CI will do the rest work, i.e. deploy packages via `publish` tasks on tag create.

Everything else looks fine for me.

inverse:
@alexesprit right now I'm calling `./node_modules/.bin/grunt publish:chrome` on the `deploy.sh` file. Is there anything else that needs to be included here?

inverse:
@alexesprit How would that tag align with what's already been drafted on GH?

So are you saying the `release` task needs to be reworked to just `bump` and `tag`? I understand the version needs to be bumped before tagging.

alexesprit:
> How would that tag align with what's already been drafted on GH?

It's only thing I have no idea how to automate. You should manually assign a new tag to a draft after pushing this tag. I used this way when I released new versions.

> So are you saying the release task needs to be reworked to just bump and tag? I understand the version needs to be bumped before tagging.

Yes, you're right. Otherwise, the package will be uploaded twice. Of course, you can bump and push things manually, and in this case the package will be uploaded by Travis CI only.

> We still need you to put the credentials into env vars on travis to automate the deployment.

I understand this; I've managed how to add env vars to Travis CI project. I'll generate new credentials and set up the project. I'll let you know when I do it, so we can release new version and see how the automatic things work.

inverse:
My assumption was that it would work like this.

- Develop + push
- Bump + push
- Tag - on GH
- Build kicked off on travis

or am I mistaken?

alexesprit:
1. Develop + push changes.
2. When the release is ready to publish, you run `grunt release`, which bumps version, add new tag and push changes to GitHub.
3. Travis CI detects new tag and deploys packages.

alexesprit:
> (IIRC, there's no push now, but it can be added)

Just found out the default configuration is to push changes to upstream after bump and commit. I didn't change this, so on `release` task changes will be pushed to upstream.

inverse:
I added a new task called `release-local` which will provide the same experience as release but with the publish steps included. It will however cause another build to be triggered from travis with the tag.

There is one final part I am not sure about and that's the publish addon part. For me this feels like it should be part of the deployment. e.g. hooked up in `deploy.sh`. What do you think?

inverse:
I've merged this for now - we can discuss how to handle the addon part later.

inverse:
My bad - I understand it now. Changed back with: https://github.com/web-scrobbler/web-scrobbler/pull/1776

PR #1764, opened 2018-10-16
Implement warp-aphextwin connector

stephanebruckert:
News: https://djmag.com/news/aphex-twins-entire-warp-discography-can-now-be-streamed-free
Website: https://aphextwin.warp.net

The connector for it is quite straightforward but let me know if there is anything to change, thanks.

### DIFF
--- a/src/connectors/warp-aphextwin.js
+++ b/src/connectors/warp-aphextwin.js
@@ -2,12 +2,19 @@
 
 Connector.playerSelector = '.player';
 
-Connector.playButtonSelector = '.track-controls .player-button.play-track';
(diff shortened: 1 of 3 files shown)


stephanebruckert:
Thanks for the quick feedback! I just added the album and current time

stephanebruckert:
@inverse thanks for merging. Do you know if/when we can create a release?

inverse:
Unfortunately that's out of my control... I created a pr so that we can automate this on tag creation which should solve this though :) it's pending some changes from someone with deployment power :)

PR #1770, opened 2018-10-25
Fixing Recognition Problems

calvinc64:


calvinc64:
Sorry, there are some Errors, tried to fix them with code from the other apps, but yeah.... you will see what Ive tried.

calvinc64:
So, Just looked at the Errors from Tarvis, now fixed ^^

src/connectors/listen.moe.js
```diff
@@ -3,7 +3,14 @@
 Connector.playerSelector = '#app';
 
 Connector.artistSelector = '.playerContainer.columns.mini .player .player-song-artist';
-
 Connector.trackSelector = '.playerContainer.columns.mini .player .player-song-title';
 
+Connector.getArtistTrack = () => {
+	return $('.player-song-title')[0].innerText.split(' [')[0];
```
inverse: Can you give some examples of tracks which this helps?

src/connectors/listen.moe.js
```diff
@@ -3,7 +3,14 @@
...
+};
+
+Connector.getArtist = () => {
+	return $('.player-song-artist-container')[0].innerText.replace(' ,', ',');
```
inverse: Same with this one.

src/connectors/listen.moe.js
calvinc64: Ok... The page sometimes displays the Show (From the Series the Opening is from) and this is in this format: artist - Trackname [Some aditional show info that confuses lastfm]
and in the artist field if multiple groups did work on one track is weird too... let me search an example...

Here an nice example, where both the title and the Artists are bad:

inverse:
@calvinc64 any update?

calvinc64:
Sorry for not leaving any update for so long, did not have time enough to spend for git at the moment, but I will look into the Problem this weekend again!

calvinc64:
Well now its just showing me this...
grafik
And yeah did test it on Firefox 66.0.3 (64B on Windows 10)

alexesprit:
@calvinc64 It was fixed in v2.16.3.

alexesprit:
I'm closing this one as outdated.

PR #1771, opened 2018-10-27
Implement residentadvisor connector

stephanebruckert:
Hey,

Here is a connector for https://residentadvisor.net. It is unusually complex because the website does not have a single player, but 7 different types for which the track information is not always located at the same place. Test examples (also added in the code):

 - https://www.residentadvisor.net/tracks ("Popular tracks" at the top)
 - https://www.residentadvisor.net/tracks/931382 ("Popular tracks" at the bottom)
 - https://www.residentadvisor.net/tracks ("Archived tracks" in the middle)
 - https://www.residentadvisor.net/tracks/931382 ("Single track" in the middle)
 - https://www.residentadvisor.net/record-label.aspx?id=15687&show=tracks ("Label tracks" in the middle)
 - https://www.residentadvisor.net/dj/hotsince82/tracks ("DJ tracks" in the middle)
 - https://www.residentadvisor.net/dj/secretsundaze/top10?chart=214484 ("Chart tracks" in the middle)

Hopefully I haven't forgotten any of them but I will open a new PR if I find any more. RA also embeds soundcloud songs in some places, but this PR does not take care of it.

The track info we have access to are: track name, track artist and cover. I think I found where to get the track duration, but I'm not sure how to get it from inside the connector. If you start playing a track and type the following in the console, it gives the track duration:

    soundManager.sounds[Object.keys(soundManager.sounds)[0]].duration

But when putting a  debugger in the connector, it's undefined. How to include it?

Thanks!

### DIFF
--- a/src/connectors/residentadvisor.js
+++ b/src/connectors/residentadvisor.js
@@ -2,6 +2,23 @@
 
 const domain = 'https://www.residentadvisor.net';
 
+let PLAYER_TYPES = {
+	// https://www.residentadvisor.net/tracks ("Popular tracks" at the top)
+	// https://www.residentadvisor.net/tracks/931382 ("Popular tracks" at the bottom)
+	POPULAR: 'Popular tracks',
+	// https://www.residentadvisor.net/tracks ("Archived tracks" in the middle)
+	ARCHIVED: 'Archived tracks',
+	// https://www.residentadvisor.net/tracks/931382 ("Single track" in the middle)
+	SINGLE: 'Single track',
+	// https://www.residentadvisor.net/record-label.aspx?id=15687&show=tracks (Label tracks)
+	// https://www.residentadvisor.net/dj/hotsince82/tracks (DJ tracks)
+	DJ_OR_LABEL: 'DJ tracks or label tracks',
+	// https://www.residentadvisor.net/dj/secretsundaze/top10?chart=214484 (Chart tracks)
+	CHART: 'Chart track',
+
+	UNKNOWN: 'Player not found'
+};
+
 Connector.playerSelector = '.content-list';
 
 Connector.playButtonSelector = '.controls .play';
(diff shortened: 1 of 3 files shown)


src/connectors/residentadvisor.js
```diff
@@ -0,0 +1,103 @@
...
+		}
+	});
+	return track;
+}
+
+function typeOfTrack(trackContainer) {
+	if ($(trackContainer).is('article')) {
+		// https://www.residentadvisor.net/tracks ("Popular tracks" at the top)
+		// https://www.residentadvisor.net/tracks/931382 ("Popular tracks" at the bottom)
+		return 'popular';
```
inverse: Could we capture this "type" into some sort of enum object?

src/connectors/residentadvisor.js
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if ($(trackContainer).is('li') && $(trackContainer).parent().hasClass('tracks')) {
+		// https://www.residentadvisor.net/record-label.aspx?id=15687&show=tracks (Label tracks)
+		// https://www.residentadvisor.net/dj/hotsince82/tracks (DJ tracks)
+		return 'djOrLabel';
+	}
+	if ($(trackContainer).is('li') && $(trackContainer).parent().hasClass('chart')) {
+		// https://www.residentadvisor.net/dj/secretsundaze/top10?chart=214484 (Chart tracks)
+		return 'chart';
+	}
```
inverse: I would return `unknown` on failure to resolve type.

src/connectors/residentadvisor.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+Connector.getTrack = () => {
+	let track = getTrackContainer();
+	switch (typeOfTrack(track)) {
```
inverse: Would it make more sense to call `typeOfTrack` once and bind `Connector.getTrack = () => {` for each type once?

stephanebruckert: Are you thinking of something like https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/daytrotter.js where the player is initialised once when the page loads? I tried and I realised that in our case the player does not depend on the URL and we don't know which type of player will be used in advance (some pages have different types of player).

stephanebruckert: Yes one page can have multiple different players, and one player can be on multiple pages. To avoid defining which page has which player(s), we detect the type of player depending on the context of the "pause/play" button.

1. Get the playing button
2. Get the parent player of that button
3. Detect what kind of player it is
4. Depending on the type of player, find track information

stephanebruckert:
Thanks for your feedback, let me know if that looks good.

src/connectors/residentadvisor.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+Connector.getUniqueId = () => {
+	return getTrackContainer().find('.play.player').attr('data-trackid');
+};
+
+Connector.isPlaying = () => $('.play.paused').length > 0;
+
+Connector.getTrackArt = () => {
+	return `https://www.residentadvisor.net${$(getTrackContainer()).find('img[src^="/images/cover/"]').first().attr('src')}`;
```
inverse: Can we use the domain cost from the top?

inverse:
@stephanebruckert any update?

stephanebruckert:
The connector seems to behave correctly but the default integration test `shouldBehaveLikeMusicSite` doesn't seem to pass. I need to have a better look and will open an issue if needed.

PR #1773, opened 2018-10-29
Provide option to disable scrobbling podcasts on Google Music.

jaccarmac:


jaccarmac:
A couple things aren't working quite yet and I'd like suggestions on how to get them in shape for a merge.

[] Internationalization doesn't seem to be working. Do I need to add the entries to all four language files?
[] The option doesn't have a default value set and it looks like connector-specific options have defaults set-up yet. Shall I add a similar hook to initialize those?

inverse:
Awesome! 

- It looks like point one is working for me. I've got en as my locale and it's displaying correctly.
- Having some sane default sounds like a good idea - not sure where this would make the most sense.

jaccarmac:
Locale was indeed some kind of caching issue. Latest commit should be ready for merge once Travis decides it's linted well enough.

inverse:
Super! Thanks for the contribution :)

PR #1786, opened 2018-11-04
Fix disallowing scrobbling & don't scrobble Spotify ads

### LINKED ISSUE #1599: Spotify web player ads scrobbled
After 10-15 minutes of music, the first ad usually plays and the Chrome plugin sees it as a song. If I get to it soon enough, I can just tell it to skip, otherwise, I have to hunt them down later in my Last.fm feed and delete them.

SeijiSuenaga:
Fixes #1599.

If a connector wants to prevent certain tracks from being scrobbled (e.g. audio ads on Spotify), it can implement `isScrobblingAllowed()` to return false when an unwanted track is playing. However, delayed execution of throttled state change processing can still allow the unwanted track to be scrobbled. In this scenario, the track will scrobble even while paused.

How this happens:
1. `stateChangedWorkerThrottled()` schedules a delayed execution of `stateChangedWorker()`.
2. An unwanted track starts playing.
3. `isScrobblingAllowed()` becomes false.
4. `onStateChange()` resets state, sets a flag to prevent resetting state again while scrobbling isn't allowed, and returns early.
5. Throttled `stateChangedWorker()` executes and processes the unwanted track, overriding the reset.
6. `onStateChange()` continues returning early without resetting state again or invoking `stateChangedWorker()`.
7. If the user pauses the unwanted track, the pause event won't be processed, so it'll scrobble even while paused.
8. Unwanted track ends and a valid track starts playing.
9. `isScrobblingAllowed()` becomes true, allowing `onStateChange()` to execute fully again. Behavior returns to normal.

To fix, we can check `isScrobblingAllowed()` every time `stateChangedWorker()` executes, including delayed execution due to throttling.

### DIFF
--- a/src/connectors/spotify.js
+++ b/src/connectors/spotify.js
@@ -25,10 +25,6 @@ Connector.isScrobblingAllowed = () => {
 };
 
 function isMusicPlaying() {
-	if (Connector.getArtist() !== 'Spotify') {
-		return true;
-	}
-
 	// When ad is playing, artist URL is like "https://shrt.spotify.com/XXX",
 	// otherwise URL leads to an artist page "https://open.spotify.com/artist/YYY".
 	let artistUrl = $('.track-info__artists a').attr('href');
--- a/src/core/content/connector.js
+++ b/src/core/content/connector.js
@@ -450,6 +450,13 @@ function BaseConnector() {
 	 * Function for all the hard work around detecting and updating state.
 	 */
 	this.stateChangedWorker = () => {
+		if (!this.isScrobblingAllowed()) {
+			this.resetState();
+			return;
+		}
+
+		isStateReset = false;
+
 		let changedFields = [];
 		let newState = this.getCurrentState();
 
@@ -484,7 +491,6 @@ function BaseConnector() {
 			}
 			// @endif
 		}
-
 	};
 
 	/**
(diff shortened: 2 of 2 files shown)


SeijiSuenaga:
Any thoughts on this?

src/connectors/spotify.js
```diff
@@ -25,10 +25,6 @@ Connector.isScrobblingAllowed = () => {
 };
 
 function isMusicPlaying() {
-	if (Connector.getArtist() !== 'Spotify') {
```
inverse: What does removing this achieve?

inverse:
So this change is basically pushing the check further down into the worker?

src/connectors/spotify.js
SeijiSuenaga: Many Spotify ads have artist values other than "Spotify", so this condition currently allows those ads to be scrobbled.

SeijiSuenaga:
Thanks! Yes, this just pushes the check further down to avoid being overridden by the throttled worker.

src/connectors/spotify.js
inverse: Where is the part that stops the disallow then? Is there anything in the DOM to highlight this?

inverse:
@alexesprit any input on this core change?

src/connectors/spotify.js
SeijiSuenaga: With this code removed, the remainder of `isMusicPlaying()` stops the disallow when the artist link href is an artist page instead of an ad URL. That URL logic still works well, but the "Spotify" artist name logic doesn't work and prevents the URL logic from being reached.

SeijiSuenaga:
Bump, been a while.

inverse:
sorry for the delay - thanks for the investigation work :+1:

Issue #1796, opened 2018-11-09
Single dashes in song title trip up parser


reitzig:
Listening to [Sock 'n' Doll on Youtube Music](https://music.youtube.com/watch?v=l3P9Nv3mHZI&feature=share).

![image](https://user-images.githubusercontent.com/4246780/48256228-3fc02700-e40f-11e8-9354-d88600c0092e.png)

### Expected behavior

Scrobbles correctly.

### Current behavior

Detects only the part between single quotes as title.

![image](https://user-images.githubusercontent.com/4246780/48256216-3a62dc80-e40f-11e8-83b5-b9b8f6ab4cb2.png)

### Environment

* Version used: 2.8.0
* Browser name and version: FF 63.0 (64bit), Ubuntu 18.04

PR #1797, opened 2018-11-09
Handle single quotes in Youtube titles

reitzig:
Test case for #1796


reitzig:
Hm, the problem seems to lie elsewhere.

reitzig:
Not knowing the use cases for that filter, it's hard to make a guess as to what's necessary. Having `.*` in there begs for trouble, imho.

Would it help to solve those use cases by adding a suitable regexp to `youtubeTitleRegExps` (in a suitable position)? Meaning, if the part between quotes is a suitable artist+title thing, use that; if not, try the next pattern (the existing ones).

inverse:
Exactly - it's really hard to gauge the usage.

Unfortunately the way the filters work right now is that'll run them all through a title before sending off to last.fm - which IMHO makes sense to some degree.

I think having some min-length could work as these right now will assume the track title to be empty if it's just single quotes.

inverse:
Looks like the regex needs whitespace around the `'` so it's not as bad as it seems. Only the ones with `'n'` would fail.

inverse:
Merged the other PR that should fix it. I wonder why this was still passing before with what you added :/ Will need to investigate.

inverse:
So it turns out that that test case you added just covers https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/content/util.js#L32

inverse:
Closing due to other test case covering this. Eventually we should tidy up this logic as there are some filters going on in util that should be consolidated.

PR #1801, opened 2018-11-11
Passthrough URL

inverse:
- Passthrough Origin URL for listenbrainz

### DIFF
--- a/src/core/background/object/song.js
+++ b/src/core/background/object/song.js
@@ -225,8 +225,8 @@ define((require) => {
 		 *
 		 * @return {String} source URL.
 		 */
-		getSourceUrl() {
-			return this.parsed.sourceUrl;
+		getOriginUrl() {
+			return this.parsed.originUrl;
 		}
 
 		/**
--- a/src/core/background/object/song.js
+++ b/src/core/background/object/song.js
@@ -59,6 +59,7 @@ define((require) => {
 				currentTime: parsedData.currentTime,
 				isPlaying: parsedData.isPlaying,
 				trackArt: parsedData.trackArt,
+				sourceUrl: parsedData.sourceUrl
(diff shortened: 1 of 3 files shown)


Freso:
Ping: @paramsingh - see discussion starting at https://github.com/web-scrobbler/web-scrobbler/pull/1679#issuecomment-437685104

Currently the PR suggests using `origin_url`, but after my last comment on the other PR, I think the `{track,artist,release}_url` scheme might be a more suitable one. @paramsingh is the lead dev of ListenBrainz though, so maybe he has another view. :)

paramsingh:
Ideally, we'd prefer to know `track_url`, `artist_url` and `release_url` specifically. But if that requires a lot of work, it is probably  okay to submit as much data with the `origin_url` key and then eventually we'll look into matching the data via the [messybrainz](https://messybrainz.org) project.

paramsingh:
So I'd personally merge this for now, and maybe add an issue to submit more granular URLs, so that users can submit as much data to ListenBrainz as possible.

inverse:
@paramsingh thanks for the feedback and insights into the other projects you're working on. Definately something that can be improved to allow "connectors" to provide additional meta data. :+1:

MonkeyDo:
Sorry it took this long to be used on the ListenBrainz side, but we just implemented Youtube as a source for the ListenBrainz player, and it now detects and extracts the Youtube url from listens. Thanks!

For reference: https://github.com/metabrainz/listenbrainz-server/pull/882

alexesprit:
@MonkeyDo Hi! Thanks for the info.

By the way, it wasn't sent to ListenBrainz actually (was fixed recently in 393b8c4fa8d605505a85fae4c692066ac0375cf6).

Issue #1809, opened 2018-11-18
from Jamendo it no longer srobbelt


onkellu:
for almost 2 weeks there is no more scrobbelt from Jamendo to Last.fm

### Expected behavior
the page of Jamendo has changed and I ask for a review
i am not technically experienced enough to give a solution to the problem

### Current behavior
Nothing happens that I can see.
the scobbling should be restored

### Possible solution
i'm not a technician and i can't conjure a solution out of a hat.
??

### Steps to reproduce (for bugs)
https://www.jamendo.com/album/177674/el-aullido-del-lobo
i can't provide code

My Browser Firefox 63.0 64Bit for Linux Mint-1.0, my operating system is Linux Mint 18.1

1.
2.
3.

### Context
What should I do if the scrobble doesn't work?

### Steps for resolving the issue

I made
it didn't help.

- [ ] Reauthenticate on Last.fm

- [ ] Try to use unpacked extension from latest master

- [ ] Check other issues for duplicates

### Environment
<!---
* Version used:
* Browser name and version: My Browser Firefox 63.0 64Bit for Linux Mint-1.0, my operating system is Linux Mint 18.1

SORRY, i can't speak english and have to help myself with a translated one, which is not so easy, because i only understand half of it

NamJinWoo:
I think Play button and TrackArt is not working.

So, I fixed it like this:
Connector.isPlaying = () => {
	return $('.js-player-play-pause > .icon-pause').length !== 0;
};

Connector.getTrackArt = () => {
	return `${$('.hero-cover').find('img').attr('src')}`;
};

I compiled it in my local pc, and it works now.

If you need more help, please comment here.

Thanks:)

onkellu:
Hi NamJinWoo, First of all, thank you for your efforts.
i can't do anything with your signs because i don't know where to find them and then change them.
i think it has changed on the jamendo side what i can't see as a layman, also i have no idea of technical connections and their improvement
i also think the problem is not with me but with jamendo.

the problem what had existed until recently with the page You Tube has then been corrected without doing mine and it works again...
This was done by programming by Git Hub.

sorry my english is very bad and I have translated with translater.

KwonDae:
Hi onkellu, I hope this comments will help you.
in` /src/connectors/jamendo.js`

> Connector.playButtonSelector = '.js-player-play-pause > .icon-play';

this part should change by 

> Connector.isPlaying = () => {
[...]

PR #1811, opened 2018-11-23
Issue #1809 Jamendo scrobbler error fixed

NamJinWoo:
Issue #1809 Jamendo scrobbler error fixed

Issue https://github.com/web-scrobbler/web-scrobbler/issues/1809

Play button and TrackArt were not working.

So, I fixed it like this:
Connector.isPlaying = () => {
return $('.js-player-play-pause > .icon-pause').length !== 0;
};

return ${$('.hero-cover').find('img').attr('src')};
};

I compiled it in my local pc, and it works now.

Thanks:)


NamJinWoo:
I had a conflicts in radioparadise.js because I was trying to solve the problems that had been occured before. I rolled it back to the one before.

Jamendo site error is fixed now.

For the issue #1809 

Thanks:)

src/connectors/jamendo.js
```diff
@@ -10,6 +10,10 @@ Connector.currentTimeSelector = '.hidden-xs > .js-player-position';
 
 Connector.durationSelector = '.hidden-xs > .js-player-duration';
 
-Connector.playButtonSelector = '.js-player-play-pause > .icon-play';
+Connector.isPlaying = () => {
+	return $('.js-player-play-pause > .icon-pause').length !== 0;
+};
 
-Connector.trackArtSelector = '.js-full-player-cover-img';
```
inverse: Wouldn't changing this selector to `.hero-cover img` be better?

NamJinWoo: That also works. I tried to get a "src" from image. I guess both works.

What do you prefer more? If you prefer that one more, please tell me. I will take it over.

NamJinWoo:
I tried what you said, and compiled it in my local laptop. And it works too.

I just changed it into that code, and sent you a PR again.

Please check! Thanks:)

inverse:
Thanks!

PR #1812, opened 2018-11-23
Implements mnet connectors

KwonDae:
I add a connector that mnet music site.
please check my request :) 👍


KwonDae:
Also, I add an album information in Ollehmusic.js.

package-lock.json
```diff
@@ -0,0 +1,9473 @@
+{
```
inverse: Can this be removed? We're using yarn for dependency locking.

src/connectors/mnet.js
```diff
@@ -0,0 +1,19 @@
...
+Connector.currentTimeSelector = '#player .progressControl .timeStart';
+
+Connector.durationSelector = '#player .progressControl .timeEnd';
+
+Connector.isPlaying = () => {
+	return $('.playControl .btnPause').length !== 0;
+};
+
+Connector.getTrackArt = () => {
+	return `${$('#player .albumArea .albumImg').find('img').attr('src')}`;
```
inverse: This could be simplified using the trackArtSelector and using `#player .albumArea .albumImg img`

package-lock.json
KwonDae: I deleted the file `package-lock.json`

I would appreciate it if you could confirm this change.

src/connectors/mnet.js
KwonDae: Okay I'll try it. Thanks you :)

KwonDae: I added the trackArtSelector code.
Please check my commit :)

inverse:
Thanks!

PR #1813, opened 2018-11-23
monkey3 connector done

JeonHyungJin:
I added Monkey3 Player`s connector.
This is one the famous music player in South Korea.
I like this Website!


JeonHyungJin:
I saw your recommendations and i modified that parts as you talked.
Thank you 👍

Issue #1817, opened 2018-12-01
Xiami's design changed


cptchi:
Hi, Xiami just went through a huge redesign and the player isn't on a different tab anymore, so the scrobbler isn't working.

KwonDae:
#1818 
I changed some code about Xiami.
please check my codes and If there's anything wrong, I'd appreciate it if you let me know.

fanmuxxx:
Hi, I know this has been brought up a few weeks ago. However, my scrobbler doesn't work on xiami in the recent 3 days, and it was totally OK last week. Could someone please have a test and let me know?

inverse:
@KwonDae any input on this?

KwonDae:
No, i haven't. Is there any problem in xiami.com?

fanmuxxx:
please see my last comment. My scrobbler doesn't work on https://www.xiami.com/play page

KwonDae:
Because xiami.com/play was changed to xiami.com. 
xiami.com/play doesn't supported until page updated.

cptchi:
fanmuxx, xiami got a new design and therefore the old scrobbler didn't work on it
if you haven't been redirected to it, click the little home icon on the bottom right corner of the xiami page
![new xiami](https://i.imgur.com/VAqM7yJ.png)

fanmuxxx:
@KwonDae @cptchi Thank both for your help. It works for me now.

I preferred the old design but I suppose that now I have to switch to the new version for the scrobbler function.

inverse:
The extension can be made to work with both designs. you just need a way to detect the design and wire up the behavior accordingly.

PR #1818, opened 2018-12-01
Issue #1817 Xiami scrobbler error fixed

KwonDae:
About #1817 , 
Because Xiami's design changed, I changed the some codes in Xiamai.js .
If there's anything wrong, I'd appreciate it if you let me know :)
Thanks.


src/connectors/xiami.js
```diff
@@ -1,22 +1,17 @@
...
 
-Connector.currentTimeSelector = '#J_positionTime';
+Connector.trackArtSelector = '.music .cover-link .active img';
 
-Connector.trackArtSelector = '#J_playerCoverImg';
-
-Connector.getDuration = () => {
-	let text = Connector.getCurrentTime() ? $('#J_durationTime').text() : '';
-	return Util.stringToSeconds(text);
+Connector.isPlaying = () => {
```
inverse: Do you need this? Would it work with just the defined `playButtonSelector`?

src/connectors/xiami.js
```diff
@@ -1,22 +1,17 @@
...
 
-Connector.playButtonSelector = '.play-btn';
+Connector.currentTimeSelector = '.audio-progress .range .bar .handle';
 
-Connector.currentTimeSelector = '#J_positionTime';
+Connector.trackArtSelector = '.music .cover-link .active img';
 
-Connector.trackArtSelector = '#J_playerCoverImg';
-
-Connector.getDuration = () => {
```
inverse: Is there no way to get the duration of the track now?

src/connectors/xiami.js
```diff
@@ -1,22 +1,17 @@
...
-	let text = Connector.getCurrentTime() ? $('#J_durationTime').text() : '';
-	return Util.stringToSeconds(text);
+Connector.isPlaying = () => {
+	return $('.main-control .play-btn').length !== 0;
 };
-
-Connector.getUniqueID = () => $('.ui-track-current .c1').data('id');
```
inverse: IS this no longer relevant?

src/connectors/xiami.js
KwonDae: I removed codes 

`Connector.currentTimeSelector = '.audio-progress .range .bar .handle';`

and add these

```
Connector.getDuration = () => {
	let text = $('.audio-progress .range .bar .handle').val();
	let str = text.split("/");
	return str[0];
};
```

I removed the currentTimeSelector and added getDuration() method.
Please check :)

src/connectors/xiami.js
KwonDae: I don't know how to use getUniqueID. If you let me know, I'll try to reflect it.
Thank you. :)

src/connectors/xiami.js
KwonDae: It doesn't work well without an isPlaying function.

inverse: Can we drop the `playButtonSelector` definition then?

KwonDae:
I modified some codes. 
Please check and If there's anything wrong, I'd appreciate it if you let me know.

src/connectors/xiami.js
KwonDae: I erased the playButtonSeltter and tested it, but there was a service error.

inverse: Really? Weird. this can be used instead of the isPlaying function.

KwonDae: I don't know why this is happening, but it doesn't work properly without either `isPlaying` or `playButtonSeltter` method.

KwonDae: I just drop the `playButtonSector` and tested it and it works. I'll make a new commitment and post it. 👍

PR #1826, opened 2018-12-11
Add Radiozenders.FM support

calvinturbo:


src/connectors/radiozenders.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+Connector.playerSelector = '#jp_container_1';
+
+Connector.getArtistTrack = function() {
+	let artistTrack = $('#player-station-info').text();
+	return Util.splitArtistTrack(artistTrack, [' - '], true);
+};
+
+Connector.isPlaying = () => $('.player-equalizer').is(':visible');
```
inverse:
```suggestion
Connector.playButtonSelector = '.player-equalizer';
```

calvinturbo:
Sorry for the delay but I added the test case and fixed scrobbling for a specific radio station (Q-Music).

src/connectors/radiozenders.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+Connector.playerSelector = '#jp_container_1';
+
+Connector.getArtistTrack = function() {
+	let artistTrack = $('#player-station-info').text();
+	return Util.splitArtistTrack(artistTrack, [' - '], reverse);
+};
+
+Connector.playButtonSelector = '.player-equalizer';
```
inverse: I think this needs some more work. This is used to determine if the player is playing or not. Looking at the site I think we should overload the `Connector.isPlaying` function like this:

```
Connector.isPlaying = () => {
	return $('#jp_container_1').hasClass('jp-state-playing');
};
```

calvinturbo: Changed it, thanks

calvinturbo:
Changed the isPlaying check, time to merge I think

PR #1831, opened 2018-12-12
Privacy consent

inverse:
@alexesprit started working on this to meet FF toc.

image


src/core/background/main.js
```diff
@@ -437,7 +437,16 @@ require([
...
+
+		chrome.tabs.create({
+			url: '/startup/startup.html'
+		  });
 	}
 
 	startup();
+
+
+	// chrome.runtime.onInstalled.addListener(function() {
```
inverse: Was thinking of plugging in here for first time users.

inverse:
@alexesprit got a working prototype ready. def could use some design love.

inverse:
It's now pretty usable - just needs localization support but that could come later I guess.

flip-bs:
thank you a lot <3

PR #1845, opened 2018-12-23
Privacy fixes

inverse:
@alexesprit  can you test this on Firefox? I noticed my previous implementation only worked on chrome and thus is useless.

### DIFF
--- a/src/core/background/main.js
+++ b/src/core/background/main.js
@@ -408,6 +408,16 @@ require([
 	 * Called on the extension start.
 	 */
 	function startup() {
+		chrome.runtime.onInstalled.addListener((object) => {
+			if ('install' !== object.reason) {
+				return;
+			}
+
+			chrome.tabs.create({
+				url: '/startup/startup.html'
+			});
+		});
+
 		Migrate.migrate().then(() => {
 			updateVersionInStorage().then(notifyOfNotableChanges);
 			setupChromeEventListeners();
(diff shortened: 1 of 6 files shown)


src/startup/startup.js
```diff
@@ -29,19 +29,12 @@ require([
 		window.close();
 	}
 
-	chrome.runtime.getPackageDirectoryEntry((root) => {
```
inverse: This didn't work :cry:

src/core/background/main.js
```diff
@@ -437,13 +437,14 @@ require([
...
 		});
-
-
 	}
-
 	startup();
 
-	chrome.runtime.onInstalled.addListener(function() {
+	chrome.runtime.onInstalled.addListener((object) => {
+		if ('install' !== object.reason) {
```
inverse: Prevent updates firing the new tab

inverse:
@alexesprit tested locally seems fine - had to add some work around to give feedback since FF doesn't like window.close() when launched from `chrome.tabs.create` know a workaround?

inverse:
@marsjaninzmarsa Interesting - how is the tabId determined?

marsjaninzmarsa:
@inverse I was going to bet it's returned by `tab.create`, but checked docs and no - Promise is returned. Promise resolving to `Tab` object containing `id` property... c'mon, people, MDN don't hurt! 😂

PR #1847, opened 2018-12-26
Fix async event binding

inverse:
Tested locally and works :+1:

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -24,9 +24,11 @@
 
   "background": {
     "scripts": [
+      "core/background/hooks.js",
       "vendor/require.min.js", "core/background/requirejs-config.js",
       "core/background/main.js"
-    ]
+    ],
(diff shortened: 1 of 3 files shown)


src/manifest.json
```diff
@@ -24,9 +24,11 @@
 
   "background": {
     "scripts": [
+      "core/background/hooks.js",
```
inverse: Not the most elegant way to handle this but works for this case since it's registered first.

inverse:
@alexesprit although not the most elegant solution it works and can hopefully get us back on AMO until we have a better solution.

inverse:
Interesting - Perhaps we could remove this flag since it was previously working without. 

> Background scripts are registered in the manifest under the "background" field. They are listed in an array after the "scripts" key, and "persistent" should be specified as false.

PR #1848, opened 2018-12-30
Issue #1817 Xiami scrobbler error fixed

KwonDae:
About #1817 ,
Because Xiami's design changed, I changed the some codes in Xiamai.js .
If there's anything wrong, I'd appreciate it if you let me know :)
Thanks.


src/connectors/xiami.js
```diff
@@ -1,22 +1,19 @@
...
-
-Connector.trackArtSelector = '#J_playerCoverImg';
+Connector.trackArtSelector = '.music .cover-link .active img';
 
 Connector.getDuration = () => {
-	let text = Connector.getCurrentTime() ? $('#J_durationTime').text() : '';
-	return Util.stringToSeconds(text);
+	let text = $('.audio-progress .range .bar .handle').val();
+	let str = text.split('/');
+	return str[0];
```
inverse: Is this correct? Isn't `0` the currentTime not the duration? Could we cover currentTime as well as I saw it was removed.

src/connectors/xiami.js
```diff
@@ -1,22 +1,19 @@
...
 };
 
-Connector.getUniqueID = () => $('.ui-track-current .c1').data('id');
```
inverse: Is there no way to get uniqId?

src/connectors/xiami.js
KwonDae: When the page's music is playing, currentTimes is viewed like 01:03/04:12.
 So i get currentTime using 
`let str = text.split('/');`
`return str[0];`
If it don't be needed, I'll erase it and commit it again.

src/connectors/xiami.js
KwonDae: I could not find a way to get an uniqId from the website.

src/core/connectors.js
```diff
@@ -293,7 +293,7 @@ define(function() {
 
 	{
 		label: 'Xiami',
-		matches: ['*://www.xiami.com/play*'],
+		matches: ['*://www.xiami.com/*'],
```
naoyeye: need to consider the case no slash at the end? e.g `https://www.xiami.com`

src/connectors/xiami.js
naoyeye: `$('.play-bar .content .title').attr('href').split('?')[0].split('/song/')[1]`

image

image

src/core/connectors.js
inverse: Should match since `*` is wildcard

KwonDae:
> @KwonDae any update?

This codes are working well in my local laptop(MacOs).
So i think that this PR will work well :)

KwonDae:
> @KwonDae any update?

Moreover, It seems that player site was changed 'https://www.xiami.com/play' to 'https://www.xiami.com'. So i modified some codes about that part.

src/connectors/xiami.js
inverse: @KwonDae Can you apply the changes that @naoyeye suggested?

src/connectors/xiami.js
inverse: I still think this is wrong. so text will contain `00:30/02:45` for example and splitting this will yeild `00:30` in position `0`... which is the current time - not the duration. Does that make sense?

Perhaps a better solution would be to just set

```
Connector.timeInfoSelector = '.audio-progress .range .bar .handle';
```

This handles the case when the duration and remaining time is in this format.

KwonDae: > I still think this is wrong. so text will contain `00:30/02:45` for example and splitting this will yeild `00:30` in position `0`... which is the current time - not the duration. Does that make sense?
> 
> Perhaps a better solution would be to just set
> 
> ```
> Connector.timeInfoSelector = '.audio-progress .range .bar .handle';
> ```
> This handles the case when the duration and remaining time is in this format.

Oh.. I understand!
I modified duration part just now.
Is it right?

src/connectors/xiami.js
KwonDae: @naoyeye 
Oh Thanks!!
I'll try it now. But can you explain the getUniqueId why it is needed?

inverse: @KwonDae read the docblock above the following: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/content/connector.js#L251

src/connectors/xiami.js
```diff
@@ -8,11 +8,7 @@ Connector.artistSelector = '.music .info .singers';
 
 Connector.trackArtSelector = '.music .cover-link .active img';
 
-Connector.getDuration = () => {
-	let text = $('.audio-progress .range .bar .handle').val();
-	let str = text.split('/');
-	return str[0];
-};
+Connector.timeInfoSelector = '.audio-progress .range .bar .handle';
```
KwonDae: @inverse 
Is this right your proposed?

inverse: looks good to me!

src/connectors/xiami.js
```diff
@@ -8,6 +8,8 @@ Connector.artistSelector = '.music .info .singers';
 
 Connector.trackArtSelector = '.music .cover-link .active img';
 
+Connector.getUniqueID = () => $('.play-bar .content .title').attr('href').split('?')[0].split('/song/')[1];
```
KwonDae: @inverse  @naoyeye 

I add codes that @naoyeye 's suggested.
Please check and let know me anything that needs changed :)

KwonDae:
Can you explain detail?

xTpx:
The old version of xiami music move to emumo.xiami.com

xTpx:
The old version of xiami music move to emumo.xiami.com, So can we provide a xiami-legacy?

Issue #1874, opened 2019-01-23
TIDAL stopped working again.


daniel-stoneuk:
TIDAL isn't being detected anymore. 

### Environment
* TIDAL Web Player Version:  3.7.0--31 (NP: 2.3.8.1)
* Version used: 2.6.0
* Browser name and version: Chrome Beta 72

daniel-stoneuk:
Fixed in #1875

PR #1875, opened 2019-01-23
Fix TIDAL issues (#1874)

daniel-stoneuk:
Simple change of attribute name stopped the connector from recognising playing state. Fixed for TIDAL Web Player 3.7.0--31 (NP: 2.3.8.1)


jaccarmac:
LGTM

inverse:
@daniel-stoneuk thanks for the fix :+1: and @jaccarmac for confirming! will be in the next release!

Issue #1876, opened 2019-01-24
Please support to musi.sh new Apple Music Web Player option


DiegoMarcondesD:
It would be amazing to get support for this site: https://musi.sh
It's a new alternative to Apple Music 

thanks!

inverse:
Resolved with #1877

PR #1877, opened 2019-01-24
Add support for musi.sh (#1876)

eriksf:
This is another Apple Music player based on MusicKit JS. Simple addition to src/core/connectors.js.


inverse:
That was quick ;)

DiegoMarcondesD:
I can't wait!

PR #1879, opened 2019-01-28
Gives "Now Playing" notifications the silent attribute

### LINKED ISSUE #1670: Notifications are not silent when using native notifications
In the latest builds of Chrome, the notifications system now hooks into the built-in Windows 10 panel.  Song change notifications are now no longer silent, so the notification call is most likely missing the 'silent' attribute in some manner.

### Expected behavior
On song change, the notification about the newly recognised song should not trigger a notification sound.

### Current behavior
On song change, the new song is displayed and a notification sound also occurs.

### Possible solution

The code [here](https://github.com/web-scrobbler/web-scrobbler/blob/48761ea87cf6adc2b6ec7c659e9e127fdd547c04/src/core/background/notifications.js#L105) needs to be changed most likely.

https://developer.chrome.com/apps/notifications#type-NotificationOptions is the place to go for the spec, but I don't see any obvious 'sound' aspect other than the 'priority' field.  Perhaps this could be related somehow?

### Environment
* Version used: 2.5.0
* Browser name and version: Google Chrome Version 69.0.3472.3 (Official Build) dev (64-bit)

Xalaxis:
This pull request should resolve #1670 , at least partially.

No issues found with local testing.

### DIFF
--- a/src/core/background/notifications.js
+++ b/src/core/background/notifications.js
@@ -166,6 +166,7 @@ define((require) => {
 				iconUrl: song.getTrackArt() || chrome.extension.getURL('/icons/default_cover_art.png'),
 				// @ifdef CHROME
 				title: song.getTrack(),
+				silent: true,
 				message: song.getArtist(),
 				contextMessage: connectorLabel
 				// @endif


inverse:
Thnx - seems like a legit fix - never had the problem on linux but can imagine it being annoying.

Issue #1882, opened 2019-01-28
Add other means of identifying a service


RodrigoLeiteF:
Is it possible to add an option for connectors to identify a service through means other than the website's URL? It would make adding support for self-hosted platforms such as https://funkwhale.audio, which are not tied to a specific pattern or URL, but work consistently across all of them.

inverse:
We provide the ability to change the URL for a given connector. See https://github.com/web-scrobbler/web-scrobbler/wiki/Custom-URL-patterns

RodrigoLeiteF:
It works in the short term, but I think ideally it's best to have something like having the extension look for certain meta tags to automatically identify such services.

EliotBerriot:
Hi, Funkwhale maintainer here :)

If needed, I can easily add a meta tag into Funkwhale HTML, like `<meta name="generator" content="funkwhale 0.18">`. I'm not sure how the actual scrobbling work though. How does Web Scrobbler identifies that the player switched to another track? If I can help with that, let me know!

EliotBerriot:
> The scrobbling normally works by binding to DOM elements and watching for changes. e.g. like this: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/mnet.js

Makes sense, I'll try to keep stable class names for artists / albums / track data then!

alexesprit:
I've added a basic support for identifying websites via checking for match rules in a content script. Will merge these changes into master branch after some testing and extending the match rules.

alexesprit:
@EliotBerriot would be great if you add a meta tag like you suggested.  I think it should contain no version:
```xml
<meta name="generator" content="Funkwhale">
```
as the version is redundant here (can be added as a separate `meta` tag, though).

EliotBerriot:
@alexesprit, sure, I'll do that today, and it will be included in our next release :)

EliotBerriot:
@alexesprit So it's been implemented, and deployed at https://open.audio/library if you want to try it. Let me know if you need me to set some static classes / attributes on the player to make the job easier for you :)

stephanebruckert:
Can we find a method to do this with the Spotify and Bandcamp embeds as well? These are still open/unmerged:
 
- https://github.com/web-scrobbler/web-scrobbler/issues/2024
- https://github.com/web-scrobbler/web-scrobbler/pull/2026

alexesprit:
[...]

PR #1884, opened 2019-01-29
Add funkwhale connector

RodrigoLeiteF:
Allows Web Scrobbler to work on [Funkwhale](https://funkwhale.audio/). As stated in #1882, Funkwhale is a self hosted service that can work on many different URLs, so no URL matching was added.

### DIFF
--- a/src/connectors/funkwhale.js
+++ b/src/connectors/funkwhale.js
@@ -1,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.player';
+
+Connector.trackSelector = '.small.header.discrete.link.track';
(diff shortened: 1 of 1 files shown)


RodrigoLeiteF:
At the moment I set up a custom URL pattern on the extension's options to match that of my instance's. I assumed that worked without the dummy listing on the connectors list, but that doesn't appear to be the case so I went ahead and did so.

inverse:
Could you fix the CI issues? and then it should be good :+1:

inverse:
Ahh that was me :+1: all good.

Issue #1852, opened 2019-01-01
Please add support for R-a-d.io


quietmansoath:
It would be great if you could add support for https://r-a-d.io/. Thanks for all of your hard work on this addon!

ExpandingShapes:
Hi @quietmansoath !
I've implemented your requested. The feature will be added in the next release.

quietmansoath:
Perfect!! Thanks!

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

alexesprit:
Implemented in #1852.

PR #1901, opened 2019-02-05
feature/#1852Support_For_R-a-d.io

ExpandingShapes:
Added support for R-a-d.io required in #1852 
I've tested scrobbling to Last.fm in Chrome Version 71.0.3578.98 (Official Build) (64-bit) both unpacked and compiled versions.
Command `>grunt` exited without errors.

### DIFF
--- a/src/connectors/r-a-d.io.js
+++ b/src/connectors/r-a-d.io.js
@@ -5,3 +5,7 @@ Connector.artistTrackSelector = '#np';
 Connector.playButtonSelector = '#stream-play';
 
 Connector.playerSelector = '.dynamic-row > .col-md-6';
+
+Connector.durationSelector = '#progress-length';
+
+Connector.currentTimeSelector = '#progress-current';
--- a/src/connectors/r-a-d.io.js
+++ b/src/connectors/r-a-d.io.js
@@ -1,7 +1,7 @@
-'use strict'
+'use strict';
 
-Connector.artistTrackSelector = '#np'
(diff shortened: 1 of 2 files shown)


src/core/connectors.js
```diff
@@ -1279,5 +1279,9 @@ define(function() {
 		label: 'Funkwhale',
 		matches: [''],
 		js: ['connectors/funkwhale.js'],
+	}, {
+		label: 'R-a-dio',
+		matches: ['*://r-a-d.io/*', '*://www.r-a-d.io/*'],
```
inverse:
```suggestion
		matches: ['*://r-a-d.io/*'],
```

www redirects to apex domain so this isn't needed.

src/connectors/r-a-d.io.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+Connector.artistTrackSelector = '#np';
+
+Connector.playButtonSelector = '#stream-play';
+
+Connector.playerSelector = '.dynamic-row > .col-md-6';
```
inverse: There should also be mapping for the total duration:
```
Connector.durationSelector = '#progress-length';
```
and current time: 
```
Connector.currentTimeSelector = '#progress-current';
```

ExpandingShapes: Thanks. Ah, I see. I've comitted the necessary changes.

inverse:
Thanks! Will be in the next release

PR #1910, opened 2019-02-10
Apply cache size to popup title correctly

alexesprit:
Fixes an issue when cache size was applied multiple times.

Cache size valuse is applied every time the popup window is opened. Here's a representation of issue on the screenshot. The edited tracks popup window was opened triple times.
image__


alexesprit:
Made this PR just to describe the issue and attach the screenshot. Then I remembered I can comment commits. 😃

inverse:
Ahah nice one :+1:

PR #1919, opened 2019-02-26
[yt-rotate] Rotate keys

### LINKED ISSUE #1913: Youtube does not scrobble on Google Chrome, but Open Spotify scrobbles, uBlock Origin is disabled
### Steps to reproduce (for bugs)

1\.
2\.
3\.

### Steps for resolving the issue

-   [x] Reauthenticate on Last.fm
    
    
    
-   [ ] Try to use unpacked extension from latest master
    
    
-   [x] Check other issues for duplicates

### Environment

-   Version used: v2.14.0
-   Browser name and version: Google Chrome Version 72.0.3626.96

inverse:
Double the keys in use. What do you think? 

Resolves #1913

### DIFF
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -27,7 +27,7 @@ const videoSelector = '.html5-main-video';
  */
 const YT_API_KEYS = [
 	'AIzaSyA3VNMxXEIr7Ml3_zUuzA7Ilba80A657KE',
-	'AIzaSyCZOorF9FeQ8q63zYkoqR_j4nVT85w4EYg',
+	'AIzaSyBAYYYOe4CIfiMnef6y6dAPR9YxneCB87w',
 ];
 
 readConnectorOptions();
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -177,14 +177,14 @@ function getVideoCategory(videoId) {
 	if (videoId === null) {
 		return null;
 	}
-	if (categoryCache.has(videoId)) {
+	if (!categoryCache.has(videoId)) {
 		fetchCategoryId(videoId).then((category) => {
 			if (category === null) {
-				throw Error(`Failed to resolve category for ${videoId}`);
+				console.log(`Failed to resolve category for ${videoId}`);
 			}
 			categoryCache.set(videoId, category);
-		}).catch((err) => {
-			console.log(err);
+		}).catch((error) => {
+			console.log(error.message);
 		});
 	}
 
@@ -199,7 +199,7 @@ async function fetchCategoryId(videoId) {
 		try {
 			const response = await fetch(url);
 			if (!response.ok) {
-				throw Error(response.statusText);
+				throw new Error(response.statusText);
 			}
 
 			const data = await response.json();
@@ -208,9 +208,8 @@ async function fetchCategoryId(videoId) {
 			if (typeof category === 'string') {
 				return category;
 			}
-		}
-		catch (err) {
-			console.log(`Failed fetching category for ${videoId} due to ${err}`);
+		} catch (error) {
+			console.log(`Failed fetching category for ${videoId} due to ${error.message}`);
 		}
 	}
 
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -193,8 +193,7 @@ function getVideoCategory(videoId) {
 
 async function fetchCategoryId(videoId) {
 	let found = false;
-	for (let i = 0; i < YT_API_KEYS.length; i++) {
-		let key = YT_API_KEYS[i];
+	for (let key of YT_API_KEYS) {
 		const url = `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${videoId}&key=${key}`;
 		try {
 			const response = await fetch(url);
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -183,8 +183,6 @@ function getVideoCategory(videoId) {
 				console.log(`Failed to resolve category for ${videoId}`);
 			}
 			categoryCache.set(videoId, category);
-		}).catch((error) => {
-			console.log(error.message);
 		});
 	}
 
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -195,11 +195,12 @@ async function fetchCategoryId(videoId) {
 		const url = `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${videoId}&key=${key}`;
 		try {
 			const response = await fetch(url);
+			const data = await response.json();
+
 			if (!response.ok) {
+				console.log(JSON.stringify(data, null, 2));
 				throw new Error(response.statusText);
 			}
-
-			const data = await response.json();
 
 			let category = data.items[0].snippet.categoryId;
 			if (typeof category === 'string') {
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -178,36 +178,37 @@ function getVideoCategory(videoId) {
 		return null;
 	}
 	if (!categoryCache.has(videoId)) {
-		let found = false;
-		for (let i = 0; i < YT_API_KEYS.length; i++) {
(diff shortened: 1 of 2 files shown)


alexesprit:
The idea is good.

But I'd use promise sequences (an example of usage is in `coverartarchive` pipeline stage), or ES `async/await` feature.

With simple `for` loop the connector can use from one up to all API keys, even if API response using 1st key is successful.

inverse:
Good call, also I think we need to make sure that we block for each state mutation.

inverse:
@alexesprit how does this look?

inverse:
Just need to figure out the eslint rules to allow

inverse:
Actually there are still issues. Any pointers?

alexesprit:
You should add this block in ESLint config:
```js
'parserOptions': {
	'ecmaVersion': 8,
}
```

alexesprit:
I can't use the second token due to error:
<details>
  <summary>Click to expand</summary>

```json
{
 "error": {
  "errors": [
   {
    "domain": "usageLimits",
    "reason": "accessNotConfigured",
    "message": "Access Not Configured. YouTube Data API has not been used in project 177713213466 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/youtube.googleapis.com/overview?project=177713213466 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "extendedHelp": "https://console.developers.google.com/apis/api/youtube.googleapis.com/overview?project=177713213466"
   }
  ],
  "code": 403,
  "message": "Access Not Configured. YouTube Data API has not been used in project 177713213466 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/youtube.googleapis.com/overview?project=177713213466 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry."
 }
}
```
</details>

src/connectors/youtube.js
```diff
@@ -174,25 +177,46 @@ function getVideoCategory(videoId) {
...
+async function fetchCategoryId(videoId) {
+	let found = false;
+	for (let i = 0; i < YT_API_KEYS.length; i++) {
+		let key = YT_API_KEYS[i];
+		const url = `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${videoId}&key=${key}`;
+		try {
+			const response = await fetch(url);
 			if (!response.ok) {
-				throw new Error('Invalid response');
+				throw Error(response.statusText);
```
alexesprit: It should be `throw new Error...`.

src/connectors/youtube.js
```diff
@@ -174,25 +177,46 @@ function getVideoCategory(videoId) {
 	if (videoId === null) {
 		return null;
 	}
-	if (!categoryCache.has(videoId)) {
-		const url = `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${videoId}&key=${YT_API_KEY}`;
-		fetch(url).then((response) => {
+	if (categoryCache.has(videoId)) {
+		fetchCategoryId(videoId).then((category) => {
+			if (category === null) {
+				throw Error(`Failed to resolve category for ${videoId}`);
```
alexesprit: Well, in this case the connector will fetch category for video id constantly. I suggest to use this algorithm:
```js
fetchCategoryId(videoId).then((category) => {
	if (category === null) {
		console.log(`Failed to resolve category for ${videoId}`);
	}
	categoryCache.set(videoId, category);
}).catch((err) => {
	console.log(err);
});
```
Add category in cache anyway, even if it's null.

src/connectors/youtube.js
```diff
@@ -174,25 +177,46 @@ function getVideoCategory(videoId) {
...
+			console.log(err);
+		});
+	}
+
+	return categoryCache.get(videoId);
+}
+
+async function fetchCategoryId(videoId) {
+	let found = false;
+	for (let i = 0; i < YT_API_KEYS.length; i++) {
```
alexesprit: Should be replaced with `for...of` loop.

src/connectors/youtube.js
```diff
@@ -174,25 +177,46 @@ function getVideoCategory(videoId) {
...
+				return category;
 			}
-		}).catch((err) => {
-			console.log(`Unable to get category for ${videoId}: ${err.message}`);
-		});
-		return null;
+		}
+		catch (err) {
+			console.log(`Failed fetching category for ${videoId} due to ${err}`);
+		}
```
alexesprit: It's ok, but almost all code is in this format:
```js
if (condition) {
    statement
} else {
    statement
}
```
The same thing for `try...catch` blocks.

src/connectors/youtube.js
```diff
@@ -174,25 +177,46 @@ function getVideoCategory(videoId) {
 	if (videoId === null) {
 		return null;
 	}
-	if (!categoryCache.has(videoId)) {
-		const url = `https://www.googleapis.com/youtube/v3/videos?part=snippet&id=${videoId}&key=${YT_API_KEY}`;
-		fetch(url).then((response) => {
+	if (categoryCache.has(videoId)) {
```
alexesprit: Should be:
```js
if (!categoryCache.has(videoId)) {
```

inverse: n00b mistake

src/connectors/youtube.js
inverse: Ahh formatting ;)

inverse: WOuldn't this be it? https://eslint.org/docs/rules/brace-style 1tbs which we already use?

alexesprit: It's already set up for 1TBS, but does not work for `try...catch` statements.

https://eslint.org/docs/rules/brace-style

src/connectors/youtube.js
```diff
@@ -177,14 +177,14 @@ function getVideoCategory(videoId) {
...
+	if (!categoryCache.has(videoId)) {
 		fetchCategoryId(videoId).then((category) => {
 			if (category === null) {
-				throw Error(`Failed to resolve category for ${videoId}`);
+				console.log(`Failed to resolve category for ${videoId}`);
 			}
 			categoryCache.set(videoId, category);
-		}).catch((err) => {
-			console.log(err);
+		}).catch((error) => {
```
alexesprit: I noticed this `catch` is unused, because `fetchCategoryId` never throws any error.

inverse:
Not sure what's up with the token, I set everything up through the panel but it's causing that issue you listed.

src/connectors/youtube.js
inverse: Should we tackle this seperately?

alexesprit:
Also, maybe it's a good idea to add error response to debug output:
```js
const response = await fetch(url);
const data = await response.json();

if (!response.ok) {
	console.log(JSON.stringify(data, null, 2));
	throw new Error(response.statusText);
}
```

inverse:
Will that work with your example? can you call await response.json() on a non-ok request?

alexesprit:
Yes, I tested it, and it works as intented. AFAIK, `response.ok` means 2xx response code, and vice versa.

alexesprit:
@inverse also what do you think about storing YT API keys in one place? Now one key is in my project, and one is yours. There's an option to add members to projects, so all members can view stat of all API keys and manage them.

Unfortunately, it's not possible to make several API keys in one project, so I made 2 projects, which project contains one API key. Both keys are valid; first one exceeded its quota, though, but the second key works fine. If you're in, please tell me your Google account, and I will add you to the project team (you can email me or send it via Gitter, rather that posting email here, if you want).

src/connectors/youtube.js
```diff
@@ -175,24 +178,42 @@ function getVideoCategory(videoId) {
...
-		return null;
+		} catch (error) {
+			console.log(`Failed fetching category with ${key}`);
+		}
 	}
-	return categoryCache.get(videoId);
 
+	if (!found) {
+		return null;
+	}
```
alexesprit: We can just return null value at the end of function, the `found` variable isn't used in the function.

PR #1936, opened 2019-03-10
[multi-lang-privacy] Allow loading multi lang privacy

inverse:


src/startup/startup.js
```diff
@@ -26,15 +26,36 @@ require([
...
+	const locale = window.navigator.language.split('-')[0];
+	const defaultPrivacyDoc = 'PRIVACY.md';
 
-	fetch(privacyUrl)
+	let privacyDoc = defaultPrivacyDoc;
+	if (locale !== 'en') {
+		privacyDoc = `PRIVACY.${locale}.md`;
+	}
+
+	fetch(chrome.runtime.getURL(privacyDoc))
```
inverse: Is it worth splitting this out into some md loader class?

alexesprit: No, but I'd use for..of loop and async/await instead to simplify the logic and code. Something like this:
```js
let privacyDoc = `PRIVACY.__MSG_@@ui_locale__.md`;

let langFiles = [defaultPrivacyDoc];
if (locale !== 'en') {
	// Insert translated MD file, so we can try to use it first
	languages.unshift(privacyDoc);
}

for (let f of langFiles) {
    // Fetch language; if everything is ok, break the loop
}

File: src/startup/startup.js [hunk 2]
```diff
@@ -26,15 +26,36 @@ require([
 		$('.finished').show();
 	};
 
-	let privacyUrl = chrome.runtime.getURL('PRIVACY.md');
+	const locale = 'riu'; // window.navigator.language.split('-')[0];
```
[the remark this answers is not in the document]
inverse: Whats wrong with using `window.navigator.language.split('-')[0]`?

src/startup/startup.js
inverse: Updated

inverse:
@alexesprit updated :+1:

Issue #1925, opened 2019-03-01
Free Music Archive (FMA) artist recognition broken on album pages


douglasawh:
### Expected behavior

recognize the same artists recognized on bandcamp.com

### Current behavior

Extension asks me to provide artist name. It picks up the track and album correctly

### Steps to reproduce (for bugs)

Go to http://freemusicarchive.org/music/Lorenzos_Music/Rom-Comm_Mixtape/ and start playing. The artist pages, such as http://freemusicarchive.org/music/Lorenzos_Music/ do work.

### Environment

-   Version used: 2.14.0
-   Browser name and version: Chrome Version 72.0.3626.109 (Official Build) (64-bit)

PR #1941, opened 2019-03-21
[#1925] fix FMA not scrobbling authors on album page

romab1998:
Fix Free Music Archive connector for album page (issue #1925)
It seems like page structure change and selector was no longer valid.


alexesprit:
Nice, thanks!

PR #1942, opened 2019-03-21
[travis-test] Add run-tests, tidy up

inverse:
Run browser tests against changed connectors only


scripts/run-tests.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+TEST_PATH="$(pwd)/tests/connectors/"
+MATCHES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | grep 'src/connectors/' | sed 's/src\/connectors\///' | rev | cut -c 4- | rev)
```
inverse: Could rework this to handle both local and travis env

inverse:
@alexesprit what's your opinion on this?

alexesprit:
@inverse LGFM, but I think some test might be outdated a little. Also, last time I tried to implement browser testing via Travis CI, I was not able to get all tests worked correctly.

Issue #1937, opened 2019-03-11
Firefox 60.5.1esr can't connect to listenbrainz


moinmoi:
### Expected behavior

Getting signed in to ListenBrainz and be able to scrobble to it.

### Current behavior

I get redirected to an URL much like this one (id and state redacted): https://musicbrainz.org/oauth2/authorize?response_type=code&redirect_uri=https%3A%2F%2Flistenbrainz.org%2Flogin%2Fmusicbrainz%2Fpost&scope=profile&state=state&client_id=id

I login and end up on https://listenbrainz.org/profile/ where I can view my user token et.c, but it doesn't seem to link the account to the add on.

Last.fm login and scrobbling works fine.

### Steps to reproduce (for bugs)

1\. Preferences -> Accounts -> ListenBrainz -> click "Sign In" 
2\. Login on the website you get redirected to.

-   [X] Try to use unpacked extension from latest master
-   [X] Check other issues for duplicates

### Environment

-   Version used: 2.15.1 (2.15.0 when unpacked)
-   Browser name and version: Firefox 60.5.1esr

inverse:
@moinmoi but when you start playing music does it show up on listenbrainz? e.g. https://listenbrainz.org/user/<your-username>

I can confirm your behaviour described but it seems working to me.

alexesprit:
@inverse is it working for you on 60.5.1esr (or 60.6.1esr)? I can reproduce the issue on 0.6.1esr, but have no idea how to resolve it.

moinmoi:
@inverse No, I'm afraid nothing shows up on listenbrainz, even though things do get scrobbled to last.fm.

Currently on 60.6.0esr.

inverse:
That's strange. just verified again from fresh install.

I'm running FF 66.0 (64-bit)

moinmoi:
Yea, works fine on non-esr firefox. And since I ran esr for no logical reason, I happily changed to FF 66.0.1 and don't mind at all if this issue is closed now.

Thanks for the help :)

inverse:
I didn't even know ESR was a thing. Def worth having this ticket open to see if others are able to help. thnx for raising it!

alexesprit:
Closed as fixed.

PR #1946, opened 2019-03-30
[#1937] fix login with listenbrainz on firefox 60.x esr

romab1998:
It seems like firefox had different default fetch settings for plugins. Therefore cookies for listenbrainz simply were not provided to request. And it seems like somewhere between versions 60.x and 66.x default behavior was changed, that's why you can successfully connect with listenbrainz now.
This change explicitly sets credentials policy of fetch to 'same-origin' allowing to pass cookies to fetch and connect account.


alexesprit:
Thanks for the bugfix!

PR #1947, opened 2019-03-31
Add album selector for Hype Machine premieres.

jaccarmac:
For some reason I didn't implement this when writing the connector, but it's easy enough to add and in the same vein as the artist selector.


inverse:
thnx

PR #1952, opened 2019-04-08
Add duration selectors for Hype Machine (Premieres).

jaccarmac:
Previously this wasn't implemented so scrobbling would always happen after 30s. This is very annoying when trying to make edits to track info.


jaccarmac:
Not sure why the Selenium tests are timing out, I don't have a working install locally right now. I'll see if I can fiddle with Chrome binaries and see what's up. I suspect the support modal that's been around as of late.

jaccarmac:
Both tests pass locally so it looks like a Travis hiccup.

alexesprit:
Thanks!

alexesprit:
> Not sure why the Selenium tests are timing out, I don't have a working install locally right now.

Probably, tests were failed, because Chrome is running in headless mode.

PR #1965, opened 2019-04-25
[bump] bump dependencies

inverse:


alexesprit:
Should we update package versions in the `package.json` file?

inverse:
We can do :+1: and should do if it's compatible :+1:

inverse:
Should be a case of running `yarn upgrade --latest`

Issue #1705, opened 2018-09-07
PLEX scrobbler bug when expanding music player


Botanistt:
Scrobbling with plex works just fine normally, but sometimes I want to expand the music player, because it would be more comfortable for me, but once I do it starts scrobbling songs incorrectly. I have attached a screenshot above where you can see the problem in action.
https://i.imgur.com/MZxZwJP.jpg
In some cases the song name would repeat multiple times, but in the screenshot you can see that it is being repeated only twice though it can be repeated even 5 times. It does scrobble it too, which obviously is not correct.

I believe this could be fixed.

The only way how to replicate this bug is if you listen on your own and enable the extended music player view. That's all.

Not sure how these are relevant to the issue, but sure.
Browser: Opera version 55.0.2994.44
Scrobbler version: 2.6.0

inverse:
Thanks for reporting. Hopefully someone also who uses plex can reproduce and fix :+1:

Botanistt:
anyone using PLEX? 🙄

jaccarmac:
Is this still a problem? I see the connector was updated November 3, but if it still is I have an idea how to fix this bug given something similar I encountered in the TIDAL connector.

jaccarmac:
Also might be worth pinging the author of the commit I'm talking about: b9e063a39530684ff2ef56aa81a35fd021875994

alexesprit:
> I have an idea how to fix this bug given something similar I encountered in the TIDAL connector.

@jaccarmac I can reproduce this issue. What's your idea?

jaccarmac:
My first attempt at the TIDAL connector was d817566fed098c763c84bac55c5b8a859fb3974c, but when expanding the player *there*, another `infoTable` would be created and I would get garbage data. I fixed this in b194990379c97a0bee7e87cedf2b4046526ad8f9 with `nth-child(1)`, though the fix could also be implemented in the `Connector` base class. I don't think there is any case where we would want to pull data from anything other than the first result of a given selector?

alexesprit:
I updated `Connector.playerSelector` property and fixed this issue. Now I'm trying to fix play/pause toggle.

> though the fix could also be implemented in the Connector base class. I don't think there is any case where we would want to pull data from anything other than the first result of a given selector?

Do you mean we should replace base implementation with the following:
```js
/**
 * Default implementation of track name lookup by selector.
 *
 * Override this method for more complex behaviour.
 *
 * @return {String} Song title
 */
[...]

PR #1971, opened 2019-05-08
Add option to use Tidal-provided shorter track names.

jaccarmac:
I noticed that Tidal recently started updating their info table, especially for jazz records. There is now a table row for the track name stripped of any extra bits of information. For example, the track [Out Of The Night (Rudy Van Gelder Edition / 1999 Digital Remaster / 24-Bit Mastering)](https://listen.tidal.com/album/1381684/track/1381690) shows up in the info table as simply "Out Of The Night". We can use this to make for cleaner scrobbles. As a bonus that info table appears to be populated now even if the fullscreen now playing drawer isn't open.

I can put this behind an option if people want to be able to scrobble the full-information track names, but for now this is perfect for my own use. Any other Tidal user opinions?

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -52,6 +52,14 @@
     "message": "Scrobble podcasts on Google Music",
     "description": "Option label"
   },
+  "optionTdlShortTrackNames": {
+    "message": "Use short track names",
+    "description": "Option title"
+  },
+  "optionTdlShortTrackNamesTitle": {
+    "message": "Tidal provides shortened track names for certain tracks (without \"Remastered\", etc.)",
+    "Description": "Option label"
+  },
   "optionYtDesc": {
     "message": "If you turn off both options, the extension will scrobble Youtube videos from all categories.",
     "description": "Description of Youtube options"
--- a/src/connectors/tidal.js
+++ b/src/connectors/tidal.js
@@ -1,5 +1,9 @@
 'use strict';
 
+let useShortTrackNames = false;
+
+readConnectorOptions();
+
 Connector.playerSelector = '[class^="nowPlaying"]';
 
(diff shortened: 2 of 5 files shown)


jaccarmac:
@daniel-stoneuk, thoughts?

jaccarmac:
This causes some weird behavior with albums like last year's Both Directions At Once, so I think I'd better add an option before it's truly ready to merge.

alexesprit:
@jaccarmac what's blocking this issue to be merged? If you think we should use an option, feel free to add it.

I remember you [suggested][suggestion] to get a text of first non-empty selector. It was implemented in v2.21.0.

[suggestion]: https://github.com/web-scrobbler/web-scrobbler/issues/1705#issuecomment-491567805

jaccarmac:
Just that. I've been too lazy to implement it since the existing code is working for my case. Planning on sitting down and updating the branch fully this week or next though.

Noticed the Travis build failing recently, but think that might be a problem with Travis and not the PR. Is that the case?

inverse:
@jaccarmac traffic is fine - Error is:


```
/home/travis/build/web-scrobbler/web-scrobbler/src/connectors/tidal.js
  11:7  error  'fullTrackSelector' is assigned a value but never used  no-unused-vars
```

jaccarmac:
Huh, for some reason didn't notice that when I checked logs. Will fix, looks like the API got ahead of this PR.

jaccarmac:
All right, this should be ready for prime-time now! #2041 strips some of the same things but in a different ways. I believe the two PRs can happily coexist.

jaccarmac:
They changed this again, and since I liked the short names I am tempted to resurrect the option. Any potential footguns in adding it again?

jaccarmac:
#3065 has the resurrected behavior without the option.

PR #1976, opened 2019-05-15
Remove grunt-rename-util from yarn.lock.

jaccarmac:
No idea if this is a peculiarity related to my particular version of Yarn, but it's modifying this file every time I go to install. Worth including in the tree if anyone else is seeing the same thing.


alexesprit:
Thanks!

I removed `grunt-rename-util` in 155b26cafce9fd68bb479fc62b28f38f9d1c982e, but forgot to remove the package from `yarn.lock`.

PR #1993, opened 2019-06-06
[fix-stroage] Fix storage

inverse:


alexesprit:
I've already fixed that, but haven't pushed.

Core and LocalCache storages should be local. I'll update this by myself.

Btw, why do you prefix the commit message with [...]?

inverse:
I've got a hook to prefix commit message with the branch name. I use it at work for the ticket number for tracking with JIRA.

inverse:
https://gist.github.com/bartoszmajsak/1396344

alexesprit:
I think it's redundant in this project, as almost all branches are insta-merged and contain a single commit.

inverse:
I've set it up as a global hook and requires zero effort from my side ;)

inverse:
Perhaps I should tidy it up and match only on the branch naming e.g `\w+-\d+`

inverse:
```
#!/bin/bash

# Prepend branch name to commit message in []

BRANCH_NAME=$(git symbolic-ref --short HEAD)
BRANCH_NAME="${BRANCH_NAME##*/}"

BRANCH_EXCLUDED=$(printf "%s\n" "${BRANCHES_TO_SKIP[@]}" | grep -c "^$BRANCH_NAME$")
BRANCH_IN_COMMIT=$(grep -c "\[$BRANCH_NAME\]" $1)

if [ -n "$BRANCH_NAME" ] && [[ $BRANCH_NAME =~ ^[A-Z]+-[0-9]+$ ]] && ! [[ $BRANCH_IN_COMMIT -ge 1 ]]; then
    sed -i.bak -e "1s/^/[$BRANCH_NAME] /" $1
fi
```

done :)

PR #1994, opened 2019-06-06
[resonate-fix] Fix for resonate

inverse:


src/connectors/resonate.js
```diff
@@ -11,5 +11,5 @@ Connector.currentTimeSelector = `${Connector.playerSelector} .currentTime`;
 Connector.durationSelector = `${Connector.playerSelector} .duration`;
 
 Connector.isPlaying = () => {
-	return $('.play-button use').attr('xlink:href') === '#icon-pause';
+	return !$('.play-button svg').hasClass('icon-play');
```
alexesprit: I'd replace the function with:
```js
Connector.playButtonSelector = '.play-button svg.icon-play';
```

inverse: Updated

PR #2006, opened 2019-07-01
Rework bandcamp

inverse:


alexesprit:
I started to refactor Bandcamp based on updated Connector API to support multiple players on the page.

inverse:
@alexesprit Want to take over? I got the selectors for the collections page.

alexesprit:
> Want to take over?

I was just interesting we could reimplement the connector using array selectors. If you want to continue and have some time, feel free to do that. And, of course, you can use the code from my branch.

> Do you think your one provides enough flexibility for custom mapping on the various pages. e.g. ones that are not just selector based.

What do you mean by "custom mapping on the various pages"?

alexesprit:
@inverse I updated my branch, fixed few regressions and improved support for "Collections" player (the `#collection-player` element is missing on [discovery][discovery] page, but the players are the same).

Now the rewritten connector should work as before, plus, it should support "Collections" player. At least, my main goal is to keep the connector to work w/o regressions, e.g. missing scrobbles from pages supported before refactoring.

What do you think about these changes?

[discovery]: https://bandcamp.com/tag/discovery

alexesprit:
I'm not sure if this check is really useful:
```js
function isArtistLetious(artist, track) {
	...

	/*
	 * Discover & song pages: true if artist name is 'letious' or
	 * 'letious Artists' and track contains a hyphen or vertical bar.
	 * Also takes into account misspelling of letious as letios
	 * Example: http://krefeld8ung.bandcamp.com/album/krefeld-8ung-vol-1
	 */
	return /^letiou?s(\sArtists)?$/.test(artist) &&
		Util.findSeparator(track, separators) !== null;
}
```

I support letious = various, so, in this case we should check for Various Artists too.

But I doubt the Various (or Letious, whatever) artists are on **song** (single) page. Plus, the example is not correct. I see "Varios Artists" on that page, and this is "album" page, so this code won't be executed.

inverse:
Nice improvements. Do you want to close this branch and continue to hash out yours?

alexesprit:
Yep, if you don't mind.

inverse:
done ;)

alexesprit:
> Your's is based around targeting players not pages right? Do you find that approach easier to read from a code stand point?

You're right. I wrote some time ago Bandcamp can display multiple players on a single page, so the way we use to support multiple players won't work; this way assumes only one player is displayed on the page.

The change I suggested allow to apply selectors for each player available on the page. I added some objects (I call them "player objects") contain selectors for each player supported by the connector. You can think of these objects as mini-connectors work at the same time. This is done using updated Connector API. Plus, I updated the connector to set higher priority to `Connector.getArtistTrack` function.

I found only AlbumPlayer and InlinePlayer (really don't know how to name that one) are displayed simultaneously. Other players aren't. So, this approach to use player objects is overkill on the one side, but on the other it resolves the issue with album and inline players, plus quite easy to read, understand, and maintain the connector. It was really hard (at least for me) to say where all selectors are used.

I think I'll make PR after #2010 is merged. Adding albumArtist feature for rewritten connector will require additional changes in the core (I have an idea how to do this, but no real implementation yet).

Issue #1662, opened 2018-06-27
AlbumArtist support


haeckle:
### Expected behavior
Scrobbling an album from e.g. bandcamp with different artists (like [this](https://feinestier.bandcamp.com/album/feines-tier-zoo-vol-1)) should lead to one album in last.fm (like [this](https://www.last.fm/de/music/Various+Artists/Feines+Tier+Zoo,+Vol.+1))

### Current behavior
Every artist of the album get an own album with the albumtitle but only their own songs in it (like [this](https://www.last.fm/de/music/The+Cheapers/Feines+Tier+Zoo+Vol.1))

### Possible solution
Also submit the albumartist

### Steps to reproduce (for bugs)
Scrobble the above mentioned album (or any other with various artists) with we-scrobbler.

Maddieman:
I'd like to see this too.

inverse:
@haeckle @Maddieman - just to clarify in the example URL listed.

artist would be - "Various Artists"
albumArtist would be - the actual artist

?

haeckle:
albumArtist should be "Various Artists" and artist the actual artist oft the song, in this case "The Cheapers"

Thanks for your effort!

alexesprit:
Closed via #2010.

PR #2010, opened 2019-07-08
Add support for albumArtist

inverse:
Resolution of #1662 

## Different artist/album artist
image

image

## Same  artist/album artist

image

image


src/core/content/connector.js
```diff
@@ -163,6 +174,15 @@ function BaseConnector() {
...
 	this.getAlbum = () => $(this.albumSelector).text();
 
+	/**
+	 * Default implementation of album artist name lookup by selector.
+	 *
+	 * Override this method for more complex behaviour.
+	 *
+	 * @return {String} Song album artist
+	 */
+	this.getAlbumArtist = () => $(this.albumArtistSelector).text();
```
alexesprit: I changed getters recently (108ae84ea74c821e2f70091aeec7facd4a62df42). Now they return `Util.getTextFromSelectors(selector)` instead.

inverse: Rebased from master

src/connectors/bandcamp.js
```diff
@@ -25,7 +25,7 @@ function getArtist() {
 	if (artist === null) {
 		artist = $('.waypoint-artist-title').text().substr(3);
 	}
-	return artist;
+	return $.trim(artist);
```
alexesprit: Why did you add trim? It's useless because trim is done in the default metadata filter.

src/connectors/bandcamp.js
```diff
@@ -109,4 +109,6 @@ Connector.getUniqueID = () => {
 	return null;
 };
 
+Connector.getAlbumArtist = () => $('span[itemprop="byArtist"]').text().trim() || null;
```
alexesprit: The same thing about trim, but also useless fallback to null.

src/connectors/bandcamp.js
inverse: I observed different behaviour

inverse:
```
Web Scrobbler: {
  "track": "Proper Preaching",
  "artist": "Luca Musto",
  "album": "Proper Preaching EP",
  "albumArtist": " \n          Luca Musto\n          ",
  "uniqueID": null,
  "duration": null,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://f4.bcbits.com/img/a1481550943_16.jpg"
}
```

src/connectors/bandcamp.js
inverse: Without trim I see

```
Web Scrobbler: {
  "track": "Proper Preaching",
  "artist": "Luca Musto",
  "album": "Proper Preaching EP",
  "albumArtist": " \n          Luca Musto\n          ",
  "uniqueID": null,
  "duration": null,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://f4.bcbits.com/img/a1481550943_16.jpg"
}
```

inverse: Should getTextFromSelectors do trim?

alexesprit: No, it should not.

albumArtist is not trimmed, because default filter is not modified to trim albumArtist.

src/connectors/bandcamp.js
alexesprit: What do you mean? If you see **artist** field is not trimmed, it should be fixed in another place.

src/connectors/bandcamp.js
inverse: Ahh I'll update the filter

inverse:
Ill finish this later.

src/connectors/bandcamp.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 const filter = new MetadataFilter({
-	all: MetadataFilter.removeZeroWidth
+	all: MetadataFilter.removeZeroWidth,
+	all: MetadataFilter.trim,
```
alexesprit: I meant the default filter defined in `filter.js`. You still have two trim calls.

alexesprit:
I suggest to update `info.css` to disable wrapping of track field:
```css
/* Added .track class */
.artist, .album, .track {
	overflow: hidden;
	text-overflow: ellipsis;
	white-space: nowrap;
}
```

src/core/content/connector.js
```diff
@@ -334,6 +355,10 @@ function BaseConnector() {
 			newState.track = artistTrack.track;
 		}
 
+		if (newState.artist === newState.albumArtist) {
```
inverse: @alexesprit this is before the filter is applied? where would be a good place to hook up this sort of validation or do you think this is alright to push both fields when they are the same?

alexesprit: Yes, in this block you compare raw data (i.e. from the page).

I think there could be cases when raw artist and raw albumArtist are different, mainly because of leading/trailing spaces. Probably, it's much better to compape right before sending the **filtered** connector state to the controller.

It could be fixed by trimming directly in the getters, but I think it's a dirty hack, and I would not recommend to go this way if there's another solution.

> or do you think this is alright to push both fields when they are the same

I don't know how Last.fm responses if we post both artist and albumArtist if they're the same. We can just test it and see how things are going. If it's okay, then this check is redundant.

inverse: Dirty hacks I'm not a fan of. Added a pre-process function to act on the filtered state.

alexesprit: Yes, that's better, althougth I'd move that block to `Connector.filterState` function instead of making the new one. I consider removing delete.albumArtist is a part of filtering.

src/popups/info.css
```diff
@@ -115,3 +115,9 @@ input {
 	cursor: default !important;
 	text-decoration: none !important;
 }
+
+.artist, .album, .track {
+	overflow: hidden;
```
alexesprit: The similar block is already defined on line 73:
```css
.artist, .album {
	overflow: hidden;
	text-overflow: ellipsis;
	white-space: nowrap;
}
```

I suggested to update that one to add `.track` class.

alexesprit:
**Note**: This remark it applied to Bandcamp update in the PR.

As I understand #1662, the proper behavior is when:
- artist = "Various Artists";
- albumArtist = actual track artist.

But Bandcamp implementation in this PR returns "Various Artists" as an albumArtist prop, and actual track artist as an artist prop. Should it be reversed?

inverse:
Good catch :+1:

alexesprit:
Btw, I left some thoughts about various artists (which are related to albumArtist feature) on Bandcamp in #2006.

src/connectors/bandcamp.js
alexesprit: I noticed the `isArtistLetious` function is broken because of raw (i.e. non-trimmed) artist passed as a function argument. If it's a reason why you added trim call, I suggest to fix `isArtistLetious` function instead.

Not sure, if we should do it in this PR, but maybe in #2006 or another.

inverse:
Flipped the logic around to meet desired behaviour - thnx for calling that out :+1:

inverse:
Any other feedback @alexesprit or is it good to go? I made sure that it now goes to correct URL.

alexesprit:
@inverse I  think it's ok now, but I haven't tested this feature yet. I can do, but not today. :(

The only thing I don't know how ListenBrainz should use albumArtist info. I read ListenBrainz API docs and found no albumArtist property.

> As I understand #1662, the proper behavior is when:
> - artist = "Various Artists";
> - albumArtist = actual track artist.

Should we update ListenBrainz to use albumArtist (the actual artist of the track) as artist, if albumArtist is not empty? If no, it will scrobble "Various Artists" as an artist title for VA tracks.

inverse:
Good point - I didn't consider such cases.  

Perhaps it should be done such that if an integration supports albumArtist than use it the way it is. Otherwise keep existing behaviour?

alexesprit:
> Perhaps it should be done such that if an integration supports albumArtist than use it the way it is. Otherwise keep existing behaviour?

I  think we should keep the existing behavior for ListenBrainz, i.e. use albumArtist first, then artist.

inverse:
@paramsingh - Pinging you since you originally chimed in on a PR around ListenBrainz before. 

This PR is basically adding support for albumArtist which is supported by last.fm but was wondering did Listenbrainz support something similar? 

The problem is that if we merge this - the way albumArtist works is that the artist would be in this field and in the artist field would be something like Various Artists

alexesprit:
@inverse I asked some guy how to scrobble albumArtist, and he said Various Artists should be albumArtist. I also found some other people told the same thing.

inverse:
@alexesprit So the flip I did was incorrect?

alexesprit:
@inverse seems like yes. Let's wait for the comments in #1662.

inverse:
@alexesprit it clearly is the original way I had it.

> Album Artist tags should be applied for all compilations that have different track artists. Otherwise, some music players will list each track artist and album name as a separate album. For these tracks, set the Album Artist to something like 'Various Artists' (or whatever is appropriate in your language) and apply this title consistently.

https://www.blisshq.com/music-library-management-blog/2010/10/12/how-to-use-album_artist/

alexesprit:
@inverse now everything is fine for me. What do you think?

inverse:
Seems good to me from testing :+1:

alexesprit:
@inverse btw, did you find any VA compilations on other Bandcamp players, except album and song players?

paramsingh:
Hullo, in LB, we use the `release_aritst_name` (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/spotify_updater/spotify_read_listens.py#L99) field in `additional_info` to submit album artists to ListenBrainz. So the artist should, imo, be the actual artist of the track  for example `Kanye West, Bon Iver` while the `release_artist_name` field contains Kanye West.

inverse:
@paramsingh thanks! I'll update the LB implementation to include that :+1:

inverse:
@alexesprit any other feedback or are we good to go?

Refactors LB to include the `origin_url` on both nowplaying/scrobble and added the `release_artist_name`

alexesprit:
@inverse LGTM except missing semicolons in last commit. Feel free to merge this PR if it's complete.

PR #2011, opened 2019-07-09
Update connector after FIP website update

romainbriand:
FIP connector update after website revamp that was published yesterday.


alexesprit:
Thanks!

Issue #1815, opened 2018-11-28
Add support for Primephonic web player


RyanDBooker:
Add support for the Primephonic web player: https://play.primephonic.com/

KwonDae:
Hi RyanDBooker.
I added the site and tested it, but there were some problems.

1. I don't know if chrome web browser and safari browse are not supported, but it won't play.
like, 
![image](https://user-images.githubusercontent.com/33443660/49189987-d4d19280-f3b3-11e8-9d75-5e2a6d17c512.png)

2. 
I try to add the site and test it, but it doesn't play like  1, so I can't test it.
![image](https://user-images.githubusercontent.com/33443660/49190015-e9158f80-f3b3-11e8-9298-3bc9cd0a5365.png)

My Chrome version is 70.0.3538.110.
If there is any browser that Primephonic web player supports, I would appreciate it if you let me know.

RyanDBooker:
Hello KwonDae,

That's very strange. I am successfully able to play using the Primephonic web player in Chrome version 70.0.3538.110 and also in Safari version 12.0.1 on MacOS.

KwonDae:
Me too, Chrome version 70.0.3538.110 and also in Safari version 12.0.1 on MacOS.
But, the problem still exists.
I can't find why Unable to play audio stream problem.

NamJinWoo:
Hi Ryan, this is Jinwoo. I've looked at it, and the same problem happens to me as it happened to @KwonDae.
I am not sure why this happens.
![image](https://user-images.githubusercontent.com/36990340/49263446-ec7d4980-f48d-11e8-8049-134e5bcf3263.png)
My chrome version is  70.0.3538.110 on MacOS.

RyanDBooker:
Unfortunately, I haven't been able to recreate this; Primephonic plays in every browser I've tried. If I disable Javascript or cookies, I'm not even able to get to the player page. Perhaps the problem is related to an ad or other content blocker?

KwonDae:
I think it is a phenomenon caused by blocking by ip.

Maddieman:
It looks like Primephonic is only available in a limited number of countries (probably due to licensing reasons) - namely: Netherlands, United Kingdom, and United States. So you'd probably need a vpn to test outside of those countries.

https://primephonichelp.zendesk.com/hc/en-us/articles/360007144711-In-which-countries-can-I-access-Primephonic-

[...]

PR #2014, opened 2019-07-15
Added Primephonic connector

wjzijderveld:
I've added a connector for Primephonic.  Supporting the basic functionality artist, album and track.
I've also implemented cover art, and progress. 
`grunt lint` had no complaints.

I understand from reading #1815 that you might not have a way of testing this. Currently I've only tested in Firefox on Linux myself. Let me know if there is anything I can do to help there.

### DIFF
--- a/src/connectors/primephonic.js
+++ b/src/connectors/primephonic.js
@@ -1,0 +1,25 @@
+'use strict';
+
+Connector.playerSelector = '.player';
+
+Connector.trackArtSelector = '#currently-playing-album-cover img';
+
+Connector.artistSelector = '#currently-playing-composer';
+
+Connector.trackSelector = '.currently-playing .movement';
+
+Connector.albumSelector = '#currently-playing-work';
+
+Connector.currentTimeSelector = '#time_container_current';
+
+Connector.getRemainingTime = () => {
+	let time = Util.getTextFromSelectors('#toggle-remainging span');
+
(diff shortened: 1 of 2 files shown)


src/connectors/primephonic.js
```diff
@@ -0,0 +1,25 @@
...
+
+Connector.artistSelector = '#currently-playing-composer';
+
+Connector.trackSelector = '.currently-playing .movement';
+
+Connector.albumSelector = '#currently-playing-work';
+
+Connector.currentTimeSelector = '#time_container_current';
+
+Connector.getRemainingTime = () => {
```
alexesprit: `Connector.getRemainingTime` is used to calculate current time based on duration. You have no duration getter or selector, so this function is useless in this context.

https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/content/connector.js#L358

wjzijderveld: Ah thanks, I'll look at that again. 
By default Primephonic doesn't show the total time, only time remaining, so I might just remove it for the first version.

alexesprit:
> I understand from reading #1815 that you might not have a way of testing this. Currently I've only tested in Firefox on Linux myself. Let me know if there is anything I can do to help there.

I have access to Primeronic (trial, though), and I'm able to test the connector. Everything works perfect. Aside my comment, the changes look good to me.

Issue #2024, opened 2019-07-27
Add support for any Spotify embeds


stephanebruckert:
The idea would be to detect songs played from Spotify embeds on any website, i.e. if you play from:
- https://developer.spotify.com/documentation/widgets/
- https://www.musicglue.com/embeds/spotify
- https://resident-archive.github.io/
- many more

Currently even if your browser is signed in with a Spotify account that's linked to Last.fm, it won't scrobble, but web-scrobbler could remediate to that!

I'm very happy to build the connector, but I wanted to confirm that you would be OK with a connector that accepts any website, since no connector would do that before:

    {
        label: 'Spotify embeds',
        matches: [
           '*://*'
        ],
        js: ['connectors/spotify-embed.js']
    }

Thanks

buremba:
If we already integrated Last.fm with Spotify, will these scrobbles be duplicated or does Last.fm handle these duplicates as well?

alexesprit:
> will these scrobbles be duplicated or does Last.fm handle these duplicates as well?

Users can turn off any connector, so it's not an issue.

alexesprit:
> I'm very happy to build the connector, but I wanted to confirm that you would be OK with a connector that accepts any website, since no connector would do that before:

As I said before, the current implementation allows to inject a single connector only. There're two options:
1. Combine all embed connectors into a single entry. That's really weird.
2. Rework the core.

But these two ways are still unacceptable for me, because the extension will display it supports every website.

stephanebruckert:
> If we already integrated Last.fm with Spotify, will these scrobbles be duplicated or does Last.fm handle these duplicates as well?

@buremba I've tried it and Spotify doesn't scrobble embeds.

PR #2026, opened 2019-07-27
Update bandcamp-embed connector to match all domains

stephanebruckert:
This change is to not only scrobble Bandcamp embeds from:
- https://daily.bandcamp.com/2019/07/26/bandcamp-essential-releases-july-26/#more-106210

but also from any other websites:
- https://en.support.wordpress.com/audio/bandcamp/
- http://www.simonlittlebass.com/2012/01/31/bandcamp-tips-embedding-bandcamp-players-on-your-site/
- etc.

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -42,8 +42,8 @@ define(function() {
 		],
 		js: ['connectors/pitchfork.js'],
 	}, {
-		label: 'Bandcamp Daily',
-		matches: ['*://daily.bandcamp.com/*'],
+		label: 'Bandcamp embeds',
+		matches: ['*://*/*'],
 		js: ['connectors/bandcamp-embed.js'],
 		allFrames: true,
 	}, {


alexesprit:
The only thing I don't like is the extension will display it supports scrobbling for EVERY website. I have some idea (but not implementation) to resolve this issue, but haven't published it yet.

inverse:
What's the overhead cost in having these global matching? e.g. injecting on each page and trying to match?

stephanebruckert:
@alexesprit good point I haven't thought of that. I would be curious to know what's on your mind as I can imagine different possibilities.

Another thing, do you know whether multiple connectors sharing the same pattern (i.e. `*://*/*` for bandcamp-embed and spotify-embed) is currently supported?

alexesprit:
> I would be curious to know what's on your mind as I can imagine different possibilities.

The idea is to add a new content script injected in pages are not supported by the extension. If this script finds some player elements on the page, it asks the core to inject the respective connector via `chrome.runtime.sendMessage`.

> Another thing, do you know whether multiple connectors sharing the same pattern (i.e. *://*/* for bandcamp-embed and spotify-embed) is currently supported?

The current implementation allows to use only one connector per page. I think there's no sense to change this behavior.

stephanebruckert:
> The idea is to add a new content script injected in pages are not supported by the extension. If this script finds some player elements on the page, it asks the core to inject the respective connector via `chrome.runtime.sendMessage`.

Do you mean that you would separate the usual `connectors.js` vs. a new `embedded-connectors.js`, and they would behave in a different way? Then we would not do `matches: ['*://*/*']` anymore?

With that solution, could the extension support a random page that contains both a Bandcamp embed and a Spotify embed? As you can guess I'm thinking of https://github.com/web-scrobbler/web-scrobbler/issues/2024

alexesprit:
> Do you mean that you would separate the usual connectors.js vs. a new embedded-connectors.js, and they would behave in a different way? Then we would not do matches: ['*://*/*'] in connectors.js anymore?

I mean to add new **content** script, which detects players on the page, and tells the core to inject the respective connector via `chrome.runtime.sendMessage`:
```js
// new-script.js
function getConnector() {
    // return connector ID (new prop) based on DOM elements displayed on the page.
}

function main() {
     const id = getConnector();
     if (id) {
         chrome.runtime.sendMessage('please inject the connector with ID=id');
     }
}

// extension.js
// Here is a handler that processes messages from new-script.
```

We do not need to split connector definitions (connectors.js vs. a new embedded-connectors.js) if it makes no sense.

stephanebruckert:
> The current implementation allows to use only one connector per page. I think there's no sense to change this behavior.

Take these two scenarios, it makes me think of a kind of many-to-many relationship between embed-connectors and domains:

1. An "existing" connector domain with any embed connector. For example:
   - `connector/nova.js` + `connector/bandcamp-embed.js` players on the same page: https://www.nova.fr/combo-lulo-ou-les-rythmes-caribeens-vus-depuis-new-york
   - `connector/residentadvisor.js` + `connector/bandcamp-embed.js` on the same domain: https://www.residentadvisor.net/reviews/23364
2. Multiple embeds on the same page (Spotify, Bandcamp, Soundcloud, Beatport). I didn't find any example but it could have good chances to happen.

PR #2032, opened 2019-08-04
Added radiojavan connector.

Slayer0248:
Added radiojavan connector.


src/connectors/radiojavan.js
```diff
@@ -0,0 +1,30 @@
...
+	return track;
+};
+
+Connector.artistTrackSelector = () => {
+	var track = $('.mainPanel .song').attr('data');
+	var artist = $('.mainPanel .artist').attr('data');
+	return { artist, track };
+};
+
+Connector.playButtonSelector = '#mp3_play';
```
inverse: Is this correct? Looking at the DOM It looks like it should be using the `#mp3_play` child span based on the class.

inverse:
```
Connector.isPlaying = () => {
	return $('#mp3_play > span').hasClass('musicPause');
};
```

alexesprit: It also could be replaced by:
```js
Connector.pauseButtonSelector = '#mp3_play .musicPause';
// or
Connector.playButtonSelector = '#mp3_play .musicPlay';
```
but the function is acceptable too.

inverse: @alexesprit I think your approach is clearer to the reader :+1:

src/connectors/radiojavan.js
```diff
@@ -0,0 +1,21 @@
...
+
+Connector.playerSelector = 'div.webPlayer';
+
+Connector.artistSelector = '.mainPanel .artist';
+
+Connector.albumSelector = '.mainPanel .album';
+
+Connector.trackSelector = '.mainPanel .song';
+
+Connector.playButtonSelector = '.playIcon';
```
inverse: This can be dropped as you've wired up the `Connector.isPlaying` function. Alternatively and perhaps better is what @alexesprit suggested and to drop `Connector.isPlaying and wire up the `stopButtonSelector` After that  It's all good

inverse:
@Slayer0248 any update?

alexesprit:
The connector does not work properly because of bugged website. It's partially reloaded on every track change (including `body` element), and the mutation observer no longer works.

The solution is to use custom mutation observer to watch the whole document:
```js
new MutationObserver(Connector.onStateChanged).observe(document, {
	childList: true, subtree: true,
	attributes: true, characterData: true,
});
```

Issue #2037, opened 2019-08-15
Please add support for 9128.live


FullyFunctionalPhil:
[9128.live](https://9128.live/) is a new streaming service dedicated to ambient/electronic music, curated by the A Strangely Isolated Place label. It launched [earlier this month](http://www.astrangelyisolatedplace.com/blog/2019/8/3/introducing-9128live-24-7-streaming).

The interface is extremely simple, and all of the music is played under permission of the artists/labels. Please consider adding support for it. I am not affiliated with the site in any way, just a longtime fan of their Mixcloud account and an avid user of Web Scrobbler.

inverse:
Thanks for posting this - didn't know they launched this!  

@alexesprit I can take this one - the side in question has an embedded iFrame - do you know if we can support this?

inverse:
@FullyFunctionalPhil will be in the next release :+1:

FullyFunctionalPhil:
Great, thank you so much!

PR #2038, opened 2019-08-15
Start support for 9128.live

inverse:
Partially solves #2037 

Works when iframe is popped out with button - need to figure out how to inject in iframe.

### DIFF
--- a/src/connectors/9128live.js
+++ b/src/connectors/9128live.js
@@ -1,3 +1,5 @@
+'use strict';
+
 Connector.playerSelector = '.radioco-player';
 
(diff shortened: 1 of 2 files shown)


alexesprit:
You can use `allFrames` flag in the `connectors.js` to allow injecting scripts in all frames.

alexesprit:
Also, seems like this website uses a 3rd-party player (https://radio.co/), so we can add support for other websites easily.

inverse:
@alexesprit how did I forget! :+1: thanks!

PR #2043, opened 2019-08-27
Update CONTRIBUTING.md

inverse:
Fix travis


alexesprit:
@inverse we already have this file stored in `.github` directory. I think we should keep one of them only.

inverse:
@alexesprit didn't see that one sorry - Makes sense :+1:

PR #2061, opened 2019-09-19
Apple Music Web Beta - Album and Title fix

Gust4Oliveira:
I updated the sources that have the song title and album information, they have undergone some minor changes in the last update of the site.

I noticed that the extension was not getting the song information I was listening to correctly, so I checked the page console and confirmed that there was a problem.

image

After changing the source for the title and album information, the extension worked again correctly.

image

Considering the site is in beta, new changes like this or others may happen.


alexesprit:
Thanks!

Issue #2064, opened 2019-09-27
Watch2Gether support


TheMusic37:
can you guys add support for watch2gether for scrobbling

alexesprit:
Closed via #2066.

PR #2066, opened 2019-09-27
Add Watch2Gether support 

jamescdcarter:
Fix for Issue #2064

Added an entry in `connectors.js` to enable [Watch2Gether ](https://www.watch2gether.com) support.

**Additional context**

Watch2Gether works by embedding a YouTube iframe.
The `connectors/youtube-embed.js` file works on this website in the same way it does on Discogs.


alexesprit:
Does it work for you? It does not for me.

inverse:
It's not for me either.

jamescdcarter:
I was woefully naive with this PR.

Not only does watch2gether support other players than just YouTube but it dynmically loads in an iframe (I don't believe this extension supports injecting the content script into dynamic iframes).

I'm looking into monkey patching some of the website's internal events to intercept player state change and track info change.

jamescdcarter:
I've added a new connector that:

 - watches the play button and timer info elements for state change using a `MutationObserver`
 - gets track info from the chat box

alexesprit:
Why did you wrap the code in a class?

jamescdcarter:
After seeing your changes I can see that I over complicated my implementation.

Your refactor looks great!

alexesprit:
Could you push these changes? I'll merge this PR then.

jamescdcarter:
I've pushed the changes and reintroduced the jQuery workaround to decode html entities (there's a bug in watch2gether that results in html entities showing in track names).

alexesprit:
> (there's a bug in watch2gether that results in html entities showing in track names).

Have you an example of such videos? I want to test if the metadata filter that replaces HTML entities will resolve the issue.

jamescdcarter:
Thank you for taking time to explain that to me.

I'd like to pick up a few other requests for connectors so I'll be sure to pay more attention to the docs and lint rules.

it looks as if watch2gether fixed this issue anyway (I've been testing against an old room, removing songs and adding them back in shows the right name now...)

alexesprit:
> it looks as if watch2gether fixed this issue anyway (I've been testing against an old room, removing songs and adding them back in shows the right name now...)

Nice, so the PR is ready to merge, right? It looks good for me, btw.

jamescdcarter:
> Nice, so the PR is ready to merge, right?

Yup, it's all good to go!

PR #2070, opened 2019-10-05
[Info] Add user play count from Last.fm

### LINKED ISSUE #2028: Display number of scrobbles of a song
Display the number of scrobbles of the song being played

anuragbhd:
It's often desirable to see how many times we have scrobbled the currently playing song in the past.

The Last.fm `track.getinfo` API returns among other things a count that represents the number of times the user has previously scrobbled the song. It is represented by the `<userplaycount>` element.

This PR makes visual and logic code changes to display user's play count as **Your scrobbles** in the Info popup.

Fixes #2028

user-play-count-web-scrobbler

**TODO:**
* Add a way to toggle 'Your scrobbles' on/off from extension options page
* Add relevant unit test(s)

cc @alexesprit

### DIFF
--- a/src/core/background/scrobbler/lastfm.js
+++ b/src/core/background/scrobbler/lastfm.js
@@ -79,7 +79,7 @@ define((require) => {
 			let trackUrl = $doc.find('track > url').text();
 			let albumUrl = $doc.find('album > url').text();
 
-			let userPlayCount = $doc.find('userplaycount').text();
+			let userPlayCount = parseInt($doc.find('userplaycount').text());
 
 			return {
 				artist, track, album, duration, userloved,
--- a/src/core/background/scrobbler/lastfm.js
+++ b/src/core/background/scrobbler/lastfm.js
@@ -79,9 +79,11 @@ define((require) => {
 			let trackUrl = $doc.find('track > url').text();
 			let albumUrl = $doc.find('album > url').text();
 
+			let userPlayCount = $doc.find('userplaycount').text();
+
 			return {
 				artist, track, album, duration, userloved,
-				artistThumbUrl, artistUrl, albumUrl, trackUrl
+				artistThumbUrl, artistUrl, albumUrl, trackUrl, userPlayCount
 			};
 		}
 	}
--- a/src/ui/popups/info.html
+++ b/src/ui/popups/info.html
@@ -30,6 +30,10 @@
 			<div class="albumArtist">
 				<a id="albumArtist" href="#" target="_blank"></a>
 			</div>
+			<div class="user-play-count">
+				<span id="userPlayCountLabel"></span>
+				<span id="userPlayCount"></span>
+			</div>
 		</div>
 		<div id="edit" data-hide="true">
 			<input type="text" id="track-input" i18n-placeholder="infoTrackPlaceholder" i18n-title="infoTrackPlaceholder" tabindex="1" />
--- a/src/ui/popups/info.html
+++ b/src/ui/popups/info.html
@@ -11,6 +11,7 @@
 	<script type="text/javascript" src="/ui/popups/info.js"></script>
 	<script type="text/javascript" src="/core/i18n.js"></script>
 
+	<link rel="stylesheet" href="/vendor/fontawesome/css/brands.min.css">
 	<link rel="stylesheet" href="/ui/popups/info.css" />
 </head>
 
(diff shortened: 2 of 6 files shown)


anuragbhd:
If the changes look good, please feel free to merge without waiting for TODO items to be implemented first. They are mostly optional.

alexesprit:
Looks neat! I have few suggestions:
1. Make font size larger a little. I changed it to 0.75rem, and it looks acceptable for me:
image
2. We have 3 scrobbling services supported, but the play count is displayed for Last.fm only, so we should note that somehow. As an option, we can add a title to the new element to indicate it's play count from Last.fm.
3. Add a top margin to the element (line spacing looks more accurate for me with top-margin set to 5px):
image

src/ui/popups/info.js
```diff
@@ -113,6 +113,11 @@ require([
 		$('#album-art').css('background-image', `url("${getCoverArt()}")`);
 	}
 
+	function fillUserPlayCount() {
+		$('#userPlayCountLabel').text(`${i18n('infoYourScrobbles')}:`);
```
inverse: What about hiding the `.user-play-count` when `getUserPlayCount` returns `0`?

Otherwise could be a bit annoying for people who do not use last.fm.

src/ui/popups/info.html
```diff
@@ -30,6 +30,10 @@
 			<div class="albumArtist">
 				<a id="albumArtist" href="#" target="_blank"></a>
 			</div>
+			<div class="user-play-count">
```
inverse: Based on what @alexesprit was saying we could use something like https://fontawesome.com/icons/lastfm to indicate this is from last.fm

src/ui/popups/info.js
alexesprit: Also, you could use string template in locales:
```json
  "infoYourScrobbles": {
    "message": "Your scrobbles: $1",
    "description": "Label for user play count"
  },
```

```js
$('#userPlayCountLabel').text(i18n('infoYourScrobbles', getUserPlayCount()));
```

anuragbhd:
@inverse and @alexesprit
Thank you for the valuable feedback. I like the suggested changes, and have incorporated them in my PR via 2 new commits.

If previously scrobbled:

user-play-count-updated-ui

If never scrobbled:

no-previous-scrobbles

src/ui/popups/info.js
anuragbhd: > Also, you could use string template in locales:
> 

Thanks for pointing that out. I deliberately kept the i18n string simple to potentially make it more convenient for translators to work on it. Perhaps it's just me who feels that way?

alexesprit: Probably, no. As for me, I think templates help translators to understand better how exactly a string is used in the extension.

Anyway, you can leave the string as is.

alexesprit:
@anuragbhd thanks! Now it looks good for me.

@inverse have you any suggestions?

src/core/background/scrobbler/lastfm.js
```diff
@@ -79,9 +79,11 @@ define((require) => {
 			let trackUrl = $doc.find('track > url').text();
 			let albumUrl = $doc.find('album > url').text();
 
+			let userPlayCount = $doc.find('userplaycount').text();
```
inverse: Would it make sense to turn this into an int and treat it as such through the pipeline? 


```suggestion
			let userPlayCount = parseInt($doc.find('userplaycount').text());
```

src/ui/popups/info.js
```diff
@@ -301,6 +313,10 @@ require([
 			'/icons/cover_art_default.png';
 	}
 
+	function getUserPlayCount() {
+		return song.metadata.userPlayCount || '0';
```
inverse:
```suggestion
		return song.metadata.userPlayCount || 0;
```

src/ui/popups/info.js
```diff
@@ -113,6 +113,17 @@ require([
 		$('#album-art').css('background-image', `url("${getCoverArt()}")`);
 	}
 
+	function fillUserPlayCount() {
+		const playCount = getUserPlayCount();
+		if (playCount === '0') {
```
inverse:
```suggestion
		if (playCount === 0) {
```

src/core/background/scrobbler/lastfm.js
anuragbhd: Yes, absolutely! There was a voice in my head that told me to go the integer way. But I guess the string voice somehow won over and made me lazy 😜. Thanks for bringing it up. My latest commit fixes this.

PR #2105, opened 2019-10-17
Fix up coverage

inverse:


.github/workflows/test.yml
```diff
@@ -5,12 +5,12 @@ jobs:
...
-    - name: Run npm install, npm test
+        cmd: install
+    - name: Run yarn run test-with-coverage
       run: |
-        npm install
-        npm test
+        yarn run test-with-coverage
       env:
         CI: true
+        CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
```
inverse: @alexesprit any idea why this is not working? states it cannot find the env var :(

inverse: It's to push code coverage results to codacy so we can track how many lines are covered with tests etc so we can get a better feel of the tests

alexesprit: Does your token work locally? Maybe it's not valid or something else?

inverse:
It works locally when I was using export

.github/workflows/test.yml
inverse: It works locally for me :/

PR #2110, opened 2019-10-19
fix up coverage

inverse:
@alexesprit it worked when I did it on a branch on this repo not my fork :/

- Changed test to use `yarn` since that's what we are locking deps with
- Introduced new coverage workflow that runs coverage against master


package.json
```diff
@@ -45,11 +44,12 @@
     "stylelint-order": "3.1.1"
   },
   "engines": {
-    "node": ">=6.0.0"
+    "node": ">=10.0.0"
   },
   "scripts": {
     "test": "grunt",
-    "test-with-coverage": "nyc --reporter lcovonly -x 'tests/**  mocha tests/**/*.js -r tests/requirejs-config.js && cat ./coverage/lcov.info | codacy-coverage && rm -rf ./coverage"
+    "test-with-coverage": "nyc -x 'tests/**' grunt",
```
inverse: This one you can run locally and see coverage output

alexesprit:
I suggest to update commands and workflow logic:
1. All `run*` scripts defined in `package.json` should use `grunt test` to run tests only.
2. Run `grunt lint` then `grunt test` in `test` workflow for a better understanding what exactly will be executed.

Also, I suggest to split `run` scripts in `package.json`:
1.  `test-with-coverage` should run tests only with `nyc`.
2. Replace `test-with-coverage-codacy` with `upload-coverage` script that will upload coverage to Codacy. This one must be executed after `test-with-coverage`.

Also, I'd merge `coverage` and `test` workflows, but added a condition: run tests with coverage if branch is master only, otherwise run tests only. This change also allows to extract linting to a separate workflow. Unfortunately, I don't know how to check if branch is master dynamically, so this suggestion is optional.

What do you think?

alexesprit:
Should we also store everything related to coverage data in `coverage` directory? I mean the `.nyc_output` and `.nyc_cache` dirs?

inverse:
@alexesprit thanks for the feedback. Whats the advantage of merging the two CI pipelines together?

alexesprit:
@inverse it will prevent running tests twice for master branch.

inverse:
@alexesprit you mean cases when you push to master directly?

alexesprit:
@inverse I mean every cases of pushing to master (direct push, merge PRs). Both `test` and `coverage` workflows will be executed for these events.

inverse:
@alexesprit makes sense

inverse:
I'll leave it as it is for now. There is a `branches-ignore` See https://help.github.com/en/articles/workflow-syntax-for-github-actions#onpushpull_requestbranchestags 

Which could be used to have similar behaviour from what I could see. But for such a simple conditional I'll keep it in one :+1:

inverse:
image


Worked :)

PR #2124, opened 2019-10-21
badge-tidy

inverse:
- Split over new lines
- Nuked codeclimate integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Chrome Web Store][WebStoreBadge]][WebStore]
 [![AMO][AmoBadge]][Amo]
+[![Gitter][GitterBadge]][Gitter]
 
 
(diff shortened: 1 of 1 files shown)


alexesprit:
What if we put badges above the "Web Scrobbler" header? I don't like the space used by badges between the header and the description (the tagline).

1. [Below the first header][1]
2. [Above the first header][2]

[1]: https://gist.github.com/alexesprit/68d53bce15fed4a6390591bbfd72f68a/a28310079ad7a16f82f6086e71b8e5607b910724
[2]: https://gist.github.com/alexesprit/68d53bce15fed4a6390591bbfd72f68a/e8fb3bd9f7bb1215ff09890867c98bfd9e359002

alexesprit:
Also, we could put Gitter badge to the first row, so the first row will contain media (?) URL.

inverse:
Sure - let me update!

PR #2132, opened 2019-10-25
Add Audiomack support

### LINKED ISSUE #2129: Support for audiomack
While listening to music on [audiomack](https://audiomack.com), the web scrobbler icon in chrome browser shows 'This website is not supported' message.

Music on the [audiomack website](https://audiomack.com) can be accessed without registering an account.

qualitymanifest:
Closes #2129

Tests and lints fine

The `getTrackArt` function is because the img src contains a query string that gives you a tiny 48px icon, so it strips out the query string and gives you the full image.

The `currentTimeSelector` doesn't appear to work and I'm not sure why. If you console.log that css selector, you get the current time, and it is in a format that is compatible with the `Util.stringToSeconds` function. The `durationSelector` has a similar css selector and works fine.

### DIFF
--- a/src/connectors/audiomack.js
+++ b/src/connectors/audiomack.js
@@ -1,0 +1,22 @@
+'use strict';
+
+Connector.playerSelector = '.player__inner';
+
+Connector.currentTimeSelector = '.player__waveform .waveform__elapsed';
+
+Connector.trackSelector = '.player__title';
+
+Connector.durationSelector = '.player__waveform .waveform__duration';
+
+Connector.artistSelector = '.player__artist';
+
+Connector.playButtonSelector = '.player__controls > .play-button--paused';
+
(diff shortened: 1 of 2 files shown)


qualitymanifest:
I see now that there was another PR for this already. I didn't see it since it didn't tag the issue. No offense taken if this gets closed in favor of that.

alexesprit:
@qualitymanifest I merged that PR because it was opened first. But it lacks album art, but your connector has. Could you add this feature either in this PR (by rebasing and adding code) or in a new PR?

alexesprit:
> The currentTimeSelector doesn't appear to work and I'm not sure why.

It works fine for me. How did you know it's not working for you? If you see in debug logs it is null, that's ok: it will be updated later.

I also slightly improved the connector in the master.

PR #2133, opened 2019-10-26
Add audiomack track art support

qualitymanifest:
Passes linting and testing

Includes track art support from previous PR https://github.com/web-scrobbler/web-scrobbler/pull/2132/


alexesprit:
Thanks!

PR #2140, opened 2019-11-02
Added a new connector for Joox, Tencent's music site

mmhan:
Joox can now be scrobbled without any dependency on other third party libraries


mmhan:
Not quite sure how I'm failing the lint here.

```
Running "unused_files" task
Fatal error: Found 1 unused files: connectors/joox.js
error Command failed with exit code 1.
```

alexesprit:
The `js` property should have `string` type, not `array`.

Also, connector tests are not used anymore.

mmhan:
They have just changed their interface again. Will rework on it.

mmhan:
All changes done. Please check @alexesprit

src/connectors/joox.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const playerSelector = '.kIpMXu';
```
alexesprit: Could you also use strings directly? For example:
```js
Connector.trackSelector = '.jlfeQT';
```

mmhan: I'd prefer to keep it this way because it's easier to change along with their client-side code.

alexesprit: I don't know why this way is easier, but the whole PR looks acceptable for me, so I resolve this conversation.

PR #2143, opened 2019-11-02
Migrate from yarn to npm

inverse:
Although yarn had it's gains back in the day I  feel like using it is more of a battle than a benefit.

This moves us back to a npm only eco-system with the addition of the `.nvmrc` it'll make it easier to maintain compatibility.


.github/workflows/deploy.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: '12.x'
-    - run: yarn install
+    - run: npm install
     - name: Deploy the extension
       run: |
-        yarn run grunt publish:chrome
-        yarn run grunt publish:firefox
+        npm run grunt publish:chrome
```
alexesprit: You should use `npx` to run local packages. `npm run` is used to run scripts from `scripts` section in `package.json`.

alexesprit:
This could be useful: https://github.com/peaceiris/actions-hugo/blob/master/.github/workflows/test.yml (Get Node.js version from `.nvmrc`).

.github/workflows/deploy.yml
inverse: WIll update - forgot about this :/

.github/workflows/deploy.yml
```diff
@@ -9,14 +9,18 @@ jobs:
...
-        node-version: '12.x'
-    - run: yarn install
+        node-version: '${{ steps.nvm.outputs.NVMRC }}'
+    - run: npm install
     - name: Deploy the extension
       run: |
-        yarn run grunt publish:chrome
-        yarn run grunt publish:firefox
+        npx run grunt publish:chrome
+        npx run grunt publish:firefox
```
alexesprit: `npx <package-name>`, not `npx run <package-name>`.

inverse: Will verify locally

inverse:
@alexesprit updated - the only thing that's open for question is we define node engine in package.json as `>=12.13.0` should we lock this to be the same as what we specify to ran against?

alexesprit:
@inverse yes, I think it should be locked, but I'm not sure if it's reqiured.

PR #2151, opened 2019-11-11
Add pan y rosas discos connector.

jaccarmac:
pan y rosas discos is a Chicago netlabel which puts out various kinds of experimental music. This connector seems to work well on most of the catalog, though it looks like pyr060 and older use a different player. Support for that player is not implemented yet.

### DIFF
--- a/src/connectors/panyrosasdiscos.js
+++ b/src/connectors/panyrosasdiscos.js
@@ -4,24 +4,18 @@ Connector.playerSelector = '#interfaceMI_0';
 
 Connector.pauseButtonSelector = '#statusMI_0 .mjp-playing';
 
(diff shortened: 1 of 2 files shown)


jaccarmac:
Does inlining it like that work or is it better to have the splitting separate so the `uniqueID` is set elsewhere?

jaccarmac:
Album art seems to be in slightly different places in the DOM every time. Can just remove that selector entirely if `:eq(0)` is too much of a hack.

alexesprit:
> Does inlining it like that work or is it better to have the splitting separate so the uniqueID is set elsewhere?

It's fine for me.

jaccarmac:
In that case we should be set.

Issue #2033, opened 2019-08-06
Please add support to Newgrounds Audio Portal


nikos3194:
**Provide information about website**
Newgrounds is a website founded in 1995 that offers space for artists, game designers and musicians. As regards the audio portal ( [https://www.newgrounds.com/audio](https://www.newgrounds.com/audio) ), many artists publish their own musical productions in the style of sites such as Bandcamp or Soundcloud.

Despite the age of the site, in the last year it gained popularity by receiving a large number of users who left the Tumblr social network.

Here are some examples of popular tracks:
https://www.newgrounds.com/audio/listen/105753
https://www.newgrounds.com/audio/listen/361366
https://www.newgrounds.com/audio/listen/75190

List of other popular tracks:
https://www.newgrounds.com/audio/browse/sort/score

**Additional context**
Screenshot:
![ng](https://user-images.githubusercontent.com/3865571/62569962-94d97e00-b865-11e9-8683-368d6f60fffe.png)

Thank you!

Ydeu:
Hello there,

As this is labelled as "good first issue", as a newcomer to the Github, I would like to try my best to resolve the issue. 

With that in mind, any additional tips on how to proceed with the "connector request" type of issues? (Tips not mentioned in wiki/Connectors-development and wiki/Setup-development-environment)

Best regards, Ydeu

alexesprit:
> With that in mind, any additional tips on how to proceed with the "connector request" type of issues? (Tips not mentioned in wiki/Connectors-development and wiki/Setup-development-environment)

No additional tips; you can freely start to write a code. After you decide it's complete you can submit a PR for review.

Ydeu:
Hello there,

I have stumbled upon some interesting names of tracks on the site. I would like to maximize the amount of successful scrobbles but I am not sure what is considered a valid track name in these cases. Also, how and if I should "bother" with some of the cases bellow. Track name examples:

"Black Cat" by Libby Shimmz feat. Black Dog
Axeph-Null - Menger's N-Arc Theorem
.- KIRBY 64 | Boss 2 -
{Heim} Ethnic Battle
-[PK]- Blue world
"The Bacon Song"
'All I Got'' (Billis23)
N3Ogouds - City of imagination (glitch hop)
<PS> A Heart Worth Mending
[Knox]Brokendeck

These names may negatively impact the result of scrobbling, as people sometimes tend to:
- Mention a genre or artist of the song in the track name in (), [], {} braces or in <>.
[...]

PR #2155, opened 2019-11-13
#2033 Newgrounds Audio Portal Connector

Ydeu:
Added a connector for [newgrounds](https://www.newgrounds.com/audio) audio portal.

**Additional context**
Not sure about the usage of document methods to get the names of artist and song in embedded player, but in the utils.js I have not found any function that could be of use and selectors did not pick up the data at all.


src/connectors/newgrounds.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const artistSelector = '.ng-apg-media-artist';
+const trackSelector = '.ng-apg-media-title';
+
+function setupConnector() {
+	if (isEmbeddedPlayer()) {
+		setupEmbeddedPlayer();
+
+		Connector.getArtistTrack = () => {
```
alexesprit: I would move `Connector.getArtistTrack` inside of `setupXXXPlayer` function.

src/connectors/newgrounds.js
```diff
@@ -0,0 +1,66 @@
...
+			let { artist, track } = Util.splitArtistTrack(document.getElementsByTagName('h2')[0].innerText, '-');
+
+			if (!artist) {
+				artist = document.getElementsByTagName('h4')[0].innerText;
+				track = document.getElementsByTagName('h2')[0].innerText;
```
alexesprit: 1. These selectors are ambiguous. I think they should be updated to make sure every selector is unique, e.g. `.item-user h4` for artist (username).
2. `Util.getTextFromSelectors` is prefferable to get text from elements, as it safer to use.

src/connectors/newgrounds.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function isEmbeddedPlayer() {
+	return $('#audio-listen-player').length > 0;
+}
+
+function setupGlobalPlayer() {
+	Connector.playerSelector = '#_ngHiddenAudioPlayer';
+
+	Connector.playButtonSelector = '#global-audio-player-play';
```
alexesprit: You should use either play or pause button selector. Using both selectors is redundant, unless there's a special case to use both (but these cases should be documented).

inverse:
Awesome :clap:

nikos3194:
Thanks so much for including this site. :smiley:  :+1:

PR #2164, opened 2019-11-19
Add global option to disable podcasts.

### LINKED ISSUE #2154: Podcasts outside Play Music
How hard would it be to keep podcasts from being scrobbled from Spotify, etc? Does Spotify even treat podcast traffic differently than music? I don't know how you would write a script for that, but if you made one it would be a helpful addition to this already great extension. Ignoring podcasts from other platforms manually as they're played isn't too challenging, but having an automated process do it would be convenient.

jaccarmac:
Closes #2154.


jaccarmac:
OK, this is going to require some thinking about how to do. See the line comments below.

src/_locales/de/messages.json
```diff
@@ -39,11 +39,11 @@
         "message": "Deaktiviere das Senden von Daten an Google Analytics",
         "description": "Option title"
     },
-    "optionGmPodcasts": {
```
jaccarmac: Locales other than English will need to be updated, I decided to update the key names instead of deleting for now since the translations should (in theory) be close to the changed versions.

src/core/content/connector.js
```diff
@@ -661,11 +669,27 @@ function BaseConnector() {
...
+	 */
+	let shouldScrobblePodcasts = true;
+	Util.getOption(Options.SCROBBLE_PODCASTS).then((sp) => shouldScrobblePodcasts = sp);
+
 	/**
 	 * Function for all the hard work around detecting and updating state.
 	 */
 	this.stateChangedWorker = () => {
-		if (!this.isScrobblingAllowed()) {
+		/**
```
jaccarmac: I decided to implement this where `isScrobblingAllowed()` is used as that's how the Google Play connector had podcast checking implemented. It could also be checked as part of the play state check, I believe. Not sure what's preferred.

src/core/content/connector.js
```diff
@@ -661,11 +669,27 @@ function BaseConnector() {
 	 */
 	this.reactorCallback = null;
 
+	/**
+	 * Used to determine whether or not podcasts should be scrobbled in onStateChanged()
+	 * @type {Boolean}
+	 */
+	let shouldScrobblePodcasts = true;
+	Util.getOption(Options.SCROBBLE_PODCASTS).then((sp) => shouldScrobblePodcasts = sp);
```
jaccarmac: Two problems here.

1. `Options` is not available in this file at the moment. Not sure if I need to do the giant `define` wrap which is in the other places the options file is `require`d or what.
2. `Util.getOption` is async. This wasn't a problem in the connector since we could write an async function and call it at the top. Here we can't add an `await` without converting the whole `BaseConnector` function async. I believe that this is equivalent to the Google Play behavior of before (initial connector load would include the default value).

jaccarmac:
Ah, one other thing I forgot to mention in that first round of comments. Currently the new option also defaults to scrobbling podcasts, which means people who have disabled GPM podcast scrobbling will need to be notified to set the new option. Alternately,

* The global behavior could be changed and everyone would have to set the option as they see fit.
* The UI for the GPM option could be removed while the code could be left in the connector.
* Some migration path for options which I missed could be used.

src/core/content/connector.js
alexesprit: I would pass `Connector.isPodcast` value to the core, and check if a track is a podcast in `Controller`, e.g. in a function where connector state is processed.

src/core/content/connector.js
alexesprit: These two issues will be resolved if we move the check to the core.

alexesprit:
> Currently the new option also defaults to scrobbling podcasts, which means people who have disabled GPM podcast scrobbling will need to be notified to set the new option.

We have a migration module, where we can convert data.

jaccarmac:
Cool, did not realize either of those things. I will take a look.

jaccarmac:
Migration still isn't quite right. On top of the migration code not seeming quite right to me, I'm seeing an error: `Unknown connector: GoogleMusic`.

jaccarmac:
That should about do it. One thing I noticed is that playing a podcast without another track causes console errors on the background page and keeps the red extension icon instead of the greyed out skip version. However, it doesn't break the extension and cause future scrobbles on the page to fail or anything. But if I've put the code in the wrong place in the state machine, let me know!

src/core/background/object/controller.js
```diff
@@ -208,6 +212,8 @@ define((require) => {
 
 			if (!isSongChanged && !this.isReplayingSong) {
 				this.processCurrentState(newState);
+			} else if (!this.shouldScrobblePodcasts && newState.isPodcast) {
+				this.skipCurrentSong();
```
alexesprit: I was thinking about ignore podcasts completely, like YouTube connector ignores videos from non-music categories.

Skip is probably ok, but songs should not be skipped in this place, as `skipCurrentSong` will be called on every state change.

src/core/background/storage/options.js
```diff
@@ -47,6 +48,12 @@ define((require) => {
 	 */
 	DEFAULT_OPTIONS[USE_UNRECOGNIZED_SONG_NOTIFICATIONS] = false;
 
+	/**
+	 * Scrobble podcast episodes. Inherits the GoogleMusic setting.
+	 * @type {Boolean}
+	 */
+	DEFAULT_OPTIONS[SCROBBLE_PODCASTS] = undefined;
```
alexesprit: Why this option is `undefined` by default?

If you want to force migration, this could be done better:
1. Check if the Google Music option is not `undefined`.
2. If true, assign the Google Music  option value to the new option; do nothing otherwise.
3. Remove Google Music option from storage.

alexesprit:
> One thing I noticed is that playing a podcast without another track causes console errors on the background page and keeps the red extension icon instead of the greyed out skip version.

Yes, in this case you check the `newState.isPodcast` and call `Controller.skipCurrentSong` befor `currentSong` property is created from a new state.

src/core/background/object/controller.js
jaccarmac: > I was thinking about ignore podcasts completely, like YouTube connector ignores videos from non-music categories.

Yeah, it occurred to me that these are more or less instances of the same thing, but for now I resisted the urge to generalize the YouTube option.

jaccarmac: This works without even adding a new flag, as `isPodcast` is already part of the song info. No visible issues, and the `newState`-related error is gone.

alexesprit:
@jaccarmac Finally, I have some time to test changes. Everything looks good for me.

The only weird thing I noticed is podcasts are editable in the info popup window.

I fixed this issue by adding a condition in `info.js`:
```diff
diff --git a/src/ui/popups/info.js b/src/ui/popups/info.js
index f0502e05..491a3562 100644
--- a/src/ui/popups/info.js
+++ b/src/ui/popups/info.js
@@ -29,7 +29,7 @@ require([
                updatePopupView();

                let isSongValid = song.flags.isValid || song.flags.isCorrectedByUser;
-               setEditMode(!isSongValid);
+               setEditMode(!(isSongValid || song.flags.isSkipped));
        }

        function setEditMode(flag) {
```
This does not look as a hack, as skipped songs aren't supposed to be edited anyway. Could you apply this change (or provide a better solution)?

And I would remove other translations, because they are incorrect a little. I think it's better to have correct, but non-translated strings.

jaccarmac:
Will do with the translations, and will apply that patch unless I suddenly dream up a better idea. To be honest, I didn't even consider the possibility of editing skipped tracks so never clicked on the icon while it was greyed out.

jaccarmac:
I opened a separate PR because it seems like a feature change and I also don't completely understand how that part of the core works (the way that flag works without the change seems opposite to me).

Issue #1972, opened 2019-05-09
PlayIrish


stephenjudge:
**Provide information about website**
PlayIrish is a new web radio station dedicated to playing new unsigned Irish artists as a showcase for radio stations to pick up the artists music and add them to their playlist.

The information webpage is at https://playirish.ie
The online webplayer is at https://playirish.ie/popup/
Looking at the code for the webplayer player you can extracted the stream feed from http://tx.sharp-stream.com/http_live.php?i=iasca.mp3 which includes current playing track metadata and plays fine in VLC. There is also references in the page code to Icecast XML so this might also be useful
A short description of the website, sample URLs, (demo) accounts for testing purposes. If (demo) accounts are not for public use, you can mail them to the project developers.

**Additional context**
I understand this is a very very niche music website, but as these are all new emerging Irish artists that are trying to get exposure and airtime, I really want to at least including my listenership in scrobbled data.

alexesprit:
I think it won't be implemented, because there's no track info on the page, and using other sources (mp3 stream or Icecast) is quite hard.

clasick:
Added a pull request that implements this player.

Issue #1733, opened 2018-09-24
Add https://provoda.ch/ connector


toby3d:

clasick:
Opened a PR for this.

PR #2184, opened 2019-12-07
Add support for PlayIrish, Imago Radio and Provoda.ch

clasick:
Added support for PlayIrish.ie as requested by @stephenjudge in https://github.com/web-scrobbler/web-scrobbler/issues/1972


src/connectors/playirish.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+Connector.playerSelector = '.qt-musicplayer';
+
+Connector.artistTrackSelector = '#streamCurrentlyPlaying';
+
+Connector.getArtistTrack = () => {
+	let text = $(Connector.artistTrackSelector).text().substring(16);
```
alexesprit: I thinks it's better to replace `NOW PLAYING:` string directly, as it's more clear.

src/connectors/playirish.js
```diff
@@ -0,0 +1,17 @@
...
+	let m = text.match(/ - /g);
+	if (m && m.length === 2) {
```
alexesprit: Have you any example where are two dashes in the artistTrackString?

alexesprit:
I would also add a `.sm2_playing` selector as a `Connector.playerButtonSelector`.

src/connectors/playirish.js
clasick: Noted, added that change

src/connectors/playirish.js
clasick: This check was actually not needed.

clasick:
> I would also add a `.sm2_playing` selector as a `Connector.playerButtonSelector`.

Done.

clasick:
Added support for Imago Radio as well.

On a side note, isPlaying and playButtonSelector both needed? Both of them implement the same check right?

clasick:
Added support for Provoda as requested at https://github.com/web-scrobbler/web-scrobbler/issues/1733

clasick:
There is a issue with Provoda.ch currently where the first track gets scrobbled, but after that, the scrobbling gets stuck on that first song. To get it scrobble, again, I need to press Pause and Play and the scrobbling will find the next song.

Any idea why this could be? I see that the #track selector is an anchor... Does that have do anything?

src/connectors/playirish.js
```diff
@@ -0,0 +1,13 @@
...
+
+Connector.artistTrackSelector = '#streamCurrentlyPlaying';
+
+Connector.getArtistTrack = () => {
+	let text = $(Connector.artistTrackSelector).text().replace('NOW PLAYING:', '');
+	return Util.splitArtistTrack(text);
+};
+
+Connector.isPlaying = () => $('.sm2_playing').length > 0;
+Connector.playButtonSelector = '.sm2_playing';
```
alexesprit: There's no sense to use `Connector.playButtonSelector` if `Connector.isPlaying` is overriden. You should use only one of them; but selectors is preferrable in most cases.

src/connectors/imago.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.now-play';
+
+Connector.artistTrackSelector = '#current-track';
+
+Connector.isPlaying = () => $('.play-icon.stop').length === 0;
+
+Connector.pauseButtonSelector = '.play-icon.stop';
```
alexesprit: The same thing for play/pause button selectors and the `isPlaying` function here.

`Connector.pauseButtonSelector` is used in default `Connector.isPlaying` function, which is overriden here.

Using `Connector.pauseButtonSelector` only is enough.

clasick:
Fixed those two comments. Any idea about this issue:

> There is a issue with Provoda.ch currently where the first track gets scrobbled, but after that, the scrobbling gets stuck on that first song. To get it to scrobble again I need to press Pause and Play and the scrobbling will find the next song.
> Any idea why this could be? I see that the #track selector is an anchor... Does that have do anything?

alexesprit:
> There is a issue with Provoda.ch currently where the first track gets scrobbled, but after that, the scrobbling gets stuck on that first song. To get it to scrobble again I need to press Pause and Play and the scrobbling will find the next song.

> Any idea why this could be? I see that the #track selector is an anchor... Does that have do anything?

This happens because of wrong `Connector.playerSelector` used. The connectors work quite simple: they observe player element which is defined by that property. If the element is changed (e.g. a new track is playing), the connector grabs data using getters.

In your case `#player` does not contains a track info, that's why the extension does not recognize track change.

So, basically, the player element should contain as much elements related to player: track info, control elements, album art, etc.

clasick:
> > There is a issue with Provoda.ch currently where the first track gets scrobbled, but after that, the scrobbling gets stuck on that first song. To get it to scrobble again I need to press Pause and Play and the scrobbling will find the next song.
> 
> > Any idea why this could be? I see that the #track selector is an anchor... Does that have do anything?
> 
> This happens because of wrong `Connector.playerSelector` used. The connectors work quite simple: they observe player element which is defined by that property. If the element is changed (e.g. a new track is playing), the connector grabs data using getters.
> 
> In your case `#player` does not contains a track info, that's why the extension does not recognize track change.
> 
> So, basically, the player element should contain as much elements related to player: track info, control elements, album art, etc.

Oooh, that makes sense. Made a change & tested it. Adding the commit to the PR.

src/core/connectors.js
```diff
@@ -1210,6 +1210,21 @@ const connectors = [{
...
+	matches: ['*://*.playirish.ie/*'],
+	js: 'connectors/playirish.js',
+	id: 'playirish',
+}, {
+	label: 'Imago Radio',
+	matches: ['*://*.imago.fm/*'],
+	js: 'connectors/imago.js',
+	id: 'imago',
+}, {
+	label: 'Provoda',
```
alexesprit: Could you also change `label` to `Provodach`? The `ch` is a part of website name (Provodach is "Проводач" for Russian). The connector file should be named `provoda.ch.js`, and the `id` property should be `provoda.ch`, see the [naming convention] for details.

[convention]: https://github.com/web-scrobbler/web-scrobbler/wiki/Connectors-development#how-to-name-connector-file

alexesprit:
> On a side note, isPlaying and playButtonSelector both needed? Both of them implement the same check right?

Sorry, didn't notice this question. The one of them is required, unless `Connector.isPlaying` is overriden.

<details>
  <summary>See the `Connector.isPlaying` default implementation (Click to expand)</summary>

```js
/**
 * Default implementation of check for active playback by play/pause button
 * selector. The state of playback allows the core to detect pauses.
 *
 * Override this method for custom behaviour.
 *
 * @return {Boolean} True if song is now playing; false otherwise
 */
this.isPlaying = () => {
    if (this.playButtonSelector) {
        const playButton = Util.queryElements(this.playButtonSelector);
        if (playButton) {
            return !playButton.is(':visible');
        }

        return false;
    }

    if (this.pauseButtonSelector) {
        const pauseButton = Util.queryElements(this.pauseButtonSelector);
        if (pauseButton) {
            return pauseButton.is(':visible');
        }

        return false;
    }

    /*
        * Return true if play/pause button selector is not specified. It's
        * better to assume the playback is always playing than otherwise. :)
        */

    return true;
};
```
Content
</details>

src/connectors/playirish.js
```diff
@@ -0,0 +1,12 @@
...
+Connector.playerSelector = '.qt-musicplayer';
+
+Connector.artistTrackSelector = '#streamCurrentlyPlaying';
+
+Connector.getArtistTrack = () => {
+	let text = $(Connector.artistTrackSelector).text().replace('NOW PLAYING:', '');
+	return Util.splitArtistTrack(text);
+};
+
+Connector.playButtonSelector = '.sm2_playing';
```
alexesprit: I'm sorry, I was wrong when I suggested to use this selector as play button selector. There should be `Connector.pauseButtonSelector` instead. Currently the extension detects playing state as paused.

alexesprit:
Looks fine for me now. Thanks for contributions!

clasick:
> Looks fine for me now. Thanks for contributions!

No problem. Amazing extension (and codebase) that you're maintaining here!

namikiri:
Hi all!

Maintainer of Provoda.ch (@provodach) here. Thank you for attending my radio station.
I've just downloaded this extension and saw the station in supported service list.

As I can see you are trying to scan `#track` and `#player` elements to detect the track and listening status. I want to offer you something more convenient than hacking the HTML. The player has separate info about artist/track (comes from the server) and playing state. So I can implement the method to get it or an event dor you to subscribe for (something like ontrackchange/onplay/onstop). Any requests or suggestions are welcome! 

Thank you for your extension~

alexesprit:
@namikiri Hi! That would be great.

The extension uses this approach for several websites. Here is an example how it works inside the extension:
- [src/connectors/musickit-dom-inject.js][1]
- [src/connectors/yandex-music-dom-inject.js][2]

I like how Yandex did this.

[1]: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/musickit-dom-inject.js
[2]: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/yandex-music-dom-inject.js

Issue #2152, opened 2019-11-12
Atomic Music Space scrobbling


Vyktory:
**Is your feature request related to a problem? Please describe.**
A new streaming service recently launched in Australia named the Atomic Music Space. There is no web scrobbler for it. I'd like this one to be compatible.

**Additional context**
https://stream.atomicmusic.space/ is the website's URL. It's small but the people behind it have a network of radio stations in regional Australia where it will be the focus of a radio countdown and A&R scouting.

clasick:
Added support for this through PR as mentioned above.

PR #2185, opened 2019-12-07
Added support for Atomic Music Space

clasick:
Added support for Atomic Music Space as requested by @Vyktory in https://github.com/web-scrobbler/web-scrobbler/issues/2152


clasick:
Not sure why there are 8 commits in this PR (only the last two are related to this), but the diffs only show the proper changes. Hope this is fine.

alexesprit:
> Not sure why there are 8 commits in this PR (only the last two are related to this), but the diffs only show the proper changes. Hope this is fine.

You can use separate branches for PRs to avoid this.

clasick:
> > Not sure why there are 8 commits in this PR (only the last two are related to this), but the diffs only show the proper changes. Hope this is fine.
> 
> You can use separate branches for PRs to avoid this.

Cool. I'll remember this for next time.

alexesprit:
The connector works fine for me. 

Could you also improve it by adding support for current time and duration?

clasick:
> The connector works fine for me.
> 
> Could you also improve it by adding support for current time and duration?

Sure, I'll look into it.

clasick:
> The connector works fine for me.
> 
> Could you also improve it by adding support for current time and duration?

How do I test if the added current time and duration are working? Any easy way, or any changes from  the default behavior that would be noticeable?

alexesprit:
> How do I test if the added current time and duration are working? Any easy way, or any changes from the default behavior that would be noticeable?

There are two options:
1. Use Chrome DevTools Console (Ctrl+Shift+I, "Console" tab). The extension prints track info on every track change in readable format. Note that `currentTime` property can be null, that's ok.
2. Use debug view in the info popup window. You can Alt-click on the popup body, and the debug view will be displayed.

Note that these ways are not available if you running the extension from `build` directory, but not from `src` directory directly. To be honest, I don't know if it's an issue or not; personally I use the extension from the source directory and don't build it, and I really don't know how other people develop and debug the extension. :(

image

alexesprit:
> I realize it must be something to do with "scrobble after 50% of the song is done" or something like that. I can find remainingTime being used by getCurrentState() in connect.js, but I can't find the region of code that uses that.

Yes, duration is used by the core to calculate scrobbling point correctly (50% of the duration or 240 seconds). Current time is not used directly, but it's should be implemented if available to use it in future.

Remaining time is not used by the core directly, but is used by base connector to calculate current time or duration. Some websites contain remaining time instead of current time or duration, and `Connector.getRemainingTime` or `Connector.remainingTimeSelector` can be used for such cases. In the context of Atomic Music Space, remaining getter/selector should not be used.

clasick:
Thanks for the info. I was able to confirm that the previous commit works correctly. 

One thing of note is that running the extension from `src` directly doesn't work on Firefox, had to switch to Chromium to test that part. Not sure if it's just my setup or not either.

alexesprit:
> One thing of note is that running the extension from src directly doesn't work on Firefox, had to switch to Chromium to test that part. Not sure if it's just my setup or not either.

You're totally right, that's an issue for Firefox users/devs. I'll fix that. Thanks for opening my eyes to this problem.

The connector is perfect now. Merging right away.

PR #2189, opened 2019-12-07
Add rules for stale bot

### LINKED ISSUE #2188: Dealing with old issues
We have a lot of legacy tickets - some feature requests, others issues that cannot be reproduced. To help reduce the noise and to keep our issues under control how about using some of the bots that are available on the marketplace such as: https://github.com/apps/stale

Opinions?

inverse:
Resolves #2188


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
```
inverse: Happy to remove these comments - or do you think they help explain?

alexesprit: I think it's quite enough to understand the config w/o comments - properties are named good.

inverse: Removing :)

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - feature
+  - request
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
alexesprit: Maybe use `outdated` label? I created `wontfix` label to mark issue that won't be resolved due to some reasons.

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
```
alexesprit: I'd add a new line to make GitHub happy :D

inverse: rly need to change my editor to auto do this :/

alexesprit:
Looks good for me. If you have no additional changes, we can try to use this config.

PR #2191, opened 2019-12-08
Cross Platform Builds

a93h:
Copy vendor is not longer limited to mac and linux users.
Windows is easily built, debugged and tested now.

Duplicated bash functionality to be replace with node fs library.

**Additional context**
Bootstrap would not load and other new functionality by default on Windows.
This is no longer a problem.
Windows developers can join the build process with ease.


package.json
```diff
@@ -60,7 +61,7 @@
     "test-with-coverage": "nyc grunt test",
     "upload-coverage": "cat ./coverage/lcov.info | codacy-coverage",
     "clean-coverage": "rm -rf ./coverage .nyc_cache .nyc_output",
-    "postinstall": "bash copy-vendor.sh"
+    "postinstall": "node shell-scripts/copy-vendor.js"
```
alexesprit: Could you use `scripts` directory?

shell-scripts/copy-vendor.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+
+const shell = require("shelljs");
```
alexesprit: Node has built-in `fs` module. Could you use it instead of this dependency?

shell-scripts/copy-vendor.js
```diff
@@ -0,0 +1,49 @@
...
+
+copyDep(bootstrap_dist + "/css/bootstrap.min.css", bootstrap_css);
+copyDep(bootstrap_dist + "/js/bootstrap.bundle.min.js", bootstrap_js);
+
+copyDep(fontawesome_dist + "/css/solid.min.css", fontawesome_css);
+copyDep(fontawesome_dist + "/css/brands.min.css", fontawesome_css);
+copyDep(fontawesome_dist + "/css/fontawesome.min.css", fontawesome_css);
+
+copyDep(fontawesome_dist + "/webfonts/fa-solid-900.woff2", fontawesome_webfonts);
+copyDep(fontawesome_dist + "/webfonts/fa-brands-400.woff2", fontawesome_webfonts);
```
alexesprit: This whole file should be added to the list of files to lint:
```js
// Gruntfile.js

// Files to lint
const JS_FILES = [
	'*.js', 'scripts/*.js',
	// Custom Grunt tasks
	'.grunt',
	// Connectors
	`${SRC_DIR}/connectors/**/*.js`,
	// Core files
	`${SRC_DIR}/core/**/*.js`, `${SRC_DIR}/ui/**/*.js`,
	// Tests
	'tests/**/*.js'
];
```

package.json
```diff
@@ -60,7 +61,7 @@
     "test-with-coverage": "nyc grunt test",
     "upload-coverage": "cat ./coverage/lcov.info | codacy-coverage",
     "clean-coverage": "rm -rf ./coverage .nyc_cache .nyc_output",
-    "postinstall": "bash copy-vendor.sh"
```
inverse: We should remove this old `sh` file then since makes no sense to maintain multiple versions

alexesprit: Thanks, I noticed it too, but forgot to mention.

package.json
a93h: ```scripts/copy-vendor.js```
or 
```src/copy-vendor.js```

File: shell-scripts/copy-vendor.js (hunk 2 above)
Comment (a93h) [reply shell-scripts/copy-vendor.js]: I just did, only this waiting for you to answer the question above for push.

File: shell-scripts/copy-vendor.js (hunk 3 above)
Comment (a93h) [reply shell-scripts/copy-vendor.js]: Done waiting for answer to above question for push.

File: package.json (hunk 4 above)
Comment (a93h) [reply package.json]: Done, just waiting on first question to push.

File: package.json (hunk 1 above)
Comment (a93h) [reply package.json]: I went with my gut instinct on this one.

Comment (alexesprit) [reply package.json]: > scripts/copy-vendor.js

Yes, it's fine for me.

File: Gruntfile.js [hunk 5]
```diff
@@ -36,6 +36,8 @@ const JS_FILES = [
 	`${SRC_DIR}/connectors/**/*.js`,
 	// Core files
 	`${SRC_DIR}/core/**/*.js`, `${SRC_DIR}/ui/**/*.js`,
+	// Shell Scripts
+	`${SRC_DIR}/shell-scripts/*.js`,
```
alexesprit: You should use `scripts/*.js` instead of this, as scripts are in `scipts` directory, not in `src/shell-scripts`.

a93h: Sorry for the multiple commits. You might want to squash this commit.

alexesprit: I'll squash it anyway.

Gruntfile.js
```diff
@@ -36,6 +36,8 @@ const JS_FILES = [
 	`${SRC_DIR}/connectors/**/*.js`,
 	// Core files
 	`${SRC_DIR}/core/**/*.js`, `${SRC_DIR}/ui/**/*.js`,
+	// Shell Scripts
+	`${SRC_DIR}/scripts/*.js`,
```
alexesprit: You will have this path as `src/scripts/*.js` after it's expanded. You should use `scripts/*.js` instead.

scripts/copy-vendor.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+    });
+}
+
+function copyDep(src, in_dest) {
+    
+    const dest = (in_dest === undefined) ? vendor : in_dest;
+    const dep = path.basename(src);
+
+    fs.writeFile(src, dest, (err) => {
```
alexesprit: You should use `fs.copyFile` to copy files. Note that destination path should contain file name:
```js
fs.copyFile(src, `${dest}/${dep}`, (err) => {
    ...
}
```

`fs.writeFile` is currently overwriting files in `node_modules` directory.

a93h: Yes. Sorry for the delay. I should have kept it local until I knew 100%.

a93h:
```javascript
 fs.copyFile(src, dest + path.sep + dep, (err) => {
        ...
    });
```

scripts/copy-vendor.js
```diff
@@ -0,0 +1,91 @@
...
+        copyDep(fontawesome_dist + "/css/fontawesome.min.css", fontawesome_css);
+
+        copyDep(fontawesome_dist + "/webfonts/fa-solid-900.woff2", fontawesome_webfonts);
+        copyDep(fontawesome_dist + "/webfonts/fa-brands-400.woff2", fontawesome_webfonts);
+    }
+    if (wait_counts < 1 || mkdir_count == 0) {
+        clearInterval(interval);
+    }
+    --wait_counts;
+}, 50);
```
alexesprit: Why do you use `setInterval`?

a93h: I did not want to include any async libraries.

alexesprit: 1. You can use `fs.*Sync` variants, e.g. `fs.copyFileSync`.
2. You can promisify `fs` functions, and use `async/await`.

The `setInterval` solution is really weird.

a93h: Ok . So I made this change and it's really much prettier code.
I agree and you are right.

a93h: Just finishing up lint requirements.

Gruntfile.js
a93h: I believe it is fixed.

Issue #1677, opened 2018-07-26
Amazon Music web player scrobbles wrong album


zachryzion:
When scrobbling from https://music.amazon.com/stations, the album is scrobbled as the station name (in my case "Prime Alternative Station"), instead of the real album.
### Expected behavior

It should scrobble the real album for each song/Artist

### Current behavior

It scrobbles the station name

### Possible solution

Do not take the album from the station. I think by default you search for a possible album.

### Steps to reproduce (for bugs)

Log in to https://music.amazon.com/ and start any music station and you can see the scrobble activity.

1. Log on to https://music.amazon.com/stations
2. Fire up any station
3. See that it scrobbles the station name as the album.

### Context

3-4 days, ever since I have started the station.

### Steps for resolving the issue

- [ ] Reauthenticate on Last.fm

- [ ] Try to use unpacked extension from latest master

- [ ] Check other issues for duplicates

### Environment

* Version used: 67.0.3396.99
* Browser name and version: Chrome 67.0.3396.99

clasick:
After #2193, this bug is not reproducible. Since it won't take the station name as the album name, it gets scrobbled without an album name. Last.fm then tries to it using the Artist and Track name alone. 

I think this is the intended behavior at this point. Can close this.

alexesprit:
@clasick Thanks for the info.

PR #2193, opened 2019-12-08
Add "feat. Artist" filter for Amazon tracks, Album Artist data from Artist tag and improved Explicit tag handling

clasick:
Added a filter that removes "[feat. Artist]" tags from the track's `Title` metadata in Amazon Music.

Some tracks contain "(Explicit Version)" that needed to be handling for "Explicit" filtering. Added this as a case under "Version" filtering. 

**Additional context**
Effectively, all the changes combined turn this:
image
into this:
image

### DIFF
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -6,6 +6,8 @@ Connector.getArtist = () => {
 	return $('.trackInfoContainer .trackArtist a, .trackInfoContainer .trackArtist span').attr('title');
 };
 
+Connector.getAlbumArtist = Connector.getArtist;
+
 Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
 Connector.albumSelector = '.trackSourceLink a';
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -10,6 +10,24 @@ Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
 Connector.albumSelector = '.trackSourceLink a';
 
+Connector.getAlbum = () => {
+	if ($('.trackSourceLink a').attr('href').includes('albums')) {
+		return $('.trackSourceLink a').text();
+	}
+
+	if ($('tr.selectable.currentlyPlaying td.albumCell')) {
+		return $('tr.selectable.currentlyPlaying td.albumCell').attr('title');
+	}
+
+	if ($('.nowPlayingDetail img.albumImage')) {
+		return $('.nowPlayingDetail img.albumImage').att('title');
+	}
+
+	if ($('.trackSourceLink a').data('ui-click-action') === 'selectAlbum') {
+		return $('.trackSourceLink a').attr('title');
+	}
+};
+
 Connector.currentTimeSelector = '.songDuration.timeElapsed';
 
 Connector.playButtonSelector = '.rightSide .playbackControls .playerIconPlay';
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -10,8 +10,6 @@ Connector.getAlbumArtist = Connector.getArtist;
 
 Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
-Connector.albumSelector = '.trackSourceLink a';
-
 Connector.getAlbum = () => {
 	if ($('.trackSourceLink a').attr('href').includes('albums')) {
 		return $('.trackSourceLink a').text();
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -8,19 +8,7 @@ Connector.getArtist = () => {
 
 Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
-Connector.getAlbum = () => {
-	if ($('tr.selectable.currentlyPlaying td.albumCell')) {
-		return $('tr.selectable.currentlyPlaying td.albumCell').attr('title');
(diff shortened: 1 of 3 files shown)


alexesprit:
Does Amazon always contain featuring artists in "Artist" field? I believe the common format is `Artist A — Track (feat. Artist B)`, but not `Artist A feat. Artist B — Track`.

clasick:
> Does Amazon always contain featuring artists in "Artist" field?
> 
> I believe the common format is `Artist A — Track (feat. Artist B)`, but not `Artist A feat. Artist B — Track`. What about removing featuring artists from `artist` field?

Yes, I looked into many albums and they always contain the format `Artist A feat. Artist B — Track`. I was going to fix this next, except I had a doubt. By most conventions on other scrobblers I've seen, they keep the `Artist` tag as how it's present in Amazon. The `Album Artist` tag would be the one withe the `feat. Artist B` part removed. 

How is it generally implemented on `web-scrobbler`? Following conventions, we can keep the `Artist` as is and add data for `Album Artist` and send both in the scrobble data.

alexesprit:
> Yes, I looked into many albums and they always contain the format Artist A feat. Artist B — Track. I was going to fix this next, except I had a doubt. By most conventions on other scrobblers I've seen, they keep the Artist tag as how it's present in Amazon.

I would keep the current behavior in PR, if `Artist` field **always** contains featuring artists, and `Track` field contains featuring artists **sometimes**.

Otherwise, I would remove featuring artists from `Track` field.

> How is it generally implemented on web-scrobbler?

In most cases, connectors return track info as it's displayed on website; some connectors implement their own filters to modify track fields. The core does not modify this info directly before scrobbling; it sends a request to Last.fm to correct track info, though).

alexesprit:
Could you also rebase to latest master and add tests for new filter rules? See [tests/content/filter.js][tests] for examples.

[tests]: https://github.com/web-scrobbler/web-scrobbler/blob/master/tests/content/filter.js

clasick:
> > Yes, I looked into many albums and they always contain the format Artist A feat. Artist B — Track. I was going to fix this next, except I had a doubt. By most conventions on other scrobblers I've seen, they keep the Artist tag as how it's present in Amazon.
> 
> I would keep the current behavior in PR, if `Artist` field **always** contains featuring artists, and `Track` field contains featuring artists **sometimes**.

Yes, this is the case. I was thinking scrobbling "Album Artist" field in this case with "feat. Artist" removed along with "Artist" would be the correct move, and let Last.fm figure it out on their end. (please refer to the answer to my issue in another project [here](https://github.com/hrkfdn/mpdas/issues/48#issuecomment-341999852)).

ex. `Artist` = The Weeknd feat. Ed Sheeran
      `Album Artist` = The Weeknd

> Otherwise, I would remove featuring artists from `Track` field.
Yes, it is how the code in this PR works right now.

> > How is it generally implemented on web-scrobbler?
> 
> In most cases, connectors return track info as it's displayed on website; some connectors implement their own filters to modify track fields. The core does not modify this info directly before scrobbling; it sends a request to Last.fm to correct track info, though).

I was thinking we should formulate some kind of standard similar to what Last.fm themselves suggested in their [Correction guidelines](https://getsatisfaction.com/lastfm/topics/what-are-the-guidelines-for-correcting-artist-and-track-names) and write connectors which adher to that, because the user would be expecting scrobbling data to be handled in a similar way throughout, and not differing between sites.

> Could you also rebase to latest master and add tests for new filter rules? See tests/content/filter.js for examples.

Sure, I'll get on this.

clasick:
Another idea is to scrobble both by default (for all connectors) and let Last.fm decide what to do with it. We can then add a option toggle in the Preferences that can force `Album Artist` to be scrobbled as `Artist` if the user desires (many users including myself scrobble it this way).

clasick:
I can confirm this behavior now. Scrobbling with both `Artist` (with feat. Artist B) and `Album Artist` (without feat. Artist B) leads Last.fm to intelligently use it for both formats i.e. the scrobble will show up under the `Artist A` while also under `Artist A feat. B` while counting as only one unique scrobble.

src/connectors/amazon.js
```diff
@@ -6,9 +6,17 @@ Connector.getArtist = () => {
...
+
 Connector.trackSelector = '.trackInfoContainer .trackTitle';
 
+Connector.albumSelector = '.trackSourceLink a';
+
 Connector.getAlbum = () => {
+	if ($('.trackSourceLink a').attr('href').includes('albums')) {
+		return $('.trackSourceLink a').text();
+	}
+
```
clasick: Album tag was broken. This check fixes that.

clasick:
Looks like I broke some tests because of function renaming. Will fix.

clasick:
Album Artist from Artist in action:
image
gets scrobbled as:
image

alexesprit:
Awesome work. I have no access to Amazon, so I can't verify changes. I believe you done things right.

src/connectors/amazon.js
alexesprit: Does this change fix #1677?

clasick:
> Awesome work. I have no access to Amazon, so I can't verify changes. I believe you done things right.

Yup, I've been using this test version for the past couple days, seems to be working fine.

There are a couple of other issues in Amazon that happen rarely, I'll capture try to address those as well later on.

PR #2221, opened 2020-01-02
Add SiriusXM Player .ca (canada) url match pattern

brodie-g:

player.siriusxm.ca was not being recognized as a supported website without adding a custom match pattern.
Adding this pattern so the .ca app will be recognized by default.


alexesprit:
Thanks!

Issue #2232, opened 2020-01-07
Please add support to Chillhop MUSIC


Rostov84:
Please add support to two radio stations from site chillhop.com:

1. https://chillhop.com/live/chillhop-radio/
2. https://chillhop.com/live/lofihiphop/

Thank You!

alexesprit:
Added in #2293.

PR #2293, opened 2020-03-05
Add Chillhop connector

folsomwg:
Add Chillhop connector, requested here: https://github.com/web-scrobbler/web-scrobbler/issues/2232

### DIFF
--- a/src/connectors/chillhop.js
+++ b/src/connectors/chillhop.js
@@ -1,0 +1,11 @@
+'use strict';
+
+Connector.playerSelector = '#player-controlls';
+Connector.pauseButtonSelector = '.fa-pause';
+Connector.durationSelector = '.track-length';
+Connector.artistSelector = '.jp-artists';
+
+const trackSelector = '.jp-title';
+Connector.getTrack = () => {
+	return Util.getTextFromSelectors(trackSelector);
+};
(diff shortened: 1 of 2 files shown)


src/connectors/chillhop.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+Connector.playerSelector = '#player-controlls';
```
folsomwg: This isn't a typo, it's how it is in the page.

src/connectors/chillhop.js
```diff
@@ -0,0 +1,11 @@
...
+Connector.pauseButtonSelector = '.fa-pause';
+Connector.durationSelector = '.track-length';
+Connector.artistSelector = '.jp-artists';
+
+const trackSelector = '.jp-title';
+Connector.getTrack = () => {
+	return Util.getTextFromSelectors(trackSelector);
```
alexesprit: Why don't you use `trackSelector` directly?

src/connectors/chillhop.js
alexesprit: Could you add a comment about this in the code?

src/connectors/chillhop.js
folsomwg: Fixed.

src/connectors/chillhop.js
folsomwg: Done.

Issue #784, opened 2015-09-14
High-Contrast-Mode/Theming ...


MartinL0815:
Hi,

I've started to address a missing high-contrast-mode compatibility in Chrome extension UIs.
Since I have to start with one extension , I'll start with this.

Since I have eye disabilities, I needed to switch my Windows PC to a high-contrast-theme and to 150% UI scaling.
Since then most applications behave like expected, with some exclusions - eventually in detail.

A big exclusion are all of the extensions of Chrome, not caring a bit, like Chrome, which gives a shit on it!
I needed to install hackers vision to read web pages and a dark design to let Chrome look dark, while even the settings pages don't care about my yellow/black high-contrast-mode.
But is this an excuse for all the extensions ignoring this high-contrast-mode in their UI, too?

In my ill eyes: NO!

Here a screenshot of web-scrobblers popup: 
![web-scrobbler popup](https://cloud.githubusercontent.com/assets/525700/9853131/63adf37a-5b02-11e5-93e7-030804af99b0.png)
I nearly can't see the heart to tell LastFM, I love the track and I nearly can't read the artists name on the light background.

Is there a chance to change the visualization corresponding to the high-contrast-mode and UI scaling?

What's about using the system colors for window backgrounds, texts, selections, frames in the popup?
This would result in a black background, yellow text and eventually a white bordered heart without shading?

Any chance?

Best regards,

Martin

artlung:
I just activated it in my Windows 7 / Chrome Version 45.0.2454.85 m. I get this strange look and my Chrome does not respect it fully at all. Is there a trick for Chrome to do this properly? Do others see this? `Left Shift + Left Alt + Print Screen` what I see is this:

My folders, and Outlook are all in a high contrast mode but Chrome is sort of semi-contrasty:

![highcontrast](https://cloud.githubusercontent.com/assets/70380/9855331/8a82fb84-5ac2-11e5-9c14-25e79eadff86.png)

It looks like there is an API for getting highContrast - but that's just googling: https://developer.chrome.com/extensions/accessibilityFeatures#property-highContrast -- can't tell if this is the same or different or linked to OS high-contrast.

inverse:
@artlung It looks like that API might be only available on ChromeOS

> Important: This API works only on Chrome OS.

_https://developer.chrome.com/extensions/accessibilityFeatures_

artlung:
@inverse Thanks for reading closer then me. That's disappointing.

[...]

PR #2330, opened 2020-03-28
Use icons in info popup window

alexesprit:
The reason why I replaced text buttons with icons is to have compact controls for **each** language.

Text buttons look fine when English language is set, but in others (e.g. Russian) they have pretty long labels.

Another way to resolve the issue is to increase width of the info popup window, but it has cons:
1. The info with increased width looks quite weird.
2. You don't know if it's increased enough.

Pros:
1. Love/unlove icon now is consistent, as we use icons only.
*can't find more pros :(*

**Screenshots**

image
*Screenshot of the info popup window after this PR is applied.*

image
*Screenshot of the info popup window from `master` branch.*

image
*Screenshot of the info popup window with increased width.*

### DIFF
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -375,10 +375,6 @@
         "message": "Kein Lieblingslied",
         "description": "Label of love/unlove button"
     },
-    "infoEdit": {
-        "message": "Bearbeiten",
-        "description": "Label of edit button"
-    },
     "infoEditTitle": {
         "message": "Titel-Metadaten bearbeiten",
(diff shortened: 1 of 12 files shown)


alexesprit:
@inverse What do you think?

inverse:
The idea seems good - I didn't realise how warped the view looked in other languages but I have concerns about the usability of the skip icon - however I am not sure what would be better.

inverse:
Woah - now that you made the icons use FA ones we could prob make some more progress with this ticket  #784 which was blocking me before.

PR #2331, opened 2020-03-29
Change Codacy URL to gh org, not legacy

inverse:


inverse:
@alexesprit I don't like the way my face comes up with the integration now :( Will sync with them on this to see if there is a work around as It shows as I added it - we use them at work so have quite a good business relationship with them.

alexesprit:
@inverse wow, that would be great!

PR #2351, opened 2020-04-11
✨ PocketCasts support (Fixes #1419)

### LINKED ISSUE #1419: Support PocketCasts web ui?
Would be nice to scrobble from https://play.pocketcasts.com/web/

### Expected behavior
podcast episodes would scrobble to last.fm as a music track would

### Current behavior
I don't see this happening, assuming by design. 

### Environment
* Browser name and version: Chrome 60.0.3112.113 (Official Build) (64-bit) on MacOS Sierra 10.12.6 (16G29)

goodevilgenius:
Added support for PocketCasts web (https://play.pocketcasts.com/)

### DIFF
--- a/src/connectors/pocketcasts.js
+++ b/src/connectors/pocketcasts.js
@@ -1,0 +1,17 @@
+'use strict';
+
+/* global Connector, Util */
+
+Connector.playerSelector = '.player-controls';
+Connector.artistSelector = '.podcast-title';
+Connector.trackSelector = '.episode-title';
+Connector.playButtonSelector = '.play_button';
+Connector.currentTimeSelector = '.current-time';
+Connector.remainingTimeSelector = '.time-remaining';
+Connector.trackArtSelector = '.player-image img';
+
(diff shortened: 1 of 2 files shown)


src/connectors/pocketcasts.js
```diff
@@ -10,7 +8,7 @@ Connector.currentTimeSelector = '.current-time';
 Connector.remainingTimeSelector = '.time-remaining';
 Connector.trackArtSelector = '.player-image img';
 
-let origRemainingSelector = Connector.getRemainingTime;
+const origRemainingSelector = Connector.getRemainingTime;
```
alexesprit, line 11: Can we have `origGetRemainingTime` instead? This is not a selector, but a function, so the variable name might be confusing.

goodevilgenius: Good catch

alexesprit:
Can't test, so I believe it works fine. :)

Thanks for contributing!

PR #2382, opened 2020-04-25
Add DNA Lounge connector

lindner:
- Supports video player on webcast/video.html
- Only uses artist + track
- Removes the year and (uncensored) variants

Still needs support for custom, non-video webcasts

### DIFF
--- a/src/connectors/dnalounge.js
+++ b/src/connectors/dnalounge.js
@@ -1,0 +1,18 @@
+'use strict';
+
+const DNA_PREFIX = 'DNA Pizza Music Videos: ';
+
+// The Artist+Track is displayed in the text of this Element
+Connector.playerSelector = '#metadata';
+Connector.artistTrack = '#metadata';
+
+Connector.getArtistTrack = () => {
+  let [artist, track] = $('#metadata').text().replace(DNA_PREFIX, '').split(' -- ');
+  // dnalounge has optional (uncensored) text and the year in the track name.
+  track = track.replace(' (uncensored)', '').replace(/ \([0-9]{4}\)[ ]*$/, '');
+  return { artist, track };
+};
+
+Connector.isPlaying = () => {
+  return ! $('.video_embed').hasClass('vjs-pause');
+};
--- a/src/connectors/dnalounge.js
+++ b/src/connectors/dnalounge.js
@@ -8,8 +8,8 @@ Connector.artistTrack = '#metadata';
 
 Connector.getArtistTrack = () => {
 	let [artist, track] = $('#metadata').text().replace(DNA_PREFIX, '').split(' -- ');
-	// dnalounge has optional (uncensored) text and the year in the track name.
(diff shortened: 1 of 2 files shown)


src/core/connectors.js
```diff
@@ -1700,6 +1700,13 @@ const connectors = [{
 	],
 	js: 'connectors/rtbf.js',
 	id: 'rtbf',
+}, {
+	label: 'DNA Lounge',
+	matches: [
+		'*://www.dnalounge.com/webcast/video.html'
```
alexesprit, line 1706: Could you change this pattern to "*://www.dnalounge.com/*"?

src/connectors/dnalounge.js
```diff
@@ -0,0 +1,18 @@
...
+
+// The Artist+Track is displayed in the text of this Element
+Connector.playerSelector = '#metadata';
+Connector.artistTrack = '#metadata';
+
+Connector.getArtistTrack = () => {
+	let [artist, track] = $('#metadata').text().replace(DNA_PREFIX, '').split(' -- ');
+	// dnalounge has optional (uncensored) text, star, and the year in the track name.
+	track = track.replace(' (uncensored)', '').replace(' ★', '').replace(/ \([0-9]{4}\)[ ]*$/, '');
+	return { artist, track };
```
alexesprit, line 13: Could you use MetadataFilter instead of replacing text directly?

src/connectors/dnalounge.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const DNA_PREFIX = 'DNA Pizza Music Videos: ';
+
+// The Artist+Track is displayed in the text of this Element
+Connector.playerSelector = '#metadata';
+Connector.artistTrack = '#metadata';
```
alexesprit, line 7: `artistTrack` is a not valid property.

src/connectors/dnalounge.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const DNA_PREFIX = 'DNA Pizza Music Videos: ';
+
+// The Artist+Track is displayed in the text of this Element
+Connector.playerSelector = '#metadata';
```
alexesprit, line 6: You should use other selector as a `playerSelector`, otherwise, play/pause won't work. `.video_embed` which is used to check playing state, should be a child of a `playerSelector` element.

lindner:
Thanks for the detailed feedback.  I haven't written one of these for a while.  I'll give this a good go over and apply your suggestions.

There's also another page that has an audio player that I'll add support for

inverse:
Hi @lindner looks like this branch has some minor merge conflicts that also needs to be addressed. Need any help on this PR?

lindner:
I'll get back to it this week, will get there soon!

inverse:
@lindner the bot auto-closed the issue as it wasn't touched for a while. Do you need any help with it?

PR #2391, opened 2020-05-04
fix: youtube music connector scrobbles wrong info for tracks with specific chars

aamelin:
Youtube Music connector splits song titles by some characters (dash, slash etc.) assuming the resulting parts are the song title and the artist name. PR removes this behaviour and uses the data obtained from selectors.

**Additional context**
The problem was in the usage of the utility, apparently, helpful in case of Youtube videos where there is no dedicated fields for artist and title names, but it is often harmful in the context of Youtube Music since a lot of track names use dashes or slashes, e.g.:

Shine On You Crazy Diamond (Pts. 1-5) Pink Floyd
Cross-Eyed Mary Jethro Tull
Pygmy Twylyte (Live / Helsinki, Finland / 1974) Frank Zappa

are all split incorrectly substituting the second part of the split result as the artist name.


alexesprit:
This change breaks scrobbling tracks like [this][example].

[example]: https://music.youtube.com/watch?v=Ap1fDjCXQrU

aamelin:
That is correct, there is no way to tell when the title is filled in properly, but the workaround doesn't really fix the problem either.

aamelin:
A clarification: the example track is a 'video', thus it follows youtube's naming convention, where the artist is a channel name, and the title is free form, which is impossible to process correctly with regexes anyway. Assuming that the main product is (properly named) music tracks the simpler connector code gives better results.

alexesprit:
> Assuming that the main product is (properly named) music tracks the simpler connector code gives better results.

You're right.

PR #2409, opened 2020-05-18
Bugfixes for 1001tracklists connector

WhiteAls:
Play\pause state was not tracked due to wrong playerSelector selector.
Apparently mixDuration can absent in the first player, so we check them all. For example [this mix tracks](https://1001.tl/12pz97yk) will not scrobble.
It turns out that cue and nextCue variables can be obtained directly in seconds.

### DIFF
--- a/src/connectors/1001tracklists.js
+++ b/src/connectors/1001tracklists.js
@@ -3,9 +3,9 @@
 let nextCue;
 let cue;
 let currentTime;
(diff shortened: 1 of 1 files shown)


alexesprit:
@WhiteAls Hi! Could you check a jQuery-less version of this connector and fix it instead, if it's needed?

https://github.com/web-scrobbler/web-scrobbler/blob/feature/no-jquery/src/connectors/1001tracklists.js

alexesprit:
Nevermind, I'll update that by myself.

PR #2410, opened 2020-05-19
Add SoundClick connector.

xyauhideto:
Add a connector for [SoundClick](https://www.soundclick.com/).

### DIFF
--- a/src/connectors/soundclick.js
+++ b/src/connectors/soundclick.js
@@ -17,8 +17,7 @@ function cleanupTrack(text) {
 	const artist = Connector.getArtist();
 	return text.substr(text.match(artist).index + artist.length + 3)
 		.replace(/^\*[^*]+\*/, '')
-		.replace(/([(*]|- )?buy \d+ get \d+ .*$/i, '')
-		.trim();
+		.replace(/([(*]|- )?buy \d+ get \d+ .*$/i, '');
 }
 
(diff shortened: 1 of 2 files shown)


src/connectors/soundclick.js
```diff
@@ -0,0 +1,38 @@
...
+Connector.playerSelector = '#playerWrap';
+
+Connector.artistSelector = '#playerSCLK_artistName';
+
+// use playlist to get track title
+// to avoid missing content when quotation marks present
+Connector.getTrack = () => {
+	const text = Util.getTextFromSelectors('.songActive .playlist-songtitle');
+	const artist = Connector.getArtist();
+	return text.substr(text.match(artist).index + artist.length + 3);
```
alexesprit, line 16: Can we use `.songActive .playlist-songtitle` as an `artistTrackSelector` instead? What do you think?

Alternatively, you could put removing an artist name to `cleanupTrack` function.

src/connectors/soundclick.js
```diff
@@ -0,0 +1,38 @@
...
+};
+Connector.applyFilter(filter);
+
+function cleanupTrack(text) {
+	return text.replace(/^\*[^*]+\*/, '').replace(/([(*]|- )?buy \d+ get \d+ .*$/i, '').trim();
```
alexesprit, line 21: Could you remove `trim` call? It's already covered by the default filter.

Issue #2417, opened 2020-05-22
Correct unique ID is not generated in Free Music Archive connector


eih:
**Describe the bug**
Free Music Archive uses numeric IDs for tracks but in [src/connectors/freemusicarchive.js](../tree/master/src/connectors/freemusicarchive.js#L64), a certain class name (`tid-XXXXX`) for the track is returned as the unique ID. The `tid-` bit should be stripped off the returned value for a correct ID.

**How to reproduce**
1. Go to a track page, or to [Süper FM Dub](https://freemusicarchive.org/music/Hayvanlar_Alemi/Visions_of_a_Psychedelic_Ankara/04_Sper_FM_Dub) as an example.
2. Inspect the track row in the table. You can see that the id is appended to `tid-` to use it as a class name:
```
<div class="play-item gcol gid-electronic tid-25638" style="height: 51px;">
```
When you inspect the embed code for the track, you can see that `tid-` is not part of the unique ID but an addition to use id value as a class name:
```
<object width="300" height="50"><param name="movie" value="/swf/trackplayer.swf"/><param name="flashvars" value="track=/services/playlists/embed/track/25638.xml"/><param name="allowscriptaccess" value="sameDomain"/><embed type="application/x-shockwave-flash" src="/swf/trackplayer.swf" width="300" height="50" flashvars="track=/services/playlists/embed/track/25638.xml" allowscriptaccess="sameDomain" /></object>
```

**Expected behavior**
This is not an error but for the sake of correct data, unique ID should be returned without `tid-`.

**Environment (please complete the following information)**
-   OS: macOS 10.14.6
-   Browser: Firefox 76.0.1 (64-bit)
-   Extension version: 2.31.2

I have fixed the issue and will create a PR shortly along with some other improvements.

alexesprit:
Not sure if it's a bug, but changing ID format may lead to other issues like #2403.

eih:
I think this is not a bug but more like a maintenance but I see your point. I thought it would be nice to keep the returned value inline with the FMA's unique ID schema.

Issue #2416, opened 2020-05-22
Free Music Archive genre page is not detected if the genre doesn't have subgenres


eih:
**Describe the bug**
At the moment the genre page detection depends on a `.subgenres` class name, which is used on a `div` that wraps the subgenres of a genre. See the following lines in [src/connectors/freemusicarchive.js](../tree/master/src/connectors/freemusicarchive.js#L7)
```JS
const pageSelectors = {
	artist: '.bcrumb h1 .minitag-artist',
	charts: '.page-charts',
	genres: '.subgenres',
	album: '.bcrumb h1 .minitag-album',
	song: '.bcrumb h1 .minitag-song',
};
```
However, this relies on the fact that the genre has subgenres, which isn't the case when a user browses to the deepest subgenre. Due to the lack of that classname on those pages, song detection doesn't work.

**How to reproduce**
Steps to reproduce the behavior. I will provide specific paginated page links for you to easily detect the problem:
1. Go to [Metal genre](https://freemusicarchive.org/genre/Metal?sort=track_date_published&d=1&page=2) page 2.
2. Click play for any track by Cryosyncopy. It should be scrobbled as expected.
3. Go to [Black Metal](https://freemusicarchive.org/genre/Black-Metal/) genre page, which is a subgenre of Metal. But it doesn't have any subgenres.
4. Click on the same song, you'll see that it won't scrobble properly.

**Expected behavior**
It should scrobble the played song regardless of the genre page.

**Debug logs**
<details>
  <summary>Debug log for Genre page with subgenres</summary>
  ```
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 2 webrequest.js:112
Injecting vendor/filter.js inject.js:52:12
Injecting vendor/jquery.min.js inject.js:52:12
Injecting vendor/browser-polyfill.min.js inject.js:52:12
Injecting core/content/util.js inject.js:52:12
Injecting core/content/reactor.js inject.js:52:12
Injecting core/content/connector.js inject.js:52:12
Injecting connectors/freemusicarchive.js inject.js:52:12
Injecting core/content/starter.js inject.js:52:12
Tab 109: Remove controller for Free Music Archive connector util.js:40:10
Tab 109: Created controller for Free Music Archive connector util.js:40:10
Tab 109: New song detected: {
  "track": "Aurora of Evil remix",
  "artist": "Cryosyncopy",
  "album": "Funeral Void",
  "albumArtist": null,
[...]

PR #2419, opened 2020-05-22
Fix genre page issue (#2416) and unique ID issue (#2417) for Free Music Archive connector

eih:
As described in issues #2416 and #2417 Free Music Archive connector has some issues regarding the genre pages without a subgenre and the returned unique ID value. This PR intends to fix those issues.

#2416:
At the moment the genre page detection depends on a `.subgenres` class name, which is used on a `div` that wraps the subgenres of a genre. However, this relies on the fact that the genre has subgenres, which isn't the case when a user browses to the deepest subgenre or a genre without any subgenres. Due to the lack of that classname on those pages, song detection doesn't work.

#2417:
Free Music Archive uses numeric IDs for tracks but in the connector, a certain class name (`tid-XXXXX`) for the track is returned as the unique ID. The `tid-` bit should be stripped off the returned value for a correct ID.

**Additional context**
For more details and logs please the issue tickets.

### DIFF
--- a/src/connectors/freemusicarchive.js
+++ b/src/connectors/freemusicarchive.js
@@ -59,10 +59,9 @@ function setupCommonProps() {
 			return null;
 		}
 
-		const match = /(tid-\d+)/.exec(playingItem.className);
+		const match = /tid-(\d+)/.exec(playingItem.className);
 		if (match) {
-			const idString = match[0].split('-');
-			return idString[1];
+			return match[1];
 		}
 		return null;
 	};
--- a/src/connectors/freemusicarchive.js
+++ b/src/connectors/freemusicarchive.js
@@ -61,7 +61,8 @@ function setupCommonProps() {
 
 		const match = /(tid-\d+)/.exec(playingItem.className);
 		if (match) {
-			return match[0];
+			const idString = match[0].split('-');
+			return idString[1];
 		}
 		return null;
 	};
--- a/src/connectors/freemusicarchive.js
+++ b/src/connectors/freemusicarchive.js
@@ -2,7 +2,7 @@
 
 const filter = new MetadataFilter({ track: removeQuotes });
 
-const trackSelectorCommon = '.gcol-electronic .playtxt a';
+const trackSelectorCommon = '.play-item.gcol-electronic .playtxt a';
 
 const pageSelectors = {
 	artist: '.bcrumb h1 .minitag-artist',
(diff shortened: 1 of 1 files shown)


eih:
Label pull requests / Add labels (pull_request) check is not successful. Do you think it's because of the backticks in the commit message? I haven't had any problems with them on any git repository until now. Besides, some git service providers process it as valid markdown :\

src/connectors/freemusicarchive.js
```diff
@@ -54,14 +54,15 @@ function setupCommonProps() {
...
-		const playingItem = document.querySelector('.gcol-electronic');
+		const playingItem = document.querySelector('.play-item.gcol-electronic');
 		if (!playingItem) {
 			return null;
 		}
 
 		const match = /(tid-\d+)/.exec(playingItem.className);
 		if (match) {
-			return match[0];
+			const idString = match[0].split('-');
```
inverse, line 64: Is this change even required? 

> The identifier does not  have to be in any specific format. The uniqueness is only needed within  the scope of the connector

alexesprit:
`action-labeler` is failing because of this PR is from a fork. I'll remove the workflow file, as it's pretty unused for this project.

src/connectors/freemusicarchive.js
eih: I think we need to provide the correct value for the connector in case we need it anytime in the future. FMA uses numeric IDs so the returned value should be numeric ID. I think the reason behind the prefixing is the naming rule for [CSS selectors](https://www.w3.org/TR/CSS22/syndata.html#characters):
> they cannot start with a digit, or a hyphen followed by a digit.

src/connectors/freemusicarchive.js
```diff
@@ -54,14 +54,15 @@ function setupCommonProps() {
 	Connector.pauseButtonSelector = '.playbtn-paused';
 
 	Connector.getUniqueID = () => {
-		const playingItem = document.querySelector('.gcol-electronic');
+		const playingItem = document.querySelector('.play-item.gcol-electronic');
 		if (!playingItem) {
 			return null;
 		}
 
 		const match = /(tid-\d+)/.exec(playingItem.className);
```
alexesprit, line 62: If you really want to extract a numeric part, you can wrap `\d+` into a group.

eih: Will try that and update.

src/connectors/freemusicarchive.js
alexesprit: It's not required, but omitting `tid` prefix would make the connector more robust and consistent in cases where unique IDs will have no prefix, or other prefixed.

The only drawback is loosing edited tracks, but I'm not sure if this a real issue. FMA connector was broken for a while, but no one reported.

Issue #1492, opened 2017-11-16
Connector Request: Napster


forenta:
After searching around I can't find any scrobbler that supports the Napster streaming service, formerly known as 'Rhapsody'. The Chrome-extensions 'scroblr' already has the support - as I'm back on Firefox, I'd love to keep on scrobbling.

Here's the [source ](https://github.com/cgravolet/scroblr/blob/82cee1013b3ec545a56bacf3b388879ae34ded7e/src/js/plugins/rhapsody.js)they use - maybe it'll help you to figure out how to implement. Although it's still called 'Rhapsody' there, it is working with Napster right now.

forenta:
If someone would like to code the connector, but is not able to do so because she or he does not have an account for Napster - feel free to send me a message. The access is possible from everywhere (credit card-related), so I would lend my account for two days.

inverse:
Doesn't work in Germany :(

forenta:
Sure it does, I'm just sitting in no man's land of NRW.

inverse:
I was on the US page and it threw an error.  Should be able to make a connector based on that other one you provided.

PR #2433, opened 2020-06-02
Add Napster connector

peacegiverman:
- Works with the web player at app.napster.com

Resolves: #1492


alexesprit:
Hi! Thanks for contributing!

Can't verify, as I have no access to Napster, so I believe it works fine. The connector code looks fine.

PR #2439, opened 2020-06-05
Add support for abc.com.au

### LINKED ISSUE #2436: Add Triple J to list of supported websites
Hi,

I recently started using Last.fm and in order to start scrobbling as much music as possible, I added the Web Scrobbler extension to my browser. Because I listen to internet radio quite a lot, I would really appreciate it if you could add support for Triple J, specifically the Triple J Unearthed Jukebox website.

Thanks!

inverse:
Resolves #2436

Supports all 3 players on the page.


alexesprit:
Could you not use jQuery?

src/core/connectors.js
```diff
@@ -1497,6 +1497,13 @@ const connectors = [{
 	],
 	js: 'connectors/napster.js',
 	id: 'napster',
+}, {
+	label: 'triplej',
+	matches: [
+		'*://www.abc.net.au/triplej/listen-live/*',
```
alexesprit, line 1503: We could make a more generic connector that covers all radio stations from ABC.

inverse:
When did you deprecate support and why?

alexesprit:
> When did you deprecate support and why?

https://github.com/web-scrobbler/web-scrobbler/wiki/Connector-API-changes#v2310

It was mostly used along with Bootstrap, but jQuery will be removed in Bootstrap 5. So, there's no sense to keep a dependency, when we can do everything with DOM API.

Using jQuery in connectors to get a text from an element is overkill.

src/connectors/triplej.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+Connector.playerSelector = '#main-content';
+
+Connector.getArtist = () => {
+	return $('.playingNow div[class*="songArtist"]').text() || $('.view-liveMusicNow .artist').text();
```
alexesprit, line 6: This can be replaced by an array of selectors.

inverse:
Why reducing the number of 3rd party dependencies can be a good thing I think leveraging something as well known as jQuery aids the barrier to entry for people to contribute. Although I know vanilla JS has come on a long way.

src/core/connectors.js
inverse: Which other ones did you find?

inverse: Ahhh i see what you mean. I was accessing them all via `/triplej/` but you can access them directly

src/connectors/triplej.js
inverse: Got an example where we do similar?

inverse: Figured

inverse:
What's wierd is the triplej player now for me isnt' expanding with track info

image

The other two are just fine - but they have different selectors

src/connectors/abcnetau.js
```diff
@@ -0,0 +1,19 @@
...
+Connector.playerSelector = '#main-content';
+
+Connector.artistSelector = '.playingNow div[class*="songArtist"], .view-liveMusicNow .artist';
+
+Connector.trackSelector = '.playingNow h3, .view-liveMusicNow .title';
+
+Connector.albumSelector = '.playingNow div[class*="songRelease"], .view-liveMusicNow .release';
+
+Connector.isPlaying = () => {
+	const $playerElement = document.querySelector('#jwplayerDiv, #radio-player4-player');
```
alexesprit, line 12: You can use `Util.hasElementClass` instead.

inverse: Updated

PR #2444, opened 2020-06-08
Added JQBX connector

daniel-stoneuk:
Added a connector for JQBX.

For use at *://app.jqbx.com

Let me know if you would rather the connector to be named without the app lower level domain.


alexesprit:
> Let me know if you would rather the connector to be named without the app lower level domain.

If there are no other subdomains on jqbx.com, the `app` suffix can be omitted.

daniel-stoneuk:
Brilliant - wasn't too sure. Thanks.

daniel-stoneuk:
Thanks very much!

Issue #2429, opened 2020-05-29
Remove Google Analytics


alexesprit:
I find data tracking pretty useless for me, so removing this feature could be good for me (I don't need to maintain GA related code) and for users (their data won't be tracked by the extension).

Ref: https://gist.github.com/alexesprit/3e3361550da08273e7b6d9f5e9c5d8ce.

## Tasks to do

- [x] Remove GA-related code
- [x] Update privacy policy
- [ ] Translate privacy policy

alexesprit:
> How else can we see how well adopted the extension is through the marketplaces without?

GA does not give you this information.

> I mean it's opt in on install now?

Now tracking data is enabled by default, and can be disabled either via startup page after the extension installation, or via the options page later.

inverse:
> GA does not give you this information.

Yes it does - it shows you how many current users are using the extension

> Now tracking data is enabled by default, and can be disabled either via startup page after the extension installation, or via the options page later.

You mean for current installs? 

I think having some form of telemetry is a good thing to be able to best shape the product. e.g. which connectors people use, interactions etc. How else will you know these patterns without telemetry?

alexesprit:
Like I said, I see no benefits of using telemetry.

Now the extension collects connector injects, and scrobble events. It also uses a `pageview` event, so we can view how many users are using the extension right now (not sure if it's a correct number, btw). *How* do you use this data?

> I think having some form of telemetry is a good thing to be able to best shape the product. e.g. which connectors people use, interactions etc. How else will you know these patterns without telemetry?

You're talking about what data is collected, but how do suggest to use it?

I'm not against keeping GA, but when there's no sense to use it, it's better to get rid of collecting data.

peacegiverman:
Had this done locally some time ago, so I just submited a pull requested.

inverse:
> You're talking about what data is collected, but how do suggest to use it?

I'll try to take a look when I have some time but I think It'll be interesting to get metrics on things like how much usage each connector is getting or per browser breakdowns etc.

jbwharris:
Have Analytics been straight up removed from Web Scrobbler? I am definitely curious as to data about which are the most popular connectors and inversely which aren't being used. Having done work on the SiriusXM connector recently I'm curious how many people use it regularly. I think it is used a fair bit, given the last.fm data, but there's no dates or anything tied to that historical data, just how many users might have mis-scrobbled a station tag. 

** Cases data would be interesting
[...]

PR #2449, opened 2020-06-14
WIP: Remove Google Analytics #2429

peacegiverman:
Had Google Analytics already removed in my local fork when I saw that this issue was open, so I just cleaned it up and removed all the GA related messages.

There is still one issue before this can be merged into master. What do we do with the privacy policy and the startup screen? Should we keep a privacy policy stub, or remove it altogether?

#2429


alexesprit:
I have no intentions to remove GA right now. It's supposed to be removed in v3.0.0.

> There is still one issue before this can be merged into master. What do we do with the privacy policy and the startup screen? Should we keep a privacy policy stub, or remove it altogether?

It will require removing the startup page, and updating the privacy policy. I added a link to an updated document in the issue.

alexesprit:
Ref https://github.com/web-scrobbler/web-scrobbler/tree/feature/remove-analytics.

PR #2477, opened 2020-07-10
Fixed the non-working NRK connector

areee:
I fixed the non-woking NRK connector.

**Additional context**
Näyttökuva 2020-7-10 kello 19 56 14

### DIFF
--- a/src/connectors/nrk-radio.js
+++ b/src/connectors/nrk-radio.js
@@ -2,6 +2,6 @@
 
 Connector.playerSelector = '.app';
 
-Connector.artistTrackSelector = '.css-6752n5.e79qwfn0';
+Connector.artistTrackSelector = '[data-test="playingEpisodeDesc"]';
 
 Connector.pauseButtonSelector = '.css-1o5s8kn.e79qwfn0';
--- a/src/connectors/nrk-radio.js
+++ b/src/connectors/nrk-radio.js
@@ -13,5 +13,5 @@ Connector.pauseButtonSelector = '[data-test="playPauseMiniplayer"][aria-label="P
 Connector.applyFilter(filter);
 
 function cleanUpArtist(artist) {
-	return artist.replace('+', '&');
+	return artist.replace(/\s\+\s/g, ' & ');
 }
--- a/src/connectors/nrk-radio.js
+++ b/src/connectors/nrk-radio.js
@@ -4,4 +4,4 @@ Connector.playerSelector = '.app';
 
 Connector.artistTrackSelector = '[data-test="playingEpisodeDesc"]';
 
-Connector.pauseButtonSelector = '.css-1o5s8kn.e79qwfn0';
+Connector.pauseButtonSelector = '[data-test="playPauseMiniplayer"][aria-label="Pause"]';
--- a/src/connectors/nrk-radio.js
+++ b/src/connectors/nrk-radio.js
@@ -1,7 +1,17 @@
 'use strict';
 
+const filter = new MetadataFilter({
+    artist: cleanUpArtist,
+});
+
 Connector.playerSelector = '.app';
 
 Connector.artistTrackSelector = '[data-test="playingEpisodeDesc"]';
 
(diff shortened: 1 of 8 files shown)


src/connectors/nrk-radio.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
-Connector.playerSelector = '.mini-player';
+Connector.playerSelector = '.app';
 
-Connector.artistTrackSelector = '.currently-playing .episode-title';
+Connector.artistTrackSelector = '.css-6752n5.e79qwfn0';
```
inverse, line 5: This looks very brittle - like a new release of their site will cause things to break. Could we not use the `data-test="playingEpisodeDesc"` attribute to resolve this information?

src/connectors/nrk-radio.js
```diff
@@ -1,7 +1,7 @@
...
 
-Connector.pauseButtonSelector = '.paused';
+Connector.pauseButtonSelector = '.css-1o5s8kn.e79qwfn0';
```
inverse, line 7: Same with this one 

`data-test="playPauseMiniplayerMobile"` when `aria-label="Pause"`

areee:
Thanks for your comments! I definitely agree! I'll try to enhance the selectors to more meaningful ones. :)

areee:
I recently fixed those selectors to be less brittle. I hope they are what you were thinking about. :)

Besides, I added a filter to replace the '+' character with the '&' character. The NRK player seems to use '+' char when songs have multiple artists and that's why web-scrobbler does not recognize the played song.

src/connectors/nrk-radio.js
```diff
@@ -1,7 +1,17 @@
...
 
-Connector.pauseButtonSelector = '.paused';
+Connector.artistTrackSelector = '[data-test="playingEpisodeDesc"]';
+
+Connector.pauseButtonSelector = '[data-test="playPauseMiniplayer"][aria-label="Pause"]';
+
+Connector.applyFilter(filter);
+
+function cleanUpArtist(artist) {
+	return artist.replace('+', '&');
```
alexesprit, line 16: `String.replace` replaces a given substring one time only. So, if you have a "K-391 + Alan Walker + Julie Bergan + Seungri" string as an artist name, the output string will be "K-391 & Alan Walker + Julie Bergan + Seungri". This should be changed to keep consistent artist format.

You can use regular expressions to replace all matches.

alexesprit:
Sorry for a delay.

The changes look good for me. I added a small change request, though. :)

src/connectors/nrk-radio.js
inverse: Thanks for explaining - I totally didn't realise that replace when passed string would only replace the first instance. Is that the same when passing regex without global flag? I guess so.

docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

alexesprit: Yes, it's the same for expressions w/o a global flag.

https://jsfiddle.net/zmq548ov/

alexesprit: Also, I think we could include spaces to make the filter even more robust, and to avoid unwanted replaces.

areee: Hi! Thank you both for the information! I'll try to fix this by using the regex. :)

areee:
Thanks for the regular expression tips, they were very helpful! 👍 Now I have made changes to the replacement filter feature. Is it what you were thinking about? :)

alexesprit:
> Is it what you were thinking about? :)

Yes, it's perfect now. Thank your for contribution!

hbhc:
I don't understand why scrobbling radio,nrk.no doesn't work for me at all. Any tips?

areee:
The NRK connector is broken in the latest release (2.33.0) but it has already been fixed in the source code. You can install the extension from the source code or wait till the next release.

alexesprit:
I'll publish a new release in few hours, it's better to wait for it.

hbhc:
Cool. Thanks!

PR #2478, opened 2020-07-10
Developed connector for Supla

areee:
I added connector for Supla that is a Finnish service combining some Finnish radio stations, podcasts and audio books.

**Additional context**
You can test Supla e.g. by listening [Loop](https://www.supla.fi/loop) radio station.
Näyttökuva 2020-7-10 kello 20 01 08

### DIFF
--- a/src/connectors/supla.js
+++ b/src/connectors/supla.js
@@ -1,7 +1,7 @@
 'use strict';
 
-Connector.playerSelector = '#ComponentContainer__Section-sc-1m3wxgx-0';
+Connector.playerSelector = '#RadioHeader__Container-sc-17ofob1-0 iLWJvB';
 
-Connector.artistTrackSelector = '#RadioHeader__WidgetContent-sc-17ofob1-4';
+Connector.artistTrackSelector = '#RadioHeader__WidgetContent-sc-17ofob1-4 irUZCR';
 
-Connector.playButtonSelector = '#RadioHeader__WidgetContentAction-sc-17ofob1-9';
+Connector.playButtonSelector = '#RadioHeader__WidgetContentAction-sc-17ofob1-9 kKpbyx';
--- a/src/connectors/supla.js
+++ b/src/connectors/supla.js
@@ -8,7 +8,7 @@ Connector.playerSelector = 'main';
 
 Connector.artistTrackSelector = '.RadioHeader__WidgetContent-sc-17ofob1-4.irUZCR';
 
-Connector.playButtonSelector = '.r-play-button.r-toggle-playstate-button.r-touch-button';
+Connector.playButtonSelector = '.r-play-button';
 
 Connector.applyFilter(filter);
 
--- a/src/connectors/supla.js
+++ b/src/connectors/supla.js
@@ -6,7 +6,9 @@ const filter = new MetadataFilter({
 
 Connector.playerSelector = 'main';
 
-Connector.artistTrackSelector = '.RadioHeader__WidgetContent-sc-17ofob1-4.irUZCR';
+//Connector.artistTrackSelector = '[class^=RadioHeader__WidgetContent]'; // works with radio channels without the program title ('Groove FM' & 'Anna sen soida')
+Connector.artistTrackSelector = 'main > div > div > section > div > div:nth-child(2) > div:nth-child(2) > div'; // works with radio channels with the program title (all the rest except the earlier ones)
+
 
 Connector.playButtonSelector = '.r-play-button';
 
--- a/src/connectors/supla.js
+++ b/src/connectors/supla.js
@@ -4,7 +4,7 @@ const filter = new MetadataFilter({
 	artist: cleanUpArtist,
 });
 
-Connector.playerSelector = '.Main__MainElement-sc-1f3ou2j-0.dwVNSG';
+Connector.playerSelector = 'main';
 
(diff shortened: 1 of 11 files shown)


areee:
As in PR #2477, I'm still trying to avoid the 'brittle code' with the `Connector.artistTrackSelector` as its selector isn't currently the best one. Does anyone know if it's possible to use XPath or something else to select the current song name by using the _soittohistoria_ ("song history") title?

src/connectors/supla.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const filter = new MetadataFilter({
+	artist: cleanUpArtist,
+});
+
+Connector.playerSelector = 'main';
+
+Connector.artistTrackSelector = '.RadioHeader__WidgetContent-sc-17ofob1-4.irUZCR';
```
alexesprit, line 9: I would go with a way we use in Plex connector.

https://github.com/web-scrobbler/web-scrobbler/blob/3db105c7508688bb88b5184939b09032a18854b9/src/connectors/plex.js#L9

For this particular property the selector could be like `[class^=RadioHeader__WidgetContent]`. The `irUZCR` should be omitted.

src/connectors/supla.js
```diff
@@ -0,0 +1,18 @@
...
+
+Connector.playButtonSelector = '.r-play-button.r-toggle-playstate-button.r-touch-button';
```
alexesprit, line 11: It's better to avoid multiple classes in a selector. Using `.r-play-button` only is enough.

alexesprit:
> Does anyone know if it's possible to use XPath or something else to select the current song name by using the soittohistoria ("song history") title?

We have no ways to get elements via XPath.

src/connectors/supla.js
areee: Unfortunately `[class^=RadioHeader__WidgetContent]` doesn't work because it selects the program title of the radio channel. E.g. in the [Loop](https://www.supla.fi/loop) radio station page, the Web Scrobbler selects artist and track names from the title _Nyt: Loop iltapäivä - Paula "Paukkki" Rinta-Kanto_.

See an example photo below:
kuva

areee: However, your tip works with the page that doesn't have a program title. Weird...
E.g. see the [Groove FM](https://www.supla.fi/groovefm) radio channel page:
Näyttökuva 2020-7-13 kello 19 12 27

areee:
Hi! Thanks for the information! Could you help me one more time with the `Connector.artistTrackSelector` problem? :)

With Selenium IDE, I managed to get the path to the song title element. However, there are two page types in Supla radio channels: pages
1) with the program title and
2) without the program title.

As I commented to the file, the code
`Connector.artistTrackSelector = 'main > div > div > section > div > div:nth-child(2) > div:nth-child(2) > div';`
works with nearly all radio channels **except** with [Groove FM](https://www.supla.fi/groovefm) and [Anna sen soida](https://www.supla.fi/annasensoida) radio channels.

It seems that Web Scrobbler support different types of pages, doesn't it? E.g. I found that [radioplayer.js](https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/radioplayer.js#L10) has two priorities to get the track and artist informations.

src/connectors/supla.js
alexesprit: Yes, the element matching the `[class^=RadioHeader__WidgetContent]` is not unique, so this suggestion won't work.

alexesprit:
> It seems that Web Scrobbler supports different types of pages, doesn't it?

I would not say it supports different types of pages, but it provides some ways to write connectors for different types of pages/players.

> As I commented to the file, the code
Connector.artistTrackSelector = 'main > div > div > section > div > div:nth-child(2) > div:nth-child(2) > div';
works with nearly all radio channels except with Groove FM and Anna sen soida radio channels.

You can use an array of selectors in this case. But this selector will be invalid if website layout will be updated.

One of solutions is to use `.irUZCR` as an `artistTrackSelector`. This is not a perfect option, though, since this class is generated, and could be changed. But it's still acceptable.

Another solution is to find a node with track info and use it to get an artist name and a track title. See the connector code below for details. Feel free to use this code in your PR.

<details>
  <summary>supla.js (Click to expand)</summary>

```js
'use strict';

Connector.playerSelector = 'main';

Connector.playButtonSelector = '.r-play-button';

	const artistTrackContainer = getArtistTrackContainer();
	if (artistTrackContainer) {
		return Util.splitArtistTrack(artistTrackContainer.textContent);
	}

	return null;
};

function getArtistTrackContainer() {
	/*
	 * <div class="RadioHeader__WidgetContent-sc-17ofob1-4 irUZCR">
	 *   <i class="MaterialIcon__Icon-sc-1m3ozx6-0 kLFDRs">music_note</i>
	 *   Artist - Track
	 * </div>
	 */

	const containers = document.querySelectorAll(
		'[class^=RadioHeader__WidgetContent]'
	);

	for (const container of containers) {
		for (const child of container.childNodes) {
			if (child.textContent === 'music_note') {
				return container.lastChild;
			}
		}
	}

	return null;
}
```
</details>

areee:
Thank you for your help, now the Supla connector is working like a charm! 🙌 :)
If I have time later, I'd like to develop connectors also to the other Finnish radio station services. I like the idea of this project. 👍

alexesprit:
Thank your for contributing to the project!

Issue #2479, opened 2020-07-11
Emby/Jellyfin does not work on Jellyfin.


oldshensheep:
Emby/Jellyfin does not work on Jellyfin.
The song name is empty
![0.png](https://i.loli.net/2020/07/11/Pd2RkwfJZF3DBlN.png)
Here is the PlayingBar
![1.png](https://i.loli.net/2020/07/11/6NbBS5723a9p4Oz.png)

**Environment**
-   Browser: Firefox and edge
-   Extension version: v2.33.0
-  Jellyfin v10.5.5

alexesprit:
I can't reproduce the issue on Jellyfin v10.5.5 and MS Edge/Firefox.

Could you provide debug logs? [How to debug the extension.][1]

[1]: https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension

![image](https://user-images.githubusercontent.com/1119267/87446496-08514700-c602-11ea-8159-8d00c2529cb8.png)

PR #2483, opened 2020-07-15
Fixed the non-working emby connector. issue (#2479)

oldshensheep:
I did some debugging, I found that when "data-type" is "MusicAlbum", the song name can be correctly identified. Look at the picture below.

It cannot be recognized when "data-type" is "Audio".

I parsed it using another method. I have compiled and tested on jellyfin on MS Edge browser. It works fine.


alexesprit:
Good catch!

I'll check the changes, and will merge the PR if everything is OK. :)

alexesprit:
Awesome, thanks!

PR #2486, opened 2020-07-17
gds.fm launched new page

heyarne:
gds.fm moved from play.gds.fm to gds.fm and slightly changed the page layout. I updated the selectors and URL so it roughly works again. I have a question about I encountered though: `.player-meta__title`, which I chose as `Connector.trackSelector` only contains the track title. The track artist is in a sibling element… however on track change the artist is always appended to the title, with the artist also being detected independently. Any idea why that happens? Thanks for the addon, it's fantastic!


inverse:
Thanks for taking the time to fix an integration.  Can you give an example URL where this is working? 

When I visit https://gds.fm/ and hit play nothing seems to get detected for me.

alexesprit:
> When I visit https://gds.fm/ and hit play nothing seems to get detected for me.

It's missing the `Connector.playerSelector` property.

Here's a working example:
```
'use strict';

Connector.playerSelector = '.page-header__logo-player';

Connector.artistSelector = '.player-meta__artist';

Connector.trackSelector = '.player-meta__title';

Connector.playButtonSelector = '.player-meta__icon--play';
```

heyarne:
Maybe there's a misunderstanding on my side; I thought `Connector.playerSelector` is an optimization that I left out to make it easier to figure out why things are detected (or why they aren't).

I have added comments to the different selectors (I took out the overwritten callback, so this is the entire `src/connectors/gds.fm.js`):
```
'use strict';

// container wrapping all of the following selectors; this is done so the whole page doesn't have to be observed for mutations
Connector.playerSelector = '.player-meta'

// element containing *only* the artist name
Connector.artistSelector = '.player-meta__artist';

// element containing *only* the current track title
Connector.trackSelector = '.player-meta__title';

// if this element is found, nothing is playing; the song should be registered as playing by web-scrobbler as soon as this element disappears
// this is the opposite of Connector.pauseButtonSelector
Connector.playButtonSelector = '.player-meta__icon--play';
```

I have double-checked the selectors, they are correct if my understanding (that I made explicit in the comments) is correct. What happens with the above code however (I'm debugging this on Firefox 78) is:

- When opening the page, no music is playing (i.e. the element `.player-meta__icon--play` is visible). Still, the track is immediately registered by the web-scrobbler extension
- The parsed track contains the correct artist, but the track title looks like '$track $artist', which leads to the track not being found
- When the track changes, the new track is parsed in exactly the same way.

So I'm pretty sure I'm misunderstanding something, even though the comments in the BaseConnector source code seem pretty clear.

alexesprit:
Thank you!

PR #2503, opened 2020-08-02
Add rekt.network and nighride.fm connectors.

lebbe:
Added connectors to the radios on rekt.network and nightride.fm websites.


alexesprit:
Please use separate PRs for each connector.

lebbe:
I added a PR for the individual radio stations just now.

PR #2507, opened 2020-08-03
A connector for the rekt.network online radio.

lebbe:
This adds a connector for the rekt.network radio, which has a website on https://rekt.network/?station=rekt

**Additional context**
The rekt.network radio have regular live sendings with DJs and show hosts, and have an endless stream of music. All music are submitted to the radio by artists, or collected by other non-copyright infringing means.

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -1554,7 +1554,7 @@ const connectors = [{
 	js: 'connectors/epidemicsound.js',
 	id: 'epidemicsound',
 }, {
-	label: 'REKT.NETWORK',
+	label: 'Rekt Network',
 	matches: [
 		'*://rekt.network/*',
 	],
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -1553,6 +1553,13 @@ const connectors = [{
 	],
 	js: 'connectors/epidemicsound.js',
 	id: 'epidemicsound',
+}, {
+	label: 'REKT.NETWORK',
+	matches: [
+		'*://rekt.network/*',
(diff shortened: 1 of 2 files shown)


lebbe:
BTW!

I have personally tested this implementation both in Firefox and Chrome, and it started and stopped scrobbling on last.fm without problems. :)

alexesprit:
@lebbe Thank you for contribution!

Do you know these two websites have a common owner? I see the connectors for these websites are the same, and we could reuse a connector instead of copypasting the code.

src/core/connectors.js
```diff
@@ -1553,6 +1553,13 @@ const connectors = [{
 	],
 	js: 'connectors/epidemicsound.js',
 	id: 'epidemicsound',
+}, {
+	label: 'REKT.NETWORK',
```
alexesprit, line 1557: Could you change the label to "Rekt Network"? They have this name in Twitter and in their website (but it's changed to "Rekt FM" after the website is loaded, but I think it's a bug).

lebbe:
@alexesprit I had a common PR for them both, but was asked to split PR in two. :) It is same owner/provider, but implementations might change in the future though.

alexesprit:
@lebbe sorry, I didn't notice they are the same, when I asked to split that PR.

I think it's fine to leave them as is for now. The PR looks good aside the minor comment.

alexesprit:
Perfecto! Merging the PR right now.

lebbe:
Awesome thank you! :)

PR #2523, opened 2020-08-13
Refactor stringToSeconds with simple regex

russellsanborn:
Refactored Util.stringToSeconds by using two simple regex matchers. There is also no longer a need to trim the string, since the expressions only match on the negative sign or groups of two digits.

**Additional context**
I also a updated the tests to include leading whitespace and added a test with whitespace around a negative time.

### DIFF
--- a/src/core/content/util.js
+++ b/src/core/content/util.js
@@ -31,34 +31,15 @@ const Util = {
 		if (!str) {
 			return 0;
 		}
+		const negativeExpression = /-/g;
+		const digitsExpression = /\d\d/g;
 
-		let s = str.toString().trim();
-		let val = 0;
-		let seconds = 0;
+		const seconds = str.match(digitsExpression)
+			.reverse()
+			.map((current) => parseInt(current, 10))
+			.reduce((total, current, i) => total + current * Math.pow(60, i));
 
-		const isNegative = s.startsWith('-');
-		if (isNegative) {
-			s = s.substr(1);
-		}
-
-		for (let i = 0; i < 3; i++) {
-			const idx = s.lastIndexOf(':');
-			if (idx > -1) {
-				val = parseInt(s.substr(idx + 1), 10);
-				seconds += val * Math.pow(60, i);
-				s = s.substr(0, idx);
-			} else {
-				val = parseInt(s, 10);
-				seconds += val * Math.pow(60, i);
-				break;
-			}
-		}
-
-		if (isNegative) {
-			seconds = -seconds;
-		}
-
-		return seconds;
+		return (negativeExpression.test(str)) ? -seconds : seconds;
 	},
 
 	/**
--- a/src/core/content/util.js
+++ b/src/core/content/util.js
@@ -28,9 +28,11 @@ const Util = {
 	 * @return {Number} Seconds
 	 */
 	stringToSeconds(str) {
-		if (!str) {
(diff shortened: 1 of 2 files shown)


alexesprit:
Thanks for the contribution!

1. The function now returns a result for strings w/o colons, e.g. `'01 20'`, but it's not a valid time string.
2. It also returns an invalid result for strings with days, e.g. `'01:00:00:00'`.

russellsanborn:
@alexesprit Good call out. I am adding both of those cases you pointed out as test cases and see if i can make it still work and be relatively simple. If not, I will at least add the test cases in a PRs to ensure it will fail if the function is change without addressing them.

russellsanborn:
@alexesprit I added the test cases and check to make sure the `hh:mm:ss`, `mm:ss`, or `ss` pattern exists before parsing the time. If the latest regex I added is getting too complicated, let me know. I will just remove the changes to the stringToSeconds, but keep the tests.

alexesprit:
@russellsanborn it gets too compicated now. But extending tests looks good for me.

alexesprit:
@russellsanborn Like I said in #2526, we can go with this PR, since it brings some benefits.

I found a regression where the function doesn't process strings in `m:ss` or `h:mm:ss` format.

```js
{
	description: 'should parse time in m:ss format',
	args: ['5:20'],
	expected: 320,
}, {
	description: 'should parse time in m:ss format',
	args: ['5:20:00'],
	expected: 19200,
}
```

These cases are failed, but passed when I running tests against the master branch.

russellsanborn:
Does it have to support formats where the single digit value is not the most significant value, like `hh:mm:s`, `h:m:ss`, `h:m:s`, 'mm:s`, or even just `s`? Or is it considered any of the lesser significant values are always 2 digits?

alexesprit:
@russellsanborn the first option is fine, plus, it should be easier to implement.

russellsanborn:
They both seem easy enough to me and I am torn, mainly because I do not have the history and knowledge with this project like you do. Since you understand the possible time values that this method receives, I want to leverage your advice.

**Option 1**
```javascript
// I decided it was easier to read if I combined the three different match patterns
// I have a friend who says if you use regex to solve a problem, you now have two problems
const timeFormatExpression = /^\s*-?((\d{1,2}:\d\d:\d\d)|(\d{1,2}:\d\d)|(\d{1,2}))\s*$/g;
```
Handles
* `s` and `ss`
* `m:ss` and `mm:ss`
* `h:mm:ss` and `hh:mm:ss`

Option 2 (seems unrealistic for any of time strings i have ever seen)
```javascript
const timeFormatExpression = /^\s*-?(\d{1,2}:){0,2}(\d{1,2})\s*$/g;
```
Handles
* `s` and `ss`
* `m:s`, `m:ss`, `mm:s`, and `mm:ss`
* `h:m:s`, `h:m:ss`, `h:mm:s`, `h:mm:ss`, `hh:m:s`, `hh:m:ss`, `hh:mm:s`, and `hh:mm:ss`

alexesprit:
I've never seen something like `mm:s`, but a lot of players have the `m:ss` format. 

The option 1 looks fine for me; we can add more test cases to ensure `mm:s` or `hh:m:s` aren't processed.

> // I have a friend who says if you use regex to solve a problem, you now have two problems

Haha, that's true.

alexesprit:
Thank you!

PR #2528, opened 2020-08-19
Disable individual lint warnings instead of a global disable

russellsanborn:
Instead of globally disabling all lint warnings on a line, only disable the specific lint rule.

**Additional context**
None


alexesprit:
Good catch! Thank you.

PR #2538, opened 2020-08-27
Detektor.fm

carpet-berlin:
Added connector for detektor.fm


alexesprit:
@carpet-berlin Thank you for the PR!

I have only one minor request. Could you add a support for track arts?

alexesprit (approved):

PR #2546, opened 2020-09-01
Add case for uploaded albums to the YT Music album selector.

### LINKED ISSUE #2541: Cannot detect album title for uploaded tracks in YouTube Music
https://github.com/web-scrobbler/web-scrobbler/blob/5c53ae8c8aaf5cb33cbe38d3f3c07151ad093ab3/src/connectors/youtube-music.js#L24

Selector in this line, `href*="browse/MPREb_"` may be like `href*="browse/FEmusic_library_privately_owned_release_detailb_"` in case when playing uploaded tracks.

jaccarmac:
Resolves #2541

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -21,7 +21,7 @@ Connector.getTrackArt = () => {
 	return null;
 };
 
-Connector.albumSelector = '.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/MPREb_"]';
+Connector.albumSelector = '.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/MPREb_"], .ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]';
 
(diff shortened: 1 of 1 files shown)


jaccarmac:
Lint whines at me if I break the selector into two lines with a string concat. How do we want to handle the very long line?

alexesprit:
> Lint whines at me if I break the selector into two lines with a string concat. How do we want to handle the very long line?

1. You can simplify selectors to look like `.ytmusic-player-bar [href*="browse/MPREb_"]` and `.ytmusic-player-bar [href*="browse/FEmusic_"]`. When I was reproducing the issue, I tested these selectors, and they were unique. Although I can miss some edge cases, where these simplified selectors would not work.
2. You can assign an array of selectors (this way is much safer) to a `Connector.albumSelector` property:
```js
Connector.albumSelector = [
    'selector-1',
    'selector-2',
];
```

jaccarmac:
>     1. You can simplify selectors to look like `.ytmusic-player-bar [href*="browse/MPREb_"]` and `.ytmusic-player-bar [href*="browse/FEmusic_"]`.

`browse/FEmusic_` also matches uploaded artists, since those links are to within the private library and not a YouTube channel. However, I can do a partial simplification.

>     2. You can assign an array of selectors (this way is much safer)

Will do, thanks for the correction!

alexesprit (approved):

PR #2548, opened 2020-09-04
Archive.org Drop jQuery and legacy

inverse:

- Drop legacy design (it's been a long while and not supported from what I could see)
- Drop jQuery based selectors

**Pages to test**

- https://archive.org/details/AH013_sarin_sunday_-_the_lonely_hike/01_Light_Fragments.mp3

### DIFF
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -1,18 +1,35 @@
 'use strict';
 
+const artistSelectors = [
+	'.key-val-big a span',
+	'.metadata-definition > dd',
+];
+const trackSelectors = [
+	'.jwrowV2.playing .ttl',
+	'.audio-track-list .selected .track-title',
+];
+
 Connector.currentTimeSelector = '.jw-text-elapsed';
 
 Connector.durationSelector = '.jw-text-duration';
 
-Connector.isPlaying = () => !document.querySelector('video').paused;
+Connector.isPlaying = () => {
+	const videoElement = document.querySelector('video');
+
+	if (videoElement === null) {
+		return false;
+	}
+
+	return !videoElement.paused;
+};
 
 Connector.playerSelector = '#theatre-ia';
 
 Connector.trackArtSelector = '#theatre-ia center > img';
 
 Connector.getArtistTrack = () => {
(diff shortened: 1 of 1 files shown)


src/connectors/archive.js
```diff
@@ -1,89 +1,28 @@
...
 
-	Connector.durationSelector = '.jw-text-duration';
+Connector.trackArtSelector = '#theatre-ia center > img';
 
-	Connector.isPlaying = () => {
-		return !$('video')[0].paused;
-	};
-}
+Connector.getArtistTrack = () => {
+	let track = document.querySelector('.jwrowV2.playing .ttl, .audio-track-list .selected .track-title').textContent;
```
alexesprit, line 14: It's better to use array of selectors, as they are much safer.

```js
const artistSelectors = [
	'.key-val-big a span',
	'.metadata-definition > dd',
];
const trackSelectors = [
	'.jwrowV2.playing .ttl',
	'.audio-track-list .selected .track-title',
];

	let track = Util.getTextFromSelectors(trackSelectors);
	const artist = Util.getTextFromSelectors(artistSelectors);
	...
};
```

src/connectors/archive.js
```diff
@@ -1,89 +1,28 @@
...
-	case 'legacy':
-		bindLegacy();
-		break;
+Connector.durationSelector = '.jw-text-duration';
 
-	default:
-		Util.debugLog('Unable to match version. Please report at https://github.com/web-scrobbler/web-scrobbler/issues', 'warn');
-		break;
-}
+Connector.isPlaying = () => !document.querySelector('video').paused;
```
alexesprit, line 7: `document.querySelector` could return `null`, so I would add an additional check here.

alexesprit (changes requested):

alexesprit:
Thanks! I've already rewritten all of them, but still happy to merge this PR.

src/connectors/archive.js
inverse: I think we should document these Util functions more if we want people to adopt them.

inverse:
@alexesprit no way - on a branch? Isn't that a pain to maintain a long running branch with churn on master?

alexesprit (approved):

PR #2551, opened 2020-09-09
Add RecMusic connector

yayuyokitano:
Adds a connector to grab metadata and scrobble off recmusic.jp

Includes a filter to convert half-width katakana to full-width katakana,
as RecMusic frequently uses half-width katakana where it is incorrect.

Examples of where this filter is necessary:
[This single and first track](https://recmusic.jp/album/?id=1011218626)
Filter correctly converts ﾚｽﾎﾟｰﾙ to レスポール
[This artist name](https://recmusic.jp/artist/?id=2001132142)
Filter correctly converts ｺﾄﾌﾙ to コトフル

This conversion will inevitably lead to some false positives.
However, false positives should be _far_ less common.

### DIFF
--- a/src/connectors/recmusic.js
+++ b/src/connectors/recmusic.js
@@ -18,7 +18,9 @@ Connector.durationSelector = '.seekbar-duration:last-child';
 
 Connector.trackSelector = 'p.track-info-title';
 
-Connector.isPlaying = () => document.querySelector('.button-play').classList.contains('mk-web-icon-stop');
+Connector.pauseButtonSelector = ".button-play.mk-web-icon-stop";
+
+//Connector.isPlaying = () => document.querySelector('.button-play').classList.contains('mk-web-icon-stop');
 
 Connector.applyFilter(filter);
 
--- a/src/connectors/recmusic.js
+++ b/src/connectors/recmusic.js
@@ -6,19 +6,19 @@ const filter = new MetadataFilter({
 
 Connector.playerSelector = '.the-player';
 
-Connector.artistSelector = "a.track-info-artist";
+Connector.artistSelector = 'a.track-info-artist';
 
-Connector.albumSelector = "a.track-info-album";
+Connector.albumSelector = 'a.track-info-album';
 
-Connector.getTrackArt = () => document.querySelector('.track-info-image img').style["background-image"].slice(5, -2);
+Connector.getTrackArt = () => document.querySelector('.track-info-image img').style['background-image'].slice(5, -2);
 
-Connector.currentTimeSelector = ".seekbar-duration:nth-child(2)";
+Connector.currentTimeSelector = '.seekbar-duration:nth-child(2)';
 
-Connector.durationSelector = ".seekbar-duration:last-child";
+Connector.durationSelector = '.seekbar-duration:last-child';
 
-Connector.trackSelector = "p.track-info-title";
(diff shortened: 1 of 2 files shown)


src/connectors/recmusic.js
```diff
@@ -0,0 +1,51 @@
...
+	all: katakanaHanToZen,
+});
+
+Connector.playerSelector = '.the-player';
+
+Connector.artistSelector = 'a.track-info-artist';
+
+Connector.albumSelector = 'a.track-info-album';
+
+Connector.getTrackArt = () => document.querySelector('.track-info-image img').style['background-image'].slice(5, -2);
```
alexesprit, line 13: It's better to use `Util.extractUrlFromCssProperty` function to get a URL from CSS properties.

src/connectors/recmusic.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Replace half-width katakana with full-width katakana in the text.
+ * @param  {String} text String to be filtered
+ * @return {String} Filtered string
+ */
+
+function katakanaHanToZen(text) {
+
+	const katakanaHanToZenMap = { '｡': '。', '｢': '「', '｣': '」', '､': '、', '･': '・', 'ｦ': 'ヲ', 'ｧ': 'ァ', 'ｨ': 'ィ', 'ｩ': 'ゥ', 'ｪ': 'ェ', 'ｫ': 'ォ', 'ｬ': 'ャ', 'ｭ': 'ュ', 'ｮ': 'ョ', 'ｯ': 'ッ', 'ｰ': 'ー', 'ｱ': 'ア', 'ｲ': 'イ', 'ｳ': 'ウ', 'ｴ': 'エ', 'ｵ': 'オ', 'ｶ': 'カ', 'ｷ': 'キ', 'ｸ': 'ク', 'ｹ': 'ケ', 'ｺ': 'コ', 'ｻ': 'サ', 'ｼ': 'シ', 'ｽ': 'ス', 'ｾ': 'セ', 'ｿ': 'ソ', 'ﾀ': 'タ', 'ﾁ': 'チ', 'ﾂ': 'ツ', 'ﾃ': 'テ', 'ﾄ': 'ト', 'ﾅ': 'ナ', 'ﾆ': 'ニ', 'ﾇ': 'ヌ', 'ﾈ': 'ネ', 'ﾉ': 'ノ', 'ﾊ': 'ハ', 'ﾋ': 'ヒ', 'ﾌ': 'フ', 'ﾍ': 'ヘ', 'ﾎ': 'ホ', 'ﾏ': 'マ', 'ﾐ': 'ミ', 'ﾑ': 'ム', 'ﾒ': 'メ', 'ﾓ': 'モ', 'ﾔ': 'ヤ', 'ﾕ': 'ユ', 'ﾖ': 'ヨ', 'ﾗ': 'ラ', 'ﾘ': 'リ', 'ﾙ': 'ル', 'ﾚ': 'レ', 'ﾛ': 'ロ', 'ﾜ': 'ワ', 'ﾝ': 'ン' };
```
alexesprit, line 34: This map could be extracted to a global scope to avoid creating on every function call.

alexesprit (changes requested):

alexesprit (approved):

alexesprit:
Thanks!

PR #2552, opened 2020-09-10
Add Eggs connector

yayuyokitano:
Adds a connector to grab metadata and scrobble off eggs.mu.
Includes connectors for both artist player and song player.

[Artist player example](https://eggs.mu/artist/RemLinne66666)
[Song player example](https://eggs.mu/song/0e11b8b0d0eafe5c88e17de0fd873314762b9dc39f0ef0cc076d317c2e344d29)

The artist player one is a bit janky but I don't see a better option.

### DIFF
--- a/src/connectors/eggs.js
+++ b/src/connectors/eggs.js
@@ -4,6 +4,8 @@ Connector.getDuration = () => document.querySelector('#aPlayer').duration;
 
 Connector.getCurrentTime = () => document.querySelector('#aPlayer').currentTime;
 
+Connector.playerSelector = '.m-upload-list';
+
 if (window.location.href.includes('/artist/')) {
 	setupArtistPlayer();
 } else {
@@ -11,7 +13,6 @@ if (window.location.href.includes('/artist/')) {
 }
 
 function setupArtistPlayer() {
-	Connector.playerSelector = '.m-upload-list ul';
 
 	Connector.getTrackArt = () => document.querySelector('.pause[style*="display: block;"]').closest('.img-album').querySelector('img').src;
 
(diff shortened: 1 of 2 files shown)


src/connectors/eggs.js
```diff
@@ -0,0 +1,35 @@
...
+
+if (window.location.href.includes('/artist/')) {
+	setupArtistPlayer();
+} else {
+	setupSongPlayer();
+}
+
+function setupArtistPlayer() {
+
+	Connector.getTrackArt = () => document.querySelector('.pause[style*="display: block;"]').closest('.img-album').querySelector('img').src;
```
alexesprit, line 17: This could be improved by overriding `Connector.getTrackInfo` method:
```
Connector.getTrackInfo = () => {
	const songItem = /* get song item */;

	/* get all properties from songItem here */

	return { artist, track, trackArt };
};
```

Also, I would get the parent element by selecting `[id^=song] .pause[style*="display: block;"]` to avoid using `closest` method. But I'm not sure if all song items will match this selector.

alexesprit (changes requested):

src/connectors/eggs.js
yayuyokitano: > Also, I would get the parent element by selecting [id^=song] .pause[style*="display: block;"] to avoid using closest method. But I'm not sure if all song items will match this selector.

I mean i could move the selectors to id^=song, but I would still have to use closest no? As far as I know there is no way to get the parent of a matched element in css (well, other than :has()). `[id^=song] .pause[style*="display: block;"]` will match the pause button itself, not the li element. Unless i'm missing something.

As for the other suggestions, on it.

alexesprit: > [id^=song] .pause[style*="display: block;"] will match the pause button itself, not the li element.

Yes, you're right. Using the `closest` method is fine in this case.

alexesprit (approved):

alexesprit:
Great, thanks!

PR #2553, opened 2020-09-11
Eggspand Eggs connector to include youtube embeds

yayuyokitano:
Uses an injected script to scrobble youtube embeds on Eggs.mu.

Indicators for using injected script vs allframes:

- eggs.mu/artist adds/removes youtube iframe to/from DOM dynamically.
- Also makes it easy for it to work alongside the normal scrobbling.
- It also allows usage of the eggs.mu tags, not having to rely on parsing.

Therefore, I chose to use an injected script with a mutation observer.

Replaces the youtube iframes on eggs.mu/artist with ones made through
youtube embed api, to be able to interact with it and get playing status.

As for eggs.mu/song, youtube embed api is already used, so the script
simply adds an event listener to the existing player variable.

I could use allframes for eggs.mu/song, but its just reusing code from
eggs.mu/artist scrobbler anyway, so it seems like a worse approach.

Also added a simple filter to remove MV suffix as well as (MV) and【MV】
as these denote "music video".

Will probably require some back and forth.

### DIFF
--- a/src/connectors/eggs-dom-inject.js
+++ b/src/connectors/eggs-dom-inject.js
@@ -54,6 +54,11 @@ function placeYoutubeVideo() {
 		height: '290',
 		width: '500',
 		videoId,
+		playerVars: {
+			origin: 'https://eggs.mu/',
+			wmode: 'transparent',
+			rel: 0,
+		},
 		events: {
 			'onStateChange': youtubeStateChange,
 		},
--- a/src/connectors/eggs-dom-inject.js
+++ b/src/connectors/eggs-dom-inject.js
@@ -1,0 +1,80 @@
+'use strict';
+
+/*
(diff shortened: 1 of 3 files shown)


alexesprit (changes requested):

src/connectors/eggs-dom-inject.js
```diff
@@ -0,0 +1,85 @@
...
+			wmode: 'transparent',
+			rel: 0,
+		},
+		events: {
+			'onStateChange': youtubeStateChange,
+		},
+	});
+}
+
+function youtubeStateChange(event) {
```
alexesprit, line 68: It could be named `onYouTubeStateChange` as a better indicator this is event listener callback.

alexesprit (approved):

PR #2562, opened 2020-09-21
Archive.org Remove leading numeric

inverse:
Allows clean scrobbling for releases such as https://archive.org/details/AH003_corwin_trails_-_corwin_trails/Oasis_is_Myth.mp3

### DIFF
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -11,6 +11,8 @@ const trackSelectors = [
 
 const numericTrackRegex = /^\d+\w+/;
 
+const tracksSelector = '.jwrowV2 .ttl';
+
 Connector.currentTimeSelector = '.jw-text-elapsed';
 
 Connector.durationSelector = '.jw-text-duration';
@@ -41,7 +43,7 @@ Connector.getArtistTrack = () => {
 		track = trackParts[2];
 	}
 
-	if (hasAllTracksNumericPrefix('.jwrowV2 .ttl')) {
+	if (hasAllTracksNumericPrefix(tracksSelector)) {
 		track = track.trim().replace(/^(\d+\w+)/, '');
 	}
 
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -49,16 +49,13 @@ Connector.getArtistTrack = () => {
 Connector.albumSelector = '.thats-left > h1 [itemprop=name]';
 
 function hasAllTracksNumericPrefix(trackSelector) {
-
 	const tracks = document.querySelectorAll(trackSelector);
-
 	if (tracks.length === 0) {
 		return false;
 	}
 
 	let hasAllTracksNumericPrefix = true;
 	const re = /^\d+\w+/;
-
 	for (let i = 0; i < tracks.length; ++i) {
 		if (!re.test(tracks[i].innerText)) {
 			hasAllTracksNumericPrefix = false;
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -9,6 +9,8 @@ const trackSelectors = [
 	'.audio-track-list .selected .track-title',
 ];
 
+const numericTrackRegex = /^\d+\w+/;
+
 Connector.currentTimeSelector = '.jw-text-elapsed';
 
 Connector.durationSelector = '.jw-text-duration';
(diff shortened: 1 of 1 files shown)


src/connectors/archive.js
```diff
@@ -39,7 +39,29 @@ Connector.getArtistTrack = () => {
...
+
+function hasAllTracksNumericPrefix(trackSelector) {
+	const tracks = document.querySelectorAll(trackSelector);
+	if (tracks.length === 0) {
+		return false;
+	}
+
+	let hasAllTracksNumericPrefix = true;
+	const re = /^\d+\w+/;
+	for (let i = 0; i < tracks.length; ++i) {
```
alexesprit, line 59: `tracks` is NodeList, which is iterable, so please consider using `for..of` instead.

src/connectors/archive.js
```diff
@@ -39,7 +39,29 @@ Connector.getArtistTrack = () => {
...
+		if (!re.test(tracks[i].innerText)) {
+			hasAllTracksNumericPrefix = false;
```
alexesprit, line 61: You could omit the temp variable and return the value directly.

alexesprit (changes requested):

src/connectors/archive.js
```diff
@@ -39,7 +39,29 @@ Connector.getArtistTrack = () => {
 		track = trackParts[2];
 	}
 
+	if (hasAllTracksNumericPrefix('.jwrowV2 .ttl')) {
+		track = track.trim().replace(/^(\d+\w+)/, '');
+	}
+
```
alexesprit, line 45: Will this work if we move this code inside a MetadataFilter function?

src/connectors/archive.js
```diff
@@ -39,7 +39,29 @@ Connector.getArtistTrack = () => {
...
+function hasAllTracksNumericPrefix(trackSelector) {
+	const tracks = document.querySelectorAll(trackSelector);
+	if (tracks.length === 0) {
+		return false;
+	}
+
+	let hasAllTracksNumericPrefix = true;
+	const re = /^\d+\w+/;
+	for (let i = 0; i < tracks.length; ++i) {
+		if (!re.test(tracks[i].innerText)) {
```
alexesprit, line 60: Is there a reason to use `innerText` here?

alexesprit:
Also I would include the link from PR description as a comment, like we do in the Bandcamp connector.

src/connectors/archive.js
```diff
@@ -39,7 +39,29 @@ Connector.getArtistTrack = () => {
...
 Connector.albumSelector = '.thats-left > h1 [itemprop=name]';
+
+function hasAllTracksNumericPrefix(trackSelector) {
+	const tracks = document.querySelectorAll(trackSelector);
+	if (tracks.length === 0) {
+		return false;
+	}
+
+	let hasAllTracksNumericPrefix = true;
+	const re = /^\d+\w+/;
```
alexesprit, line 58: I think this expression could be extracted to a global scope.

src/connectors/archive.js
inverse: Always forget this one

src/connectors/archive.js
inverse: I like to avoid multiple return points when possible.

src/connectors/archive.js
inverse: How would that work? what would it be like?

src/connectors/archive.js
inverse: What would you use? I wanted to get the text only of an element

inverse:
> Also I would include the link from PR description as a comment, like we do in the Bandcamp connector.

Like a link to a page that does this?

src/connectors/archive.js
alexesprit: `textContent` is prefferable in most cases. See [this](https://stackoverflow.com/a/35213639) answer for details.

src/connectors/archive.js
alexesprit:
```js
const filter = new MetadataFilter({ track: removeNumericPrefixes });

function removeNumericPrefixes(track) {
	if (hasAllTracksNumericPrefix('.jwrowV2 .ttl')) {
		return track.trim().replace(/^(\d+\w+)/, '');
	}

	return track;
}
```

Also, it's worth using a constant for `.jwrowV2 .ttl` selector.

alexesprit:
> Like a link to a page that does this?

Yes, it could be helpful for debugging the connector.

src/connectors/archive.js
inverse: Thanks - I'll give this a shot later :pray:

inverse:
@alexesprit any other feedback?

alexesprit:
@inverse it's almost perfect now, aside the `textContent` request change. :)

inverse:
Sorry - forgot to push it :see_no_evil:

alexesprit (approved):

PR #2567, opened 2020-09-24
Add subfire connector

d3m3vilurr:
Add new connector for [subfire](http://www.subfireplayer.net/)

### DIFF
--- a/src/connectors/subfire.js
+++ b/src/connectors/subfire.js
@@ -1,0 +1,26 @@
+'use strict';
+
+Connector.playerSelector = '#player-cover';
+
+Connector.getArtist = () => document.querySelector('input#artist').value;
+
+Connector.getTrack = () => document.querySelector('input#title').value;
+
+Connector.getAlbum = () => document.querySelector('input#album').value;
+
+Connector.playButtonSelector = '#player-cover #player-btn-playpause.ui-icon-play';
+
+Connector.trackArtSelector = '#player-cover img#coverart';
+
+// original selectors are `#duration-label .currentTime` and `#duration-label .endTime`
+// but these values are pretty unstable during loading time
+Connector.getCurrentTime = () => document.querySelector('#player').currentTime;
+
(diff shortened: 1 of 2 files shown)


alexesprit (changes requested):
Please don't use jQuery. It will be removed in the next major version.

src/connectors/subfire.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+$('#player').bind('playing pause timeupdate durationchange', Connector.onStateChanged);
```
alexesprit, line 3: Why do you use the `playerSelector` property and add listener for these events at the same time? Is there a certain purpose for doing this?

d3m3vilurr: a real player is `#player-cover`. `#player` is just hidden `audio` tag in the bottom of the html to play the song.
when I try to make first connector version, I got a problem that this player receives the loading/updating events from the `#player`, updating progress and duration value would be delayed.

player process could be, starting queue -> update current track -> update `audio` tag -> playing + receiving duration change events.

so I want to get the similar timing of the original actions.
could i have an another option for it?

alexesprit: I suppose adding an event listener for the `#player` element should be enough.

Assigning a value to the `playerSelector` property will make the extension to look for DOM changes of this element. I think it's redundant since we have the event listener set up. Or am I wrong?

Btw, how can I test the connector?

d3m3vilurr: I have been setuped [docker version of supysonic](https://github.com/glogiotatidis/supysonic-docker) into my own server
and I found some my mistakes (defined wrong method).
fixed review things; remove `$` and event handling of the `audio` tag.

and i also tested `playerSelector` to `audio#player`, it doesn't work, that this tag is always hidden and only changed the `src` value.

alexesprit (approved):
Thank you for the PR!

PR #2569, opened 2020-09-25
Add Idagio connector

yayuyokitano:
Adds a connector to app.idagio.com

Most of the shenanigans are to format scrobbles as best as possible

Example format (track-artist-album):
Concerto for Violin and Orchestra in D Major Op. 61: I. Allegro ma non troppo
Ludwig van Beethoven
Concerto for Violin and Orchestra (Valery Gergiev)

### DIFF
--- a/src/connectors/idagio.js
+++ b/src/connectors/idagio.js
@@ -10,24 +10,26 @@ Connector.playerSelector = '.player-PlayerBar__bar--2yos_';
 
 Connector.artistSelector = '.player-PlayerInfo__infoEl--2jhHY span:first-child';
 
-Connector.getTrack = () => getCurrentTrack();
+Connector.getTrack = getCurrentTrack;
 
-Connector.getAlbum = () => getCurrentSymphony();
+Connector.getAlbum = getCurrentSymphony;
 
 Connector.currentTimeSelector = '.player-PlayerProgress__progress--2F0qB>span';
 
 Connector.durationSelector = '.player-PlayerProgress__timeTotal--3aHlj span';
 
-Connector.isPlaying = () => document.querySelector(recordingInfoSelector).innerHTML !== 'Sponsor message' && document.querySelector(pauseButtonSelector).innerHTML === 'Pause';
+Connector.isPlaying = () => Util.getTextFromSelectors(pauseButtonSelector) === 'Pause';
+
+Connector.isScrobblingAllowed = () => Util.getTextFromSelectors(recordingInfoSelector) !== 'Sponsor message';
 
 function getCurrentTrack() {
-	if (document.querySelector(symphonySelector)) {
-		return `${document.querySelector(symphonySelector).innerText}: ${document.querySelector(movementSelector).innerText}`;
-	}
+	const symphony = Util.getTextFromSelectors(symphonySelector);
+	const movement = Util.getTextFromSelectors(movementSelector);
+	return `${symphony}: ${movement}`;
 }
 
 function getCurrentSymphony() {
-	if (document.querySelector(symphonySelector)) {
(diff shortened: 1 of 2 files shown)


src/connectors/idagio.js
```diff
@@ -0,0 +1,33 @@
...
+
+Connector.currentTimeSelector = '.player-PlayerProgress__progress--2F0qB>span';
+
+Connector.durationSelector = '.player-PlayerProgress__timeTotal--3aHlj span';
+
+Connector.isPlaying = () => document.querySelector(recordingInfoSelector).innerHTML !== 'Sponsor message' && document.querySelector(pauseButtonSelector).innerHTML === 'Pause';
+
+function getCurrentTrack() {
+	if (document.querySelector(symphonySelector)) {
+		return `${document.querySelector(symphonySelector).innerText}: ${document.querySelector(movementSelector).innerText}`;
```
alexesprit, line 25: Please avoid using long statements in the template literal, as it reduces readabiliy.

I would write this function to look like this:
```js
function getCurrentTrack() {
	const symphony = Util.getTextFromSelectors(symphonySelector);
	if (symphony) {
		const movement = Util.getTextFromSelectors(movementSelector);
		return `${symphony}: ${movement}`;
	}

	return null;
}
```

src/connectors/idagio.js
```diff
@@ -0,0 +1,33 @@
...
+	}
+}
```
alexesprit, line 27: Please use explicit return statement.

src/connectors/idagio.js
```diff
@@ -0,0 +1,33 @@
...
+
+Connector.getTrack = () => getCurrentTrack();
+
+Connector.getAlbum = () => getCurrentSymphony();
+
+Connector.currentTimeSelector = '.player-PlayerProgress__progress--2F0qB>span';
+
+Connector.durationSelector = '.player-PlayerProgress__timeTotal--3aHlj span';
+
+Connector.isPlaying = () => document.querySelector(recordingInfoSelector).innerHTML !== 'Sponsor message' && document.querySelector(pauseButtonSelector).innerHTML === 'Pause';
```
alexesprit, line 21: Skipping ads and sponsor messages should be done by using `Connector.isScrobblingAllowed` function.

By the way, is there a reason to use `innerHTML` here? If you want to get a text from an element, using `Util.getTextFromSelectors` is recommended.

src/connectors/idagio.js
```diff
@@ -0,0 +1,33 @@
...
+const recordingInfoSelector = '.player-PlayerInfo__recordingInfo--15VMv';
+const pauseButtonSelector = '.player-PlayerControls__btn--1r-vy:nth-child(2) .util-IconLabel__component--3Uitr span';
+
+Connector.playerSelector = '.player-PlayerBar__bar--2yos_';
+
+Connector.artistSelector = '.player-PlayerInfo__infoEl--2jhHY span:first-child';
+
+Connector.getTrack = () => getCurrentTrack();
+
+Connector.getAlbum = () => getCurrentSymphony();
```
alexesprit, line 15: I would not create new functions, but this way also looks fine for me. But I think it's redundant to wrap function call in an arrow function. `Connector.getTrack = getCurrentTrack` is better.

alexesprit (changes requested):

src/connectors/idagio.js
yayuyokitano: Ah i was thinking of avoiding some freak accident because it uses the same element for multiple things, but i dont think it should be an issue on second thought

yayuyokitano:
Looks like using isScrobblingAllowed allows me to yeet the if statements entirely, missed that one oop

alexesprit (approved):
Cool, thank you!

Issue #2481, opened 2020-07-12
Incorrect Title Detection for tracks with Quotations


JP-Ellis:
## Describe the bug
When the track isn't recognized, I've been noticing issues with the automatic detection of the title for tracks names which have quotations.  For example, listening to the [Elgar: Enigma Variations](https://music.youtube.com/playlist?list=OLAK5uy_nnUy3C4IwqteZc1SbUgfZNzmg1SaawjdA) album has song titles of the form:

1. Elgar: Variations On An Original Theme, Op.36 "Enigma" - Theme (Andante)
2. Elgar: Variations On An Original Theme, Op.36 "Enigma" - 1. C.A.E. (L'istesso tempo)
3. ...

and Web Scrobbler just picks out "Enigma" for all of them and puts that as the default prefill when it doesn't recognize the track. 
 This is an issue which appears mostly with classical albums (as it is pretty common to have text in quotes), and I've only noticed it on tracks which weren't recognized to begin with.  I'm not sure whether now that I've added them for this particular album, it will still appear to others.

Another (likely linked) issue encountered is that once I fixed the title for the first track, the next track is detected as being the same as the first.  So in the above case, I would label the first track correctly, and once I start playing the 2nd track it thinks I'm listening to the first again, and so on for all 14 variations.

## Environment (please complete the following information)
- OS: Linux
- Browser: Brave
- Extension version: 2.33.0
- Website: [YouTube Music](https://music.youtube.com)

alexesprit:
This happens due to a special rule to extract track info from videos like [this](https://www.youtube.com/watch?v=zYGCcpF-7nI). Not sure how to fix the issue properly for both cases. 

Perhaps, we can modify the rule to require a dash symbol before a track title, but the extension won't recognize `Artist "Track"` titles (if they exist).

> Another (likely linked) issue encountered is that once I fixed the title for the first track, the next track is detected as being the same as the first. So in the above case, I would label the first track correctly, and once I start playing the 2nd track it thinks I'm listening to the first again, and so on for all 14 variations.

Not sure if it's linked. Saving/loading edited track info is a separate process. Please open a new issue with a more detailed description including steps to reproduce.

JP-Ellis:
I understand the need for that with YouTube itself, but is this parsing necessary for YouTube Music?

alexesprit:
> but is this parsing necessary for YouTube Music?

[...]

PR #2570, opened 2020-09-26
Improve YouTube Music behavior for all channels.

### LINKED ISSUE #2564: Improve YouTube Music connector
**Description**
The song title and artist is very often wrongly recognized on [YouTube Music](https://music.youtube.com/). The track field contains both the artist and title, and the artist field contains a channel name. I looked into the issue and found that it happens on songs that are from a music promotion channel (more precisely, where the channel name is not the artist).

1. Go to a music video from a promotion channel, such as [this one](https://music.youtube.com/watch?v=6DcghQ_kCH8&feature=share).
2. Open the extension's panel and see the wrong data

**Expected behavior**
The channel name should not be scraped since it has nothing to do with the artist or song title. The artist name should appear in the artist field instead of the track field.

**Screenshot**
tmp

**Logs**
Here's a [Pastebin link](https://pastebin.com/2GfTMkkV) with the logs.

**Environment**
-   OS: Windows
-   Browser: Chrome
-   Extension version: v2.36.0

jaccarmac:
We attempt to use a mixed solution which gives correct results for both fully
tagged music (with artist and album name properly reported by YouTube) and
videos where the artist and title are both in the video title (common in the
case of promotional channels).

Resolves #2564.
Reverts 6df5242453e4a72b71b577a9cc41f6820880d8c2.
Partially reverts 58fda0810cd2dd7163ee93ae5a9fe8ada02d1630.

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -26,9 +26,11 @@ Connector.albumSelector = [
 Connector.getArtistTrack = () => {
 	let artist; let track;
 	if (Connector.getAlbum()) {
+		Connector.resetFilter();
 		artist = Util.getTextFromSelectors(channelNameSelector);
 		track = Util.getTextFromSelectors(trackSelector);
 	} else {
+		Connector.applyFilter(MetadataFilter.getYoutubeFilter());
 		({ artist, track } = Util.processYtVideoTitle(Util.getTextFromSelectors(trackSelector)));
 		if (!artist) {
 			artist = Util.getTextFromSelectors(channelNameSelector);
@@ -44,12 +46,3 @@ Connector.isPlaying = () => {
 };
 
 Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
-
-const baseStateChangedWorker = Connector.stateChangedWorker;
-Connector.stateChangedWorker = () => {
-	Connector.metadataFilter = MetadataFilter.getDefaultFilter();
-	if (! Connector.getAlbum()) {
-		Connector.applyFilter(MetadataFilter.getYoutubeFilter());
-	}
-	baseStateChangedWorker();
-};
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -23,7 +23,9 @@ Connector.albumSelector = [
 	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]',
 ];
 
-const videoHasAlbum = () => !!Connector.getAlbum();
+function videoHasAlbum() {
+	return !!Connector.getAlbum();
+}
 
 Connector.getArtistTrack = () => {
 	let artist; let track;
@@ -47,7 +49,9 @@ Connector.isPlaying = () => {
 
 Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
 
-const filterYoutubeIfNonAlbum = (text) => videoHasAlbum() ? text : MetadataFilter.youtube(text);
+function filterYoutubeIfNonAlbum(text) {
+	return videoHasAlbum() ? text : MetadataFilter.youtube(text);
+}
 
 const conditionalYoutubeFilter = new MetadataFilter({ track: filterYoutubeIfNonAlbum });
 
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -45,4 +45,11 @@ Connector.isPlaying = () => {
 
 Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
 
-Connector.applyFilter(MetadataFilter.getYoutubeFilter());
+const baseStateChangedWorker = Connector.stateChangedWorker;
+Connector.stateChangedWorker = () => {
+	Connector.metadataFilter = MetadataFilter.getDefaultFilter();
+	if (! Connector.getAlbum()) {
+		Connector.applyFilter(MetadataFilter.getYoutubeFilter());
+	}
+	baseStateChangedWorker();
+};
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -2,16 +2,13 @@
 
 const playButtonSelector = '.ytmusic-player-bar.play-pause-button #icon > svg > g > path';
 const trackArtSelector = '.ytmusic-player-bar.image';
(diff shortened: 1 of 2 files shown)


jaccarmac:
This definitely needs a pin in it for now, as it mishandles the Elgar cases from #2481. It appears that we need to apply the filters conditionally.

jaccarmac:
It does seem to handle the cases in #2391 correctly, sans the "live" part of the Zappa track. That's probably because of the metadata filter. Floyd and Tull seem :+1: though.

jaccarmac:
How's that hack look :)

jaccarmac:
Still doesn't work quite right after watching one album-less video. Seems like the code I wrote doesn't actually reset the filter to default.

src/connectors/youtube-music.js
```diff
@@ -26,10 +23,33 @@ Connector.albumSelector = [
...
 Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
+
+const baseStateChangedWorker = Connector.stateChangedWorker;
+Connector.stateChangedWorker = () => {
+	Connector.metadataFilter = MetadataFilter.getDefaultFilter();
+	if (! Connector.getAlbum()) {
+		Connector.applyFilter(MetadataFilter.getYoutubeFilter());
+	}
+	baseStateChangedWorker();
+};
```
alexesprit, line 55: This is a bad idea, and it's relies on the internal base connector implementation.

Remember, you can use built-in filter functions. I would use the following approach:
1. Use `MetadataFilter.youtube` function to filter the track title. Since it's just a function, you can use it in a condition block.
2. Add some caching to prevent calling the filter function on every DOM change.

I know this limitation, but for now we have no cleaner solution.

alexesprit:
By the way you can use the "Convert to draft" link for toggling draft state. :)

image

jaccarmac:
Ah, nice, hadn't used the draft feature before.

Not having worked with filters before and unfamiliar with the internals, I fixed the issue by adding a reset method and using that instead. The side effect in `getArtistTrack` is less than ideal, but it does prevent the filtering-every-DOM-change problem by (I think) shuffling the set of loaded filters instead.

src/connectors/youtube-music.js
jaccarmac: Can this be resolved given the new implementation?

jaccarmac:
If anyone's got remaining complaints, I'd be happy to review them. Otherwise I'll look to rebase this tomorrow and double check the change against tip.

src/connectors/youtube-music.js
alexesprit: It still looks hacky, but for now we have no other solutions.

src/connectors/youtube-music.js
```diff
@@ -26,6 +23,22 @@ Connector.albumSelector = [
...
 ];
 
+Connector.getArtistTrack = () => {
+	let artist; let track;
+	if (Connector.getAlbum()) {
+		Connector.resetFilter();
+		artist = Util.getTextFromSelectors(channelNameSelector);
+		track = Util.getTextFromSelectors(trackSelector);
+	} else {
+		Connector.applyFilter(MetadataFilter.getYoutubeFilter());
```
alexesprit, line 33: `MetadataFilter.getYoutubeFilter` returns a new instance on every call.

src/connectors/youtube-music.js
```diff
@@ -26,6 +23,22 @@ Connector.albumSelector = [
 	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]',
 ];
 
+Connector.getArtistTrack = () => {
+	let artist; let track;
+	if (Connector.getAlbum()) {
+		Connector.resetFilter();
```
alexesprit, line 29: Calling `Connector.resetFilter` will create a new filter instance on every call.

src/connectors/youtube-music.js
jaccarmac: `applyFilter` also extends a new instance of the default filter every time (https://github.com/web-scrobbler/web-scrobbler/blob/8573651318feeb9730ce11c97e9e79a8ad57110b/src/core/content/connector.js#L447). Do you have any ideas about how to proceed?

src/connectors/youtube-music.js
jaccarmac: Since the metadata filter package mutates the filter object in `extend`, I'm not sure if there's a good way to avoid this without saving a copy of the default filter and then deep cloning on every reset. Thoughts?

https://github.com/web-scrobbler/metadata-filter/blob/67f7fb53b97edc34c07946e5fb0d8c133adac7ce/src/filter.js#L63-L66

alexesprit:
```js
const youtubeMusicFilter = new MetadataFilter({ track: filterTrack });

Connector.applyFilter(youtubeMusicFilter);

function filterTrack(text) {
    if (Connector.getAlbum()) {
        return MetadataFilter.youtube(text);
    }

    return text;
}
```

Although `filterTrack` is not a pure function, it has no side effects, so it's acceptable.

src/connectors/youtube-music.js
alexesprit: Yes, but `Connector.applyFilter` is not supposed to be called in getters, so it was ok.

I suggested another solution below.

alexesprit:
Also, I think we could add a new function to check if we should filter the input.

```js
function filterTrack(text) {
    if (isVideoFromYouTube()) {
        return MetadataFilter.youtube(text);
    }

    return text;
}

function isVideoFromYouTube() {
    return !Connector.getAlbum();
}
```

jaccarmac:
Ah, of course, that makes sense. Thanks.

And yes, fun hash :D

jaccarmac:
If ternaries are verboten, I can make that more verbose. Thanks to your suggestions, I quite like the way it turned out.

src/connectors/youtube-music.js
```diff
@@ -23,14 +23,14 @@ Connector.albumSelector = [
 	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]',
 ];
 
+const videoHasAlbum = () => !!Connector.getAlbum();
```
alexesprit, line 26: Please don't use function expressions. We use function declarations everywhere to declare a function.

alexesprit:
@aamelin Hi, could you check and test this PR? This one partially reverts changes you made.

alexesprit (approved):

PR #2573, opened 2020-09-29
Fix typos in readme

yayuyokitano:
Dunno if this is the correct way to submit this but here goes
Adds a space and removes an s from one of the links in readme.
The s may not be a typo but it is the more common way to write it anyway
Space is definitely a typo.


inverse (approved):
Thanks :)

Issue #2563, opened 2020-09-22
Laut.fm connector


Seryc:
[Laut.fm](https://laut.fm/) is a web portal for creating radio programs and is offered by Laut AG, which also operates the online music magazine laut.de. It was founded on June 10, 2005 as an online radio. Internet users can use it to create their own radio program. 
Please add a connector for scrobbling. Many thanks!

PR #2577, opened 2020-10-01
Add support for Laut.fm

clasick:
Added support for Laut.fm as requested by @Seryc in  #2563

### DIFF
--- a/src/connectors/laut.fm.js
+++ b/src/connectors/laut.fm.js
@@ -7,3 +7,11 @@ Connector.artistSelector = '.player-display__meta .player-display__meta--artist'
 Connector.trackSelector = '.player-display__meta .player-display__meta--title';
 
 Connector.isPlaying = () => $('.fm-player__btn').hasClass('playbutton--playing');
+
+function removeEnclosingQuotes(track) {
+	return track.trim().slice(1, -1);
+}
+
+const filter = new MetadataFilter({ track: removeEnclosingQuotes });
+
+Connector.applyFilter(filter);
--- a/src/connectors/laut.fm.js
+++ b/src/connectors/laut.fm.js
@@ -1,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.fm-player';
+
+Connector.artistSelector = '.player-display__meta .player-display__meta--artist';
+
+Connector.trackSelector = '.player-display__meta .player-display__meta--title';
+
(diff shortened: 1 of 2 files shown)


src/connectors/laut.fm.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.fm-player';
+
+Connector.artistSelector = '.player-display__meta .player-display__meta--artist';
+
+Connector.trackSelector = '.player-display__meta .player-display__meta--title';
+
+Connector.isPlaying = () => $('.fm-player__btn').hasClass('playbutton--playing');
```
inverse, line 9:
```suggestion
Connector.pauseButtonSelector = '.fm-player__btn.playbutton--playing';
```

Same as the other PR

src/connectors/laut.fm.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.fm-player';
+
+Connector.artistSelector = '.player-display__meta .player-display__meta--artist';
+
+Connector.trackSelector = '.player-display__meta .player-display__meta--title';
```
inverse, line 7: I think we need to add a filter to strip off enclosing `"` e.g. `"Violent Sun"` -> `Violent Sun`

Something like this should work:

```
const filter = new MetadataFilter({ track: removeEnclosingQuotes });

function removeEnclosingQuotes(track) {
		track = track,substring(1);
    return track.substring(0, track.length - 1);
}
```

Although I am sure @alexesprit knows one built in that will be better :+1:

inverse (changes requested):

src/connectors/laut.fm.js
alexesprit: Instead of using `String.substring` we can use `String.slice`:
```js
function removeEnclosingQuotes(track) {
    return track.slice(1, -1);
}
```

but this function will remove any characters.

inverse: @alexesprit thanks! 

From looking it always pads with `"` so I think this is fine. Or would you rather go explicit and asset char at pos and remove only then?

clasick: @inverse Thanks for catching that quotes issue. Added a change to fix that through @alexesprit's suggestion.

alexesprit (approved):

inverse (approved):

Issue #2506, opened 2020-08-03
Scrobble https://relax-fm.ru/


pedrojvaz:
**Provide information about website**
https://relax-fm.ru/

Relax FM is a radio station out of Moscow, Russia, that features instrumental, English, and foreign music curated to relax you.  

**Additional context**
No registration needed

inverse:
Is this even possible? When I visit and click play I see no meta data for us to scrape?

alexesprit:
@inverse Yes, they have metadata, so it's not a problem to make the connector.

![image](https://user-images.githubusercontent.com/1119267/89577540-825ba100-d839-11ea-89ca-9d1ad85f4596.png)

PR #2578, opened 2020-10-01
Add support for Relaxfm

clasick:
Added support for Relaxfm as requested by @pedrojvaz in #2506

### DIFF
--- a/src/connectors/relaxfm.js
+++ b/src/connectors/relaxfm.js
@@ -1,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.main-player__wrapp';
+
+Connector.artistSelector = '.TitleExecutorAir';
+
+Connector.trackSelector = '.TitleTracksAir';
+
+Connector.isPlaying = () => $('.toggleplayer').hasClass('stop');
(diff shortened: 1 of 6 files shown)


src/connectors/relaxfm.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '.main-player__wrapp';
+
+Connector.artistSelector = '.TitleExecutorAir';
+
+Connector.trackSelector = '.TitleTracksAir';
+
+Connector.isPlaying = () => $('.toggleplayer').hasClass('stop');
```
inverse, line 9: Could be simplified to this.

```suggestion
Connector.pauseButtonSelector = '.toggleplayer.stop';
```

Would avoid the dependency on jQuery which is being phased out in preparation for v3 with bootstrap 5

inverse (changes requested):
Looks good apart from a minor comment :+1:

alexesprit (approved):

src/core/connectors.js
```diff
@@ -1646,6 +1646,11 @@ const connectors = [{
 	],
 	js: 'connectors/idagio.js',
 	id: 'idagio',
+}, {
+	label: 'Relaxfm',
```
alexesprit, line 1650: Could you also update the label to "Relax FM", as it's displayed on the website?

clasick: Updated the label.

inverse (approved):
Nice one :+1:

PR #2589, opened 2020-10-08
Add bullofheaven.com support

### LINKED ISSUE #2426: Try adding "bullofheaven.com" ?
bullofheaven.com is the site for band "Bull Of Heaven" they make Experimental/Dark Ambient/Drone tracks. not every track is playable on some media players, either the tracks are too long, or you have to do "some puzzle" to get the audio files for them.

http://bullofheaven.com/

inverse:
Resolves #2426

### DIFF
--- a/src/connectors/bullofheaven.com.js
+++ b/src/connectors/bullofheaven.com.js
@@ -4,7 +4,6 @@ const artistName = 'Bull of Heaven';
 
 const trackPrefixRegex = /\d+ - /;
 
-const playButtonSelector = '.jp-play';
 
 const filter = new MetadataFilter({ track: removeNumericPrefixes });
 
@@ -26,4 +25,4 @@ Connector.currentTimeSelector = '.jp-current-time';
 
 Connector.durationSelector = '.jp-duration';
 
-Connector.isPlaying = () => !Util.isElementVisible(playButtonSelector);
+Connector.playButtonSelector = '.jp-play';
--- a/src/connectors/bullofheaven.com.js
+++ b/src/connectors/bullofheaven.com.js
@@ -2,18 +2,23 @@
 
 const artistName = 'Bull of Heaven';
 
-const trackRegex = /\d+ - /;
+const trackPrefixRegex = /\d+ - /;
 
 const playButtonSelector = '.jp-play';
+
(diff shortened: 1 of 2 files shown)


src/connectors/bullofheaven.com.js
```diff
@@ -0,0 +1,24 @@
...
+	return trackElement.textContent.replace(trackRegex, '');
+};
+
+Connector.trackArtSelector = '#jp_poster_0';
+
+Connector.currentTimeSelector = '.jp-current-time';
+
+Connector.durationSelector = '.jp-duration';
+
+Connector.isPlaying = () => !Util.isElementVisible(playButtonSelector);
```
alexesprit, line 24: Why `Connector.playButtonSelector` is not used here?

src/connectors/bullofheaven.com.js
```diff
@@ -0,0 +1,24 @@
...
+
+const playButtonSelector = '.jp-play';
+
+Connector.playerSelector = '#jp_container_N';
+
+Connector.getArtist = () => artistName;
+
+Connector.getTrack = () => {
+	const trackElement = document.querySelector('.jp-playlist-current .playlist-title');
+	return trackElement.textContent.replace(trackRegex, '');
```
alexesprit, line 15: MetadataFilter is a preferred way to modify track info fields.

alexesprit (changes requested):

src/connectors/bullofheaven.com.js
inverse: In this case it's the negative of what's expected. e.g. `playButtonSelector` is expecting it to be visible - in this case it's the reverse.

alexesprit: When a track is play, the play button is expected to be hidden. The default implemetation perform the same check:
https://github.com/web-scrobbler/web-scrobbler/blob/a97bae9740438708ae85b77b8acc47940eeeaf68/src/core/content/connector.js#L333-L338

inverse: Weird - tried this before and the logic seemed reverse. Now it's working :boom:

alexesprit (approved):

PR #2591, opened 2020-10-09
Add Brain.fm connector.

### LINKED ISSUE #1907: support brain.fm
Support brain.fm. It's a bit different than the other 

I am not exactly sure. Since brain.fm is just a stream of music without separate tracks, it is different than most scrobbling. It would be cool to support it though in some way.

I'm just skipping the rest because I really don't know how it should work. Perhaps more thoughtful people could leave comments.

jaccarmac:
Closes #1907.


jaccarmac:
Duration is fiddly to get, as it only updates during mouseover. Thus, scrobbles happen after 30 seconds. Theoretically, we could calc the location uses the scrub bar, but for now this seems to address most of the requirements in #1907 

The `^=` selectors can possibly be simplified to use the hardcoded IDs, but seems to work using the regex match. (The trailing underscores are necessary, though.)

inverse (approved):
Seems like a good compromise to allow people to use this service.

Not a user so cannot verify. 

P.s there are merge conflicts.

alexesprit (approved):

PR #2596, opened 2020-10-16
Add Migu Music connector

xyauhideto:
Add a connector for [Migu Music](https://music.migu.cn).

### DIFF
--- a/src/connectors/migu-music.js
+++ b/src/connectors/migu-music.js
@@ -1,0 +1,33 @@
+'use strict';
+
+Connector.playerSelector = '.mod-player';
+
+Connector.trackSelector = '.song a';
+
+Connector.artistSelector = '.singer a';
+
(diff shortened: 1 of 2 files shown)


src/connectors/migu-music.js
```diff
@@ -0,0 +1,33 @@
...
+
+Connector.timeInfoSelector = '.current-time';
+
+Connector.trackArtSelector = '.song-cover img';
+
+Connector.getTrackArt = () => {
+	const element = Util.queryElements(Connector.trackArtSelector);
+	if (!element) {
+		return null;
+	}
```
alexesprit, line 21: Why do you override `getTrackArt`? When the `trackArtSelector` property is defined, the default implementation will get a track art URL from an element matching the selector.

alexesprit (approved):

PR #2597, opened 2020-10-16
Eggs Connector: Load youtube iframe api locally

yayuyokitano:
Uses local copy of youtube iframe api for Eggs connector
This is for security purposes, and to not get yeeted off AMO

This is gonna need some looking at.
So first of all this errors. A lot.
It works just fine from my testing, but it throws errors in console.
Not sure if it's fixable, I think its youtube not liking the local thing.

EDIT: It no longer uses youtube scripts.
It now uses postMessage with the iframe directly.


alexesprit:
Not sure if the extension with these obfuscated scipts will pass the review.

yayuyokitano:
I think I managed to reverse engineer it

yayuyokitano:
Hopefully this is good enough for AMO
If they don't allow this style of integration I give up

src/connectors/eggs-dom-inject.js
```diff
@@ -24,41 +27,96 @@ function toggleExternalPlayer(mutationList) {
...
+	window.postMessage({
+		sender: 'web-scrobbler',
+		playerType: `youtube${playerTypeSuffix}`,
+		isPlaying: data.info === 1,
+		timeInfo: {
+			currentTime: currentTime || 0,
+			duration,
+		},
+		trackInfo: {
+			artist: parentElmt.querySelector(`.artist_name${(isArtistPage) ? '' : ' a'}`).innerText,
```
alexesprit, line 95: `innerText` could be replaced with `textContent`.

alexesprit (changes requested):

src/connectors/eggs-dom-inject.js
```diff
@@ -6,6 +6,9 @@
  */
 
 const isArtistPage = window.location.href.includes('/artist/');
+let frameID = '';
+let currentTime = 0;
+let duration = 180;
```
alexesprit, line 11: Why did you set the initial duration value to 180?

yayuyokitano: I set it pretty arbitrarily. It's around about the lower end of the typical length of a song. It's just there in case something were to completely break, it may fail a bit more gracefully. Unless something breaks, it will never actually be sent as the duration of a song.

alexesprit: I think it should be `0` or `null`; the core will handle nullable duration.

alexesprit (approved):

PR #2621, opened 2020-10-30
Fix byte.fm connector following update

halhex:
Updated the [byte.fm](byte.fm) connector to mirror the breaking changes made to the player.


inverse (approved):
Nice one :tada: thanks for contributing!

PR #2627, opened 2020-11-08
Implement scrobbling individual tracks from album videos

rgeorgiev583:
Implement an option which enables the user to scrobble individual tracks instead of the whole album if a music video has a separate chapter for each track of an album.

The option is currently only implemented for the YouTube connector.

### DIFF
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -257,10 +257,7 @@ function getTrackInfoFromDescription() {
 }
 
 function getTrackInfoFromChapters() {
-	const trackInfo = {};
-	trackInfo.artist = null;
-	trackInfo.track = Util.getTextFromSelectors(chapterNameSelector);
-	return trackInfo;
+	return Util.splitArtistTrack(Util.getTextFromSelectors(chapterNameSelector), null, false, true);
 }
 
 function getTrackInfoFromTitle() {
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -257,7 +257,7 @@ function getTrackInfoFromDescription() {
 }
 
 function getTrackInfoFromChapters() {
-	return Util.splitArtistTrack(Util.getTextFromSelectors(chapterNameSelector), null, false, true);
+	return Util.splitArtistTrack(Util.getTextFromSelectors(chapterNameSelector), null, { swap: false }, { allowTrackOnly: true });
 }
 
 function getTrackInfoFromTitle() {
--- a/src/connectors/youtube.js
+++ b/src/connectors/youtube.js
@@ -1,12 +1,6 @@
 'use strict';
 
 /**
- * Whether to scrobble individual tracks from album videos.
- * @type {Boolean}
- */
-let scrobbleIndividualTracks = false;
-
-/**
  * Array of categories allowed to be scrobbled.
  * @type {Array}
  */
@@ -40,7 +34,11 @@ const categoryEntertainment = 'Entertainment';
 let currentVideoDescription = null;
 let artistTrackFromDescription = null;
 
-let trackInfoGetters = null;
+const trackInfoGetters = [
+	getTrackInfoFromChapters,
+	getTrackInfoFromDescription,
+	getTrackInfoFromTitle,
+];
 
 readConnectorOptions();
 setupEventListener();
@@ -48,24 +46,20 @@ setupEventListener();
 Connector.playerSelector = '#content';
 
 Connector.getTrackInfo = () => {
-	let trackInfo = {};
+	const trackInfo = {};
 
 	for (const getter of trackInfoGetters) {
-		if (scrobbleIndividualTracks) {
-			const currentTrackInfo = getter();
-			if (!currentTrackInfo) {
-				continue;
-			}
+		const currentTrackInfo = getter();
+		if (!currentTrackInfo) {
+			continue;
+		}
 
-			if (!trackInfo.artist) {
-				trackInfo.artist = currentTrackInfo.artist;
-			}
+		if (!trackInfo.artist) {
+			trackInfo.artist = currentTrackInfo.artist;
+		}
 
-			if (!trackInfo.track) {
-				trackInfo.track = currentTrackInfo.track;
-			}
-		} else {
-			trackInfo = getter();
+		if (!trackInfo.track) {
+			trackInfo.track = currentTrackInfo.track;
 		}
 
 		if (!Util.isArtistTrackEmpty(trackInfo)) {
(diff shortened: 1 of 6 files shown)


alexesprit:
Why this behavior should be optional?

src/connectors/youtube.js
```diff
@@ -203,6 +219,18 @@ async function fetchCategoryName(videoId) {
  * Asynchronously read connector options.
  */
 async function readConnectorOptions() {
+	if (await Util.getOption('YouTube', 'scrobbleIndividualTracks')) {
+		scrobbleIndividualTracks = true;
+		trackInfoGetters = scrobbleIndividualTracks ? [
```
alexesprit, line 224: So this option just allows to reorder the list of getters. Do we really need this? Have you an example where both the video description and chapters contain track info?

src/connectors/youtube.js
```diff
@@ -229,9 +257,7 @@ function getTrackInfoFromDescription() {
 }
 
 function getTrackInfoFromChapters() {
-	return Util.splitArtistTrack(
-		Util.getTextFromSelectors(chapterNameSelector)
-	);
+	return Util.splitArtistTrack(Util.getTextFromSelectors(chapterNameSelector), null, false, true);
```
alexesprit, line 260: You should not use boolean type as a parameter. You cannot say what these `false` and `true` mean, unless you read documentation for this function.

That's why the `swap` argument is wrapped in an object, so you'll call it like
```js
Util.splitArtistTrack(
		Util.getTextFromSelectors(chapterNameSelector), null, { swap: true }
);
```

But I think this logic should not be implemented in this function. This is a quite rare case.

rgeorgiev583:
> Why this behavior should be optional?

Because it's a significant modification of the current behavior of the extension for album videos, and I suppose that we want to be backwards compatible in this regard.

src/connectors/youtube.js
rgeorgiev583: > So this option just allows to reorder the list of getters. Do we really need this? Have you an example where both the video description and chapters contain track info?

The idea of this change is to check chapters for track info *before* the title so that the individual tracks would be scrobbled instead of the whole album as a single track.

Concerning your second question: chapters are usually used for track info on album videos, and (according to [this YouTube support page](https://support.google.com/youtube/answer/9884579?hl=en)) they are generated from the description.  That's why I don't think that the description needs to be checked at all when chapters are present.

src/connectors/youtube.js
rgeorgiev583: > You should not use boolean type as a parameter. You cannot say what these `false` and `true` mean, unless you read documentation for this function.
> 
> That's why the `swap` argument is wrapped in an object, so you'll call it like
> 
> ```js
> Util.splitArtistTrack(
> 		Util.getTextFromSelectors(chapterNameSelector), null, { swap: true }
> );
> ```
> 
> But I think this logic should not be implemented in this function. This is a quite rare case.

Okay, I'll fix the boolean argument.

Do you have any suggestions for any other places where to implement the logic?  I think this function is the most appropriate one.  Besides, I've tried to preserve its interface and behavior as much as possible.

rgeorgiev583: Given that I have issues using this branch with [some album videos whose track names contain various separator characters](https://www.youtube.com/watch?v=aur4RKAgbqE), I'd like to ask if we should support artist names in chapter titles at all.  Admittedly, this is a really obscure corner case.  (What is more, it is entirely possible for a compilation video to contain tracks from different artists.)

src/connectors/youtube.js
alexesprit: You're right. chapters should be checked first. The only thing I don't understand why to make this behavior optional.

alexesprit:
> Because it's a significant modification of the current behavior of the extension for album videos, and I suppose that we want to be backwards compatible in this regard.

Could you provide some examples where *toggling* this option would be useful?

src/connectors/youtube.js
alexesprit: > Do you have any suggestions for any other places where to implement the logic?

You can use it in the `getTrackInfoFromChapters` functions. Just return the `track` property only. There's no sense to add this simple logic in other functions.

alexesprit: By the way, when I suggested to wrap a boolean argument into an object, I meant wrapping multiple arguments in a *single* object.

```js
function test(arg1, { arg2 = false, arg3 = true } = {}) {
}

test(0, { arg2: true, arg3: false });
```

This is just a note, I'm still against adding the `allowTrackOnly` argument. :)

rgeorgiev583: > > Do you have any suggestions for any other places where to implement the logic?
> 
> You can use it in the `getTrackInfoFromChapters` functions. Just return an object with the `track` property only, if `Util.splitArtistTrack` returns an empty result.
> 
> There's no sense to add this simple logic in other functions.

Ok, I tend to agree that it would be an unnecessary complication of the utility functions, so I implemented it the way you want :)

> By the way, when I suggested to wrap a boolean argument into an object, I meant wrapping multiple arguments in a _single_ object.
Ah, *now* I understand the motivation behind your idea.  So the object would become something like a flags argument.

rgeorgiev583:
> > Because it's a significant modification of the current behavior of the extension for album videos, and I suppose that we want to be backwards compatible in this regard.
> 
> Could you provide some examples where _toggling_ this option would be useful?

I can't think of any, to be honest.  I only implemented it as an option for the sake of backwards compatibility.

alexesprit:
> I only implemented it as an option for the sake of backwards compatibility.

Okay, I got it.

I think it's ok to keep new behavior w/o a way to revert back. As I said previously, chapters should be parsed firstly. The previous way was wrong.

src/connectors/youtube.js
```diff
@@ -229,9 +239,12 @@ function getTrackInfoFromDescription() {
 }
 
 function getTrackInfoFromChapters() {
-	return Util.splitArtistTrack(
-		Util.getTextFromSelectors(chapterNameSelector)
-	);
+	const chapterName = Util.getTextFromSelectors(chapterNameSelector);
+	const artistTrack = Util.splitArtistTrack(chapterName);
+	if (!artistTrack.artist && !artistTrack.track) {
```
alexesprit, line 244: Could you use `isArtistTrackEmpty` from `Util` module here?

rgeorgiev583: > Could you use `isArtistTrackEmpty` from `Util` module here?

I think it is sufficient only to check whether the track name is present, so I implemented it this way instead.  Is that fine?

alexesprit (approved):

PR #2648, opened 2020-12-02
Improve Idagio connector

yayuyokitano:
Removes trailing dash from track tag if it exists.
Some affected tracks can be seen [here](https://app.idagio.com/albums/saint-saens-violin-sonata-no-1-cello-sonata-no-1-and-piano-trio-no-2)

### DIFF
--- a/src/connectors/idagio.js
+++ b/src/connectors/idagio.js
@@ -24,6 +24,12 @@ Connector.isScrobblingAllowed = () => Util.getTextFromSelectors('.player-PlayerI
 
 function getCurrentTrack() {
 	let track = Util.getTextFromSelectors(trackSelector).split(' – ').slice(1).join(': ').trim();
+	/*
+	 * Remove trailing dash. Example from https://app.idagio.com/albums/saint-saens-violin-sonata-no-1-cello-sonata-no-1-and-piano-trio-no-2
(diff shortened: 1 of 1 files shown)


src/connectors/idagio.js
```diff
@@ -23,7 +23,11 @@ Connector.isPlaying = () => Util.getTextFromSelectors(pauseButtonSelector) === '
 Connector.isScrobblingAllowed = () => Util.getTextFromSelectors('.player-PlayerInfo__recordingInfo--15VMv') !== 'Sponsor message';
 
 function getCurrentTrack() {
-	return Util.getTextFromSelectors(trackSelector).split(' – ').slice(1).join(': ');
+	let track = Util.getTextFromSelectors(trackSelector).split(' – ').slice(1).join(': ').trim();
```
alexesprit, line 26: It's not clear what (and why) transformations are performed. Could you add a comment with some examples?

alexesprit: Sorry if I was not clear enough. I was talking about splitting and joining a track title, not about removing a dash. I know this line wasn't added in this PR, but I think it's better to have a comment describing why this transformation is added.

yayuyokitano: My bad, I should've looked closer. Added now

alexesprit (approved):

PR #2652, opened 2020-12-03
Add support for Maloja's native API

angguss:
Adds support for the native Maloja API instead of using Maloja's listenbrainz compatibility interface, solves intermittent connectivity issues.

### DIFF
--- a/src/core/background/scrobbler/maloja-scrobbler.js
+++ b/src/core/background/scrobbler/maloja-scrobbler.js
@@ -97,7 +97,8 @@ define((require) => {
 			let response = null;
 			try {
 				response = await Util.timeoutPromise(timeout, promise);
-				await response.json();
+				if (response.status != 200)
+					return ServiceCallResult.ERROR_OTHER;
 			} catch (e) {
 				this.debugLog('Error while sending request', 'error');
 				return ServiceCallResult.ERROR_OTHER;
--- a/src/core/background/scrobbler/maloja-scrobbler.js
+++ b/src/core/background/scrobbler/maloja-scrobbler.js
@@ -47,6 +47,8 @@ define((require) => {
 
 		/** @override */
 		async getSession() {
+			if (!this.userToken)
+				throw ServiceCallResult.ERROR_AUTH;
 			return { sessionID: this.userToken };
 		}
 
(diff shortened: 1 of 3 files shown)


src/core/background/scrobbler/maloja-scrobbler.js
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		/** @override */
+		async getAuthUrl() {
+			return null;
+		}
+
+		/** @override */
+		async getSession() {
+			return { sessionID: this.userToken };
```
alexesprit, line 50: This makes the extension to think a user is always signed in to Maloja. I suggest to check if the `userToken` property is not empty before returning the session info; throw an error otherwise.

src/core/background/scrobbler/maloja-scrobbler.js
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		/** @override */
+		async isReadyForGrantAccess() {
+			return false;
+		}
+
+		/** @override */
+		async sendNowPlaying(song) {
+			const { sessionID } = await this.getSession();
```
alexesprit, line 60: We could get the session ID inside the `sendRequest` method.

src/core/background/scrobbler/maloja-scrobbler.js
```diff
@@ -0,0 +1,118 @@
...
+
+			const songData = this.makeTrackMetadata(song);
+			songData.time = song.metadata.startTimestamp;
+
+			return this.sendRequest(songData, sessionID);
+		}
+
+		/**
+		 * Private methods
+		 * @param params
```
alexesprit, line 77: Could you fix ESLint warnings?

src/core/background/scrobbler/maloja-scrobbler.js
```diff
@@ -0,0 +1,118 @@
...
+				body: JSON.stringify(params),
+			};
+
+			const promise = fetch(this.userApiUrl, requestInfo);
+			const timeout = BaseScrobbler.REQUEST_TIMEOUT;
+
+			let response = null;
+			try {
+				response = await Util.timeoutPromise(timeout, promise);
+				await response.json();
```
alexesprit, line 98: Calling `Response.json` method is redundant w/o assigning the result value to a variable.

For debugging we print service responses to the console (see the ListenBrainz scrobbler for example).

angguss:
The latest commits should address the issues raised above

alexesprit (approved):

PR #2671, opened 2020-12-17
Add Dash Radio Connector

uwueviee:
This adds (or re-adds) support for https://dashradio.com

### DIFF
--- a/src/connectors/dashradio.js
+++ b/src/connectors/dashradio.js
@@ -1,9 +1,17 @@
 'use strict';
 
+const filter = MetadataFilter.createFilter({
+	artist: removeByText,
+});
+
 Connector.playerSelector = '#music-dataview-container';
 
-Connector.getArtist = () => {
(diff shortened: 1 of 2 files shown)


src/connectors/dashradio.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '#music-dataview-container';
+
+Connector.getArtist = () => {
+	return Util.getTextFromSelectors('#artist-text').substring(3);
```
alexesprit, line 6: If you want to remove "By" prefix, I suggest to use MetadataFilter.

See the `focusatwill` connector for example.

uwueviee: Switched to that, thanks!

alexesprit (approved):

Sutiwas031189:
ไม่อ่ะไรดู

Issue #2604, opened 2020-10-20
Addon doesn't work with Amazon stations after the update


rahonejm:
I went to Amazon Music today and it seems to have been updated. The new UI is much worse than before, but that's a talk for another time. The problem is the addon doesn't recognize the songs being played in stations anymore. Well, it does. But only the first song. After that, it says the "track isn't recognized". Reloading the page gets the addon to recognize the song again, but only until the next one starts. It worked perfectly as recently as yesterday, before I got hit by the update. Playing songs from anything that isn't a station (playlists, artists, albums) seems to work fine.

inverse:
Hi @rahonejm 

What amazon music site are you on? For me it's working on .de

![Uploading image.png…]()

rahonejm:
> 
> 
> Hi @rahonejm
> 
> What amazon music site are you on? For me it's working on .de
> 
> ![image](https://user-images.githubusercontent.com/713525/96632470-0c36e880-1318-11eb-8f3e-11f6c11c0413.png)
Before the update, the site looked like the one in your image and it worked fine. Now it looks like this (in Brazil):
![amazon 2](https://user-images.githubusercontent.com/36547869/96636198-53f74900-12f3-11eb-80ae-7823299b6fca.png)
![amazon](https://user-images.githubusercontent.com/36547869/96636201-55287600-12f3-11eb-83f7-2f299a0187bd.png)

And, as you can see, the addon seems to have been broken in the process.

inverse:
Ouch - yes that's def a massive overhaul.

Unfortunately I don't see it on the .de version.

We will have to wait for it to be rolled out here unless you or someone else has it and feels confident with making the modifications to support both layouts.

rahonejm:
Well, I'm not good at coding so giving feedback is probably the best I can do. Since this [reddit post](https://www.reddit.com/r/AmazonMusic/comments/jci83v/do_not_upgrade_the_desktop_client_to_80_large/) is 4 days old and I got the update just today, it seems they're slowly rolling it across the world. What a terrible update, let me tell you.

inverse:
Still no update in Germany :(

rahonejm:
Ha, lucky you. Hopefully it never gets there.

alexesprit:
> Before the update, the site looked like the one in your image and it worked fine. Now it looks like this (in Brazil):

[...]

PR #2675, opened 2020-12-18
Fix amazon connector

sunasaji:
**Problem**
Songs are not recognized. On Chrome DevTools Console, "isPlaying" is always "false".

**Solution**
Change playButtonSelector for pauseButtonSelector, and icon-name from "play" for "pause".

[web-scrobbler-amazon-fix.zip](https://github.com/web-scrobbler/web-scrobbler/files/5716055/web-scrobbler-amazon-fix.zip)

### DIFF
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -50,7 +50,7 @@ function setupPropertiesForNewPlayer() {
 		return trackLink === '#';
 	};
 
-	Connector.playButtonSelector = 'music-button[icon-name=play]';
+	Connector.pauseButtonSelector = 'music-button[icon-name=pause]';
 }
 
 function setupPropertiesForOldPlayer() {


alexesprit:

Not sure how this PR fixes #2604. The issue poster said about tracks aren't recognized due to applying album name to artist name node.

sunasaji:
Thank you for developing such a good extension!

I'm also not sure if it might be a general solution to all amazon music users.
For example, the first poster said "the problem seems to be gone", but another person still said "the scrobbler not detecting".
So I'd like to know if this fixes all users with the scrobble issue, and do no harm to the other users.

The issue might be dependent on 
-which amazon music site are used
-which browser and version are used
-if new player are used or not

So I'm thinking to ask the following questions to the issue thread #2604. Is it OK?

```
I also having issue with the scrobbler not detecting, and fixed it for my environment.
I want to know if it fixes the issue, and do no harm to the other users.
So could you help testing the sample version and report the result?

-Do you have scrobbling issue? (Yes/No)
-Do you have scrobbling issue with this sample version? (Yes/No)
-Which amazon music site are you using? ()
-Which browser/version are you using? ()
-Are you using new player or old player? (New/Old)
 You can check it by pressing F12 key on the amazon music site, and open console tab, then you can see:
 "Web Scrobbler: Setup connector for the new player" or
 "Web Scrobbler: Setup connector for the old player".

For example, my case is:
-Do you have scrobbling issue? (Yes)
-Do you have scrobbling issue with this sample version? (No)
-Which amazon music site are you using? (amazon.co.jp)
-Which browser/version are you using? (Google Chrome, 87.0.4280.88)
-Are you using new player or old player? (New)
```

alexesprit:
> So I'm thinking to ask the following questions to the issue thread #2604. Is it OK?

Sure.

I have no access to Amazon, so unfortunately I cannot test your build. Probably, other users could.

Issue #2557, opened 2020-09-14
Add per-website URL pattern blacklist


fishstik:
**Is your feature request related to a problem? Please describe.**
I want to disable scrobbling only for certain YouTube channels. I will frequently forget to skip the track or disable the connector and non-music scrobbles will appear in Last.fm.

**Describe the solution you'd like**
In addition to the URL patterns feature that already exists, add a URL-pattern blacklist, so that URL patterns can be defined that will prevent scrobbling.

**Describe alternatives you've considered**
I do not want to use the `Scrobble videos from "Music/Entertainment" category` option because there are other channels I would like to scrobble as well. It makes sense to implement this feature in this way since the URL pattern matching feature already exists. Open to other alternatives.

**Additional context**
To solve my problem (always skip certain YouTube channels), a userscript that adds the YouTube channel ID to the URL is required (I am using [this one](https://greasyfork.org/en/scripts/392590-whitelist-youtube-channels-ads)). I am also using this script to whitelist certain channels for uBlock Origin.

alexesprit:
Duplicate of #2557.

Issue #2443, opened 2020-06-07
Option to not scrobble anything from a youtube channel


stephanebruckert:
**Is your feature request related to a problem? Please describe.**
I use to watch videos from the youtube channel of a radio which videos are not music, but are still categorized as "music". These end up being scrobbled and I always need to unscrobble them one by one.

**Describe the solution you'd like**
Can we have another icon "Never scrobble this channel" and have this information persisted between sessions?

![Screenshot 2020-06-07 at 20 14 18](https://user-images.githubusercontent.com/1932338/83977877-89c70280-a8fb-11ea-8877-1565e0bc287e.png) [new icon here]

**Describe alternatives you've considered**
I wish there was a way to ban an artist (aka faulty youtube channel name) on last.fm but it doesn't seem possible 

I guess this is potentially a lot of work, but would be super useful as I always come back to that channel

PR #2678, opened 2020-12-21
Implement URL pattern match blacklisting feature

fishstik:
Implements #2557 and partially addresses #2443 (see note 3).

Extends the existing custom patterns feature by marking each pattern in the `customPatterns` storage entry with a flag. When retrieving a connector based on the URL, this flag is used to check if the URL should be ignored.

The user-facing component includes a new body in the connector options modal dialogue, mirroring the existing pattern input. 

**Notes**
1. Blacklist entries take precedent over regular entries.
2. For the prompt, a new string is added to `messages.json` which will need to be translated.
3. To fully address #2443 requires a userscript which retrieves the channel ID of the playing YouTube video and appends it to the URL. The channel ID can then be added to the blacklist to blacklist all videos from that channel. Here is a link to a userscript that can be used with Tamper/Violentmonkey: https://github.com/fishstik/youtube-channel-in-url/
4. Blacklisting URLs in this way will have the extension icon show the `...does not support this website` tooltip and popup. For UI clarity, when a URL is blacklisted, it should show the `... is supported ... but you disabled it` tooltip and perhaps a new popup with something like `This URL is ignored` (since the existing `This website is turned off` popup wouldn't be accurate).

**Additional context**

Connector options UI update. This particular entry blacklists https://www.youtube.com/watch?v=dQw4w9WgXcQ
image

Contents of browser storage after submitting the above:
image

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -346,6 +346,10 @@
     "message": "Add URL patterns that the connector should match below.",
     "description": "Popup hint"
   },
+  "customBlacklistPatternsHint": {
+    "message": "Add URL patterns that the connector should ignore below.",
+    "description": "Popup hint"
+  },
   "customPatternsAdd": {
     "message": "Add pattern",
     "description": "Add button"
--- a/src/core/background/storage/browser-storage.js
+++ b/src/core/background/storage/browser-storage.js
@@ -16,9 +16,18 @@ define((require) => {
 	/**
 	 * This storage contains custom URL patterns defined by an user.
 	 *
+	 * Each pattern is a dictionary containing the pattern and a boolean
+	 * indicating if it should be used to blacklist instead.
+	 *
 	 * The format of storage data is following:
 	 * {
-	 *     connector_id: [URL_pattern_1, URL_pattern_2, ...],
+	 *     connector_id: [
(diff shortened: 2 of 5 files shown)


alexesprit:

I refactor and reorganize the extension code (see `feature/v3` and `feature/arch` branches), and these changes are incompatible with these branches. Unfortunately, I can't say when I finish them.

By the way, is this feature supposed to be used specifically for YouTube?

fishstik:
My personal use case is blacklisting certain YouTube channels from being scrobbled. However this URL pattern blacklisting can be used for any connector. For example, you can blacklist any [Dream Catalogue](https://dreamcatalogue.bandcamp.com/) album on Bandcamp by adding `*://dreamcatalogue.bandcamp.com/*` to the Bandcamp connector.

Would it be helpful to try and implement this feature in one of your refactoring branches? Or should I hold off until you finish those?

fishstik:
Another note is that blacklisting URLs in this way will have the extension icon show the `...does not support this website` tooltip and popup. Showing the `... is supported ... but you disabled it` tooltip and perhaps a new popup with something like `This URL is ignored` (since the existing `This website is turned off` popup wouldn't be accurate) when blacklisted.

Issue #2608, opened 2020-10-24
Option to automatically convert latest comma in artists field into ampersand


sc113:
I.e. scan artist field for the rightmost ", " and replace it to " & " automatically. 

It would help in converting spotify\deezer tags into last.fm standards automatically, that I usually do manually.

inverse:
Sound related to #1552

alexesprit:
Not sure if it's related to #1552.

I think this could be resolved by using a global transformation enabled by default for all websites.

AFAIK, the most common format is `Artist — Track (feat. Artist 2, Artist 3, ...)`, whereas Spotify uses `Artist, Artist 2, Artist 3 — Track` or even `Artist, Artist 2, Artist 3 — Track (feat. Artist 2, Artist 3, ...)`.

alexesprit:
By the way, we can just search for comma in the artist field directly: some artist names contain comma, e.g. "Time, The Valuator", "Invent, Animate".

Spotify (and some other websites) uses different DOM elements for each artist, so we can get an array of artist names from such websites. Using an array instead of a raw string to transform track info is much robust.

sc113:
Yeah, that's would be much better, and probably could be implemented for all the major streaming services.
Things would be quite less complicated, if last.fm moved towards implementing multiply artists for single scrobble. There is [discussion started on that on last.fm forums](https://support.last.fm/t/allow-multiple-artists-on-the-same-scrobble/4435), but unfortunately without any reaction from last.fm stuff so far.

alexesprit:
> Spotify (and some other websites) uses different DOM elements for each artist, so we can get an array of artist names from such websites. Using an array instead of a raw string to transform track info is much robust.

We use [MediaSession](https://developer.mozilla.org/en-US/docs/Web/API/Media_Session_API) API for Spotify and other websites, and this API [does not support multiple artists](https://developer.mozilla.org/en-US/docs/Web/API/MediaMetadata/artist). The data received via API has a higher priority over data from DOM, so we should handle this somehow too.

sc113:
Just bumping, as still hoping for an option to separate the last artist with an ampersand. I'm getting bloody tired of adding "&" to every track on Spotify.

Issue #2588, opened 2020-10-08
Option for scrobbling only the first/main artist


fmontesino:
**Describe the solution you'd like**
It'd be nice to have a configuration checkbox (or similar) to strip all secondary artists (after the first comma).
This can give some people (like me haha) a cleaner scrobble, as last.fm doesn't recognise Artist 1, Artist 2 a scrobble for Artist 1

alexesprit:
See my comment: https://github.com/web-scrobbler/web-scrobbler/issues/2608#issuecomment-720393528. I think it's a better way to resolve the issue (I hate new options).

flickur:
The thing is that just using the first artist in Spotify would result in more correct tags 99% of the time. Spotify themselves ignore the other credited artists in their track display since it's correct that way, as seen on the screenshot. If there's need to mention the featuring artist on the song name, it's usually already there on Spotify.

![image](https://user-images.githubusercontent.com/40724723/106357144-f371a200-630c-11eb-8beb-eb9e0f56fcf5.png)

PR #2716, opened 2021-01-24
Updated TFM Connector

NinjaLabs-Dev:
Added in connector filter for TFM, removing any artists after the first `,` in string

**Additional context**
Evidence of testing:

Image


alexesprit:
Why? Other connectors return multiple artists.

NinjaLabs-Dev:
Last.FM only tracks the first artist. 

For example, 

This song by deadmau5 has Wolfgang Gartner as a second artist, however, Last.FM only tracks the song as Channel 43 - deadmau5

Seen here: https://www.last.fm/music/deadmau5/_/Channel+43+-+Radio+Edit 

Thus, while yes it tracks the song, they track invalidly as a 'new' song

alexesprit:
I'll accept this, though I want to handle it globally to keep consistent behavior of every connector.

Ref #2588.
Ref #2608.

alexesprit (approved):

src/connectors/truckersfm.js
```diff
@@ -11,3 +15,10 @@ Connector.trackSelector = '.player-title-text';
 Connector.playButtonSelector = '.pause-btn > .fa-play';
 
 Connector.onReady = Connector.onStateChanged;
+
+Connector.applyFilter(filter);
+
+function formatArtists(text) {
+	const artist = text.split(',');
```
alexesprit, line 22: I forgot to mention this way won't work for artists who have a comma symbol in their names.

NinjaLabs-Dev: Maybe I guess, honestly, I've never seen an artist with a comma but I guess in the name of redundancy.

Maybe we make it match with space as the layout is like Artist1, Artist2. So we just use .split(', ')

Not sure if this would account for the space though.

alexesprit: > I've never seen an artist with a comma but I guess in the name of redundancy.

I have few examples of real names containing comma character (see https://github.com/web-scrobbler/web-scrobbler/issues/2608#issuecomment-731719288 for details).

> Maybe we make it match with the space as the layout is like Artist1, Artist2. So we just use .split(', ')

Using `, ` as a split symbol won't resolve this. In other connectors I use the following approach:
https://github.com/web-scrobbler/web-scrobbler/blob/d3a369b18d908deb375a763ca83e43ddb92a47c4/src/connectors/monstercat.js#L34-L42

Not sure if TruckersFM website provides artists as separate nodes, though.

Issue #2700, opened 2021-01-06
Add СберЗвук


sabinich:
Hi, alexspirit! Thank you for this extension for oldfags like me ;)
Could you add a connector for yet another music service at the Sberbank ecosystem СберЗвук?

https://sber-zvuk.com

PR #2735, opened 2021-02-12
Add support for sber-zvuk.com

UnQuaiz:
This request is going to add support for SberZvuk (sber-zvuk.com), audio streaming service, by setting up a new connector.

**Additional context**
RE: #2700

### DIFF
--- a/src/connectors/sber-zvuk-dom-inject.js
+++ b/src/connectors/sber-zvuk-dom-inject.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const getTrackInfo = (track, currentTime) => {
-	const { title, credits, duration, id, image, release_title } = track;
+	const { title, credits, duration, id, image, release_title: album } = track;
 	const trackArt = image.src.replace('{size}', '400x400');
 
 	return {
@@ -11,13 +11,13 @@ const getTrackInfo = (track, currentTime) => {
 		duration,
 		trackArt,
 		currentTime,
-		album: release_title,
+		album,
 	};
 };
 
 const queueEvent = (event) => {
 	let type = null;
-	let trackInfo, queue;
+	let trackInfo; let queue;
 
 	if (event.type === 'update') {
 		switch (event.name) {
@@ -45,7 +45,7 @@ const queueEvent = (event) => {
 		window.postMessage({
 			sender: 'web-scrobbler',
 			type,
-			trackInfo: trackInfo,
+			trackInfo,
 		}, '*');
 	}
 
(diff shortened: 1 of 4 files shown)


src/connectors/sber-zvuk-dom-inject.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+const getTrackInfo = (track, currentTime) => {
```
alexesprit, line 3: Could you use function declarations?

src/connectors/sber-zvuk-dom-inject.js
```diff
@@ -0,0 +1,58 @@
...
+		currentTime,
+		album,
+	};
+};
+
+const queueEvent = (event) => {
+	let type = null;
+	let trackInfo; let queue;
+
+	if (event.type === 'update') {
```
alexesprit, line 22: You can check for `event.type` first, and return earlier, if it's not equal "update". This allows to remove check if `type` (not event type) is not nullable.

src/connectors/sber-zvuk-dom-inject.js
```diff
@@ -0,0 +1,58 @@
...
+			case 'currentTime':
+			case 'current_idx':
+				queue = event.name === 'current_idx' ? event.object : event.object.Queue;
+				trackInfo = getTrackInfo(queue.currentTrack, queue.Player.currentTime);
+				type = 'currentTime';
+				break;
+			case 'buffered':
+			case 'isLoading':
+			case 'initialPlaying':
+			case 'manually':
```
alexesprit, line 38: Are these cases required? I think we can omit them.

src/connectors/sber-zvuk-dom-inject.js
UnQuaiz: sure

src/connectors/sber-zvuk-dom-inject.js
UnQuaiz: event.type is not nullable. `type` filled by mobx observer, and always has some type of action. We interested in `update` action, when some property have changed. I'll fix checking.

src/connectors/sber-zvuk-dom-inject.js
UnQuaiz: Yep. Forgot to cleanup

src/connectors/sber-zvuk-dom-inject.js
alexesprit: > event.type is not nullable

I didn't say it's nullable. :)

My idea was to remove check for `type` value (which is from `mobx` observer). The `type` value is null, if we received an event with `event.type` != "update", so I suggested to check for `event.type` first, and return if it's not equal "update".

UnQuaiz: already done :)

alexesprit (approved):

alexesprit:
👍🏻

Issue #2747, opened 2021-03-04
Artist and album not being scrobbled from YouTube Music


Mumumurilo:
I have just noticed this issue on Windows. All tracks I've been listening to are not being recognized due to the scrobbler only picking up the song name and album cover.

![image](https://user-images.githubusercontent.com/8595653/109894689-b3d00a00-7c6c-11eb-8840-5794b4bbd031.png)

I've used the extension on macOS during the whole day with no issues, so I believe it may be related to the OS (Windows).
I'm on Chrome ver. 88.0.4324.190.

alexesprit:
I have the same issue. Looks like they updated website's internals, and now scrobbling is broken for everything.

alexesprit:
I have the same issue. Looks like they updated website's internals, and now scrobbling is broken for everyone.

Dare2python:
+1

G4Zz0L1:
Yep, same thing on Chrome on Arch Linux. It doesn't involve only windows.

olexandr-perepiyaka:
I have the same issue.

alexesprit:
I see. I'll fix it when I come home.

mpdgraev:
Fwiw:

I fixed it in a local build with 
artistSelector:  '.ytmusic-player-bar.byline .yt-formatted-string:first-child'
albumSelector:  '.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href^="/channel/MPREb_"]'

Seems to work well, but don't have time to test it right now whether or not this breaks other things (I haven't used jquery in years).

alexesprit:
It should not break anything. Note that there're also selectors to handle uploaded songs, and I guess they should be updated too. I have no uploaded tracks atm, so I cannot check it.

mpdgraev:
Ah, I didn't know that the other selectors were for user uploaded tracks, using the x-th child approach has the advantage that both ytm tracks and user uploaded tracks both work without each needing 2 selectors, which is a bit easier on the eye imo.

Created a pull request since I had the code ready anyway now.

alexesprit:
I submitted the new release (v2.41.0) which also includes fix by @mpdgraev. It will be available in store once this version will be reviewed.

When I was writing this comment, it was approved by AMO, and it's now available on Addons.mozilla.org.

PR #2748, opened 2021-03-04
fix youtube music artist and album parsing

mpdgraev:
Fixes the youtube music album and artist parsing which broke yesterday.

Fixes https://github.com/web-scrobbler/web-scrobbler/issues/2747

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const trackArtSelector = '.ytmusic-player-bar.image';
-const artistSelectors = '.ytmusic-player-bar.byline .yt-formatted-string:first-child';
 
 const artistSelectors = [
 	'.ytmusic-player-bar.byline [href*="channel/"]:not([href*="channel/MPREb_"])',
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -2,8 +2,7 @@
 
 const trackArtSelector = '.ytmusic-player-bar.image';
 const artistSelectors = [
-	'.ytmusic-player-bar.byline [href^=channel]',
-	'.ytmusic-player-bar.byline [href*="browse/FEmusic_library_privately_owned_artist_detaila_"]',
+	'.ytmusic-player-bar.byline .yt-formatted-string:first-child',
 ];
 
 const trackSelector = '.ytmusic-player-bar.title';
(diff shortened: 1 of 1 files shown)


alexesprit:
I'll test it in few hours.

alexesprit:
I see a lot of errors when using the connector from this PR:
```
youtube-music.js:64 Uncaught TypeError: artistElements.toArray is not a function
    at getArtists (youtube-music.js:64)
    at BaseConnector.Connector.getArtistTrack (youtube-music.js:47)
    at BaseConnector.getCurrentState (connector.js:765)
    at BaseConnector.stateChangedWorker (connector.js:683)
    at MutationObserver.BaseConnector.onStateChanged (connector.js:509)
```

src/connectors/youtube-music.js
```diff
@@ -23,10 +20,7 @@ Connector.getTrackArt = () => {
 	return null;
 };
 
-Connector.albumSelector = [
-	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/MPREb_"]',
-	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]',
-];
+Connector.albumSelector = '.ytmusic-player-bar.byline .yt-formatted-string:nth-child(3)';
```
alexesprit, line 23: These selectors don't work properly with multiple artists.
<details>
  <summary><b>Screenshot </b> (Click to expand)</summary>

image

</details>

mpdgraev: Ah yes, I don't have any music with multiple artists so didn't realize they split up the names like that.

Guess the href filter would be required then. Problem is the new urls for album both start with with /channel, so that's annoying unless jquery allows something like negative forward searches so artist can be /channel/(?!=uMPREb)

alexesprit:
I suggest to use the following selectors:
```js
const artistSelectors = [
	'.ytmusic-player-bar.byline [href*="channel/"]:not([href*="channel/MPREb_"])',
	'.ytmusic-player-bar.byline [href*="feed/music_library_privately_owned_artist_detaila_"]',
];
const albumSelectors = [
	'.ytmusic-player-bar [href*="channel/MPREb_"]',
	'.ytmusic-player-bar [href*="feed/music_library_privately_owned_release_detailb_"]',
];
```

@mpdgraev What do you think?

mpdgraev:
That seems like it should do the trick. Lemme update.

src/connectors/youtube-music.js
```diff
@@ -23,9 +24,9 @@ Connector.getTrackArt = () => {
 	return null;
 };
 
-Connector.albumSelector = [
-	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/MPREb_"]',
-	'.ytmusic-player-bar .yt-formatted-string.style-scope.yt-simple-endpoint[href*="browse/FEmusic_library_privately_owned_release_detailb_"]',
+const albumSelectors = [
```
alexesprit: You still should assign selectors to the `Connector.albumSelector` property. You can assign it right here (as it was before), though I would place it at the top, so it would be like:
```js
const artistSelectors = [
    ...
];
const albumSelectors = [
    ....
];

...

// The base connector will use them to get album info
Connector.albumSelector = albumSelectors;

```

alexesprit (approved):

alexesprit:
Cool, thanks!

PR #2750, opened 2021-03-04
Fix URL for Hype Machine album premieres.

jaccarmac:
I noticed the URL of the [latest premiere](https://hypem.com/album/alostmen) changed.

Looking at the [list of albums](https://hypem.com/albums), it seems like all of them have changed in the same way.

Would it be best to change the name of the connector to "Hype Machine Albums"?


alexesprit:
> Would it be best to change the name of the connector to "Hype Machine Albums"?

Sure!

src/core/connectors.js
```diff
@@ -318,9 +318,9 @@ const connectors = [{
 	js: 'connectors/tidal.js',
 	id: 'tidal',
 }, {
-	label: 'Hype Machine Premieres',
+	label: 'Hype Machine Featured Albums',
 	matches: [
-		'*://hypem.com/premiere/*',
+		'*://hypem.com/album/*',
```
alexesprit, line 323: I think it should be updated too.

src/core/connectors.js
```diff
@@ -318,9 +318,9 @@ const connectors = [{
...
 	],
 	js: 'connectors/hypem-premieres.js',
```
alexesprit, line 325: I think it should be updated too.

jaccarmac: Just the filename or the `id` as well? If `id` is changed, should I write a migration to keep options set for it?

alexesprit: No, I meant just the filename. ID should be kept as is for compatibility.

alexesprit (approved):

PR #2760, opened 2021-03-07
Fix archive connector

### LINKED ISSUE #2759: Internet Archive connector does not work properly with certain URLs
**Describe the bug**
The connector does not recognize some albums/single tracks correctly.

- https://archive.org/details/dont-lie-beets-produce-ny-mix
**Actual behavior**: Doesn't recognize anything and spams errors to console.

- https://archive.org/details/lp_everybody-knows-this-is-nowhere_neil-young-crazy-horse-robin-lane
**Actual behavior**: Keeps "Artist-Track" as a track title.

Also, the connector does not use track art from certain pages, e.g. from [this one](https://archive.org/details/lp_everybody-knows-this-is-nowhere_neil-young-crazy-horse-robin-lane).

<details>
  <summary><b>Debug log</b> (Click to expand)</summary>

```
Uncaught TypeError: Cannot read property 'split' of null
    at BaseConnector.Connector.getArtistTrack (archive.js:50)
    at BaseConnector.getCurrentState (connector.js:790)
    at BaseConnector.stateChangedWorker (connector.js:706)
    at MutationObserver.BaseConnector.onStateChanged (connector.js:530)
```
</details>

**How to reproduce**
1. Go to one of URLs I listed above.
2. Click on the play button.

**Expected behavior**
The extension should get the proper information from these pages.

-   Extension version: v2.41.0

alexesprit:
Should fix everything I described in #2759.

### DIFF
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -57,7 +57,13 @@ Connector.isPlaying = () => {
 
 Connector.playerSelector = '#theatre-ia';
 
-Connector.trackArtSelector = '.album-cover img';
+Connector.trackArtSelector = [
+	// https://archive.org/details/AH013_sarin_sunday_-_the_lonely_hike
+	'#theatre-ia center > img',
+
+	// https://archive.org/details/lp_everybody-knows-this-is-nowhere_neil-young-crazy-horse-robin-lane
+	'.album-cover img',
+];
 
 Connector.getTrackInfo = () => {
 	const artist = getArtists(artistSelectors);
--- a/src/connectors/archive.js
+++ b/src/connectors/archive.js
@@ -34,7 +34,7 @@ const filter = MetadataFilter.createFilter({ track: removeNumericPrefixes });
 Connector.applyFilter(filter);
 
 function removeNumericPrefixes(track) {
-	if (hasAllTracksNumericPrefix(tracksSelector)) {
+	if (hasAllTracksNumericPrefix()) {
 		return track.trim().replace(/^(\d+\w+)/, '');
 	}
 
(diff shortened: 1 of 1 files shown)


alexesprit:
@inverse I'm not familiar with the connector as you are, could you please review the PR?

src/connectors/archive.js
```diff
@@ -1,18 +1,34 @@
...
+ *
+ * https://archive.org/details/lp_everybody-knows-this-is-nowhere_neil-young-crazy-horse-robin-lane
+ * Full album with artist suffixed in track names
+ *
+ * https://archive.org/details/dont-lie-beets-produce-ny-mix
+ * Single track
+ */
+
 const artistSelectors = [
 	'.key-val-big a span',
```
alexesprit, line 20: @inverse What this and `.jwrowV2.playing .ttl` selectors are for? I haven't find anything where they could be used.

inverse: Looks like it's changed to 

```
const tracksSelector = '.jwlistV2 .ttl';
```

inverse: I can see it on https://archive.org/details/AH003_corwin_trails_-_corwin_trails/Oasis_is_Myth.mp3

It looks like that's broken though due to the extra whitespace. 

We could either update line 97 to `if (!numericTrackRegex.test(track.textContent.trim())) {` or relax the regex matching.

alexesprit: Yes, I noticed two types of players on archive.org, where `.jwrowV2.playing .ttl` and `.audio-track-list .selected .track-title` are used as *track title* selectors. But I still can't see anything that match the `.key-val-big a span` selector.

alexesprit: > It looks like that's broken though due to the extra whitespace.

Interesting, it was working fine when I tested changes. :)

inverse: I wonder why it's behaving differently...

alexesprit:
@inverse have you any other suggestions?

src/connectors/archive.js
```diff
@@ -41,39 +60,54 @@ Connector.isPlaying = () => {
...
+		if (!numericTrackRegex.test(track.textContent.trim())) {
 			hasAllTracksNumericPrefix = false;
 			break;
 		}
 	}
 
 	return hasAllTracksNumericPrefix;
 }
+
+function getArtists(selectors) {
```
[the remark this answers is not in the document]
alexesprit: It gets text from all nodes (artist nodes) and returns a string of text values joined with comma.

For example
```
<a href="..." class="artist">Artist 1</a>
<a href="..." class="artist">Artist 2</a>
```

`getArtists('.artist')` will return "Artist 1, Artist 2" which is more common (vs raw "Artist 1; Artist 2" joined with semicolon, which is provided by archive.org).

For now it's copy-pasted from other connectors; though it should be extracted as an utility function.

In future we can send an array of artists to the background script for easier manipulation instead of sending a string.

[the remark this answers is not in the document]
inverse: Ahh so the background handles the extraction of the artist from that CSV?

PR #2767, opened 2021-03-12
Added Connectors for CBC Music, Indie88 and Stingray Music

jbwharris:
Added new connector for three Canadian services, CBC Music, Indie88 and Stingray Music

**CBC Music**
Only works for Music Playlists and not their streaming radio, since their web streams don't have meta data.

**Indie88**
Works with the popup player (https://cob.leanplayer.com/CINDFM) and the full page player (https://indie88.com/lean-stream-player/)

**Stingray Music**
This one works well if you're using an ad blocker like uBlock Origin, but I had noted issues with being a track behind when it did ad insertion that I wasn't able to figure out.


alexesprit:
Could you submit each connector in a separate PR?

jbwharris:
@alexesprit I will be able to submit them separately tonight. How do I best handle connectors.js additions for each PR?

inverse:
@jbwharris Git should be able to handle that 👍🏼 There might be minor conflicts but shouldn't be so bad.

alexesprit:
@jbwharris you can use separate branches per each connector. I'll resolve merge conflicts by myself. :)

PR #2777, opened 2021-03-19
Add Indie88 Connector

jbwharris:
Created a connector for Toronto's Indie88

**Additional context**

Indie88
Works with the popup player (https://cob.leanplayer.com/CINDFM) and the full page player (https://indie88.com/lean-stream-player/). The web player's meta data is inverted, so I have intentionally inverted the data. If they ever fix this it'll have to be reverted.


src/connectors/indie88.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+// Indie88 has the track and artist flipped on their main web player. The values are intentionally flipped in this Connector
+
+Connector.playerSelector = 'div.cobrp-player-main, div.mediaplayer-mainbody';
+
+Connector.artistSelector = 'div.cobrp-player-main div.cobrp-player-artist, div.mediaplayer-mainbody .mediaplayer-snp p#track-info-title span';
```
alexesprit, line 7: You could use array of selectors here for more readability and safer behavior:
```js
Connector.artistSelector = [
	'selector-1',
	'selector-2',
];
```

Also, these selectors could be simplified, e.g.: 
`div.cobrp-player-main div.cobrp-player-artist` -> `.cobrp-player-artist`

src/connectors/indie88.js
```diff
@@ -0,0 +1,11 @@
...
+
+Connector.trackSelector = 'div.cobrp-player-main .cobrp-player-song, div.mediaplayer-mainbody .mediaplayer-snp p#track-info-artist span';
+
+Connector.onReady = Connector.onStateChanged;
```
alexesprit, line 11: Does the connector works fine w/o this line? This is a workaround for some websites, and it should not be used everywhere.

src/connectors/indie88.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+// Indie88 has the track and artist flipped on their main web player. The values are intentionally flipped in this Connector
```
alexesprit, line 3: It's good you mentioned this. :) 👍🏻

src/connectors/indie88.js
jbwharris: I just tested and it does seem to work without this line. Should I push that update to the PR?

src/connectors/indie88.js
```diff
@@ -0,0 +1,18 @@
...
+
+Connector.playerSelector = [
```
alexesprit, line 5: The `playerSelector` property supports strings only.

Ideally, I would split the connector; for now you could use "setup" pattern (for example, https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/douban-artists.js).

alexesprit (changes requested):
1. Non-relevant connector should be removed from the PR.
2. ESLint errors should be fixed.
3. Please check my note about the `playerSelector` property.

jbwharris:
@alexesprit I've pushed all the changes to this connector.

jbwharris:
@alexesprit I am unclear as to what I still have outstanding as far changes. I just added the play/pause states. Hope to get to the Stingray and CBC changes this weekend.

src/connectors/indie88.js
```diff
@@ -0,0 +1,34 @@
...
+function setupConnector() {
+	if (isCobrPlayer()) {
+		setupCobrPlayer();
+	} else {
+		setupWebsitePlayer();
+	}
+}
+
+function isCobrPlayer() {
+	return $('article.cobrp-page').length > 0;
```
alexesprit, line 12: We don't use jQuery in new code.

alexesprit (approved):

PR #2789, opened 2021-03-24
Fix selectors for the Tidal redesign.

### LINKED ISSUE #2787: Tidal not scrobbling
Tidal web just updated and the extension broke on the site. The extension icon shows in red

Logs: https://pastebin.com/jNeT8tRc

-   OS: Windows 7
-   Browser: Opera
-   Extension version: 2.41.0

jaccarmac:
Closes #2787.

### DIFF
--- a/src/connectors/tidal.js
+++ b/src/connectors/tidal.js
@@ -2,13 +2,13 @@
 
 Connector.playerSelector = '#footerPlayer';
 
-Connector.playButtonSelector = `${Connector.playerSelector} [data-type="button__pause"]`;
+Connector.playButtonSelector = `${Connector.playerSelector} button[data-test="play"]`;
 
-Connector.pauseButtonSelector = `${Connector.playerSelector} [data-type="button__play"]`;
+Connector.pauseButtonSelector = `${Connector.playerSelector} button[data-test="pause"]`;
 
 Connector.isScrobblingAllowed = () => !!$(Connector.playButtonSelector);
 
-Connector.trackSelector = `${Connector.playerSelector} [data-test="footer-track-title"]`;
+Connector.trackSelector = `${Connector.playerSelector} div[data-test="footer-track-title"]`;
 
 Connector.getUniqueID = () => {
 	const trackUrl = $(Connector.trackSelector).attr('href');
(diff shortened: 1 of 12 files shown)


src/connectors/tidal.js
```diff
@@ -1,42 +1,31 @@
 'use strict';
 
-let useShortTrackNames = false;
```
jaccarmac, line 3: One of the big changes is that Tidal seems to no longer provide longer and shorter track names. If this is confirmed and the feature needs removing, I can add another commit to delete the option from the options page.

src/connectors/tidal.js
```diff
@@ -1,42 +1,31 @@
...
 		return trackUrl.split('/').pop();
 	}
 	return null;
 };
 
-Connector.artistSelector = `${Connector.playerSelector} [class^="mediaArtists"]`;
+Connector.artistSelector = `${Connector.playerSelector} .artist-link`;
 
-Connector.albumSelector = `${Connector.playerSelector} [class^="infoTable--"] a[href^="/album/"]`;
+Connector.albumSelector = ['#nowPlaying [class^="infoTable--"] a[href^="/album/"]', `${Connector.playerSelector} a[href^="/album"]`];
```
jaccarmac, line 23: The album information is no longer available in the info box. In fact, it's worse than it was when I first implemented the feature: The album name is only available if the expanded now playing box is open _and_ the credits tab is active. What I settled on is using the info table if it is available, but falling back to the "Playing From", though we ensure that the source is an album instead of a playlist. Let me know if the selector for that is overbroad.

jaccarmac:
It only occurred to me after I signed off, but since we have the track URL, we could insert a fetch step to read the album page (similar to how the Gaana connector works). This would allow:

1. Reading the album name from one place.
2. Reading the album artist.
3. Finding the long title to keep that option around (though I prefer dropping this feature).

Do you think it's worth paying the cost of a page fetch?

alexesprit:
> Do you think it's worth paying the cost of a page fetch?

Well, I have no strong opinion on this. That `fetch` workaround should be fine, though keep in mind you should manage state change manually.

jaccarmac:
> keep in mind you should manage state change manually

Is this so there's only one fetch per song, or is there some other reason to manage it?

alexesprit:
> Is this so there's only one fetch per song, or is there some other reason to manage it?

Yup, otherwise you'll spam the website with a lot of requests.

jaccarmac:
Don't think I will. Based on experimentation, it looks like the text of the `fetch` doesn't contain the title directly and relies on script execution. `tidal.com/browse` URL's might be different, but this use case seems more complicated than Gaana or even YouTube.

alexesprit:
@jaccarmac So, the PR is ready to merge?

jaccarmac:
I'd like to take away the short track name option since it's no longer relevant, and double check my selectors. But since there's been no feedback, I'll finish it up ASAP and then we can merge.

jaccarmac:
That should do it if it lints.

src/connectors/tidal.js
jaccarmac: I tried narrowing this a bit. It seems like there's no other good way to do it.

alexesprit (approved):

Issue #2643, opened 2020-11-27
Add a connector for Live365.com


thomwim:
**Live365.com**
Live365 is a website that has user-created radio stations. Here is a sample station.

https://live365.com/station/Third-Rock-a76979

**Additional context**
For the metadata on the website:
player light-on-dark
media-info
track-name
track-artist
media-art player-image
play-button

**Ignore:**
Live365 : 30 Second Ad Break, 60 Second Ad Break, 90 Second Ad Break, 120 Second Ad Break

tape-:
Seconding. I listen to this one often. http://player.live365.com/a23389

spudart:
Does live365 continue to scrobble for others? It doesn't scrobble for me. I'm listening to: https://live365.com/embed/popout.html?station=a10277&s=md&m=dark&c=mp3

PR #2801, opened 2021-03-28
Add Live365.com

willfolsom:
Add Live365.

Handles https://github.com/web-scrobbler/web-scrobbler/issues/2643.

The current PR works, but as mentioned in the issue, Live365 has ads. I've stubbed out an alternate version of live365.js that looks like this:

```
'use strict';

const LIVE_365_ARTIST = 'Live365';
const ADWTAG = 'ADWTAG';

Connector.playerSelector = '.player';
Connector.trackArtSelector = '.player-image';
Connector.trackSelector = '.track-name';

Connector.isPlaying = () => Util.hasElementClass('.playing-indicator', 'is-playing');

Connector.getArtist = () => {
	let artist = Util.getTextFromSelectors('.track-artist');

	if (artist && (artist === LIVE_365_ARTIST || artist.includes(ADWTAG))) {
		artist = ''; // can also be null
	}

	return artist;
};
```

This handles set the artist to either empty string or null if it is a recognized ad, and could also check the title if it contains "ADBREAK" (edit: or "Advertisement" based on the last.fm link below).

I could even try to handle it such that Connector.isPlaying is set to false whenever an ad is playing (at least somewhat more elegant).

All that said, I'm not sure of the protocol for handling ads from the extension vs letting the user handle it. I know I've had to delete scrobbled ads from my last.fm profile before (if they slip through ublock).

Edit: Looks like ads are already getting scrobbled from another resource: https://www.last.fm/music/Live365


alexesprit:
Usually we skip ads with `Connector.isScrobblingAllowed` function. When ad is playing, this function should return false, and vice versa.

willfolsom:
Live365 not allowed

Live365 allowed

alexesprit (approved):

alexesprit (approved):

thomwim:
Live365 just updated its website, so the scrobbler no longer recognizes the currently playing song.

https://live365.com/station/Metal---Rock-Mix-a32028

Issue #2803, opened 2021-03-28
Add EulerBeats


rstuven:
EulerBeats is a algorithmically generated art+music project. https://eulerbeats.com

PR #2804, opened 2021-03-28
Add EulerBeats connector

rstuven:
Handles #2803

### DIFF
--- a/src/connectors/eulerbeats.js
+++ b/src/connectors/eulerbeats.js
@@ -1,0 +1,41 @@
+'use strict';
+
+Connector.playerSelector = '#root';
+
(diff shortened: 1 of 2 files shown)


src/connectors/eulerbeats.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+Connector.playerSelector = '#root';
+
+Connector.isPlaying = () => {
+	return $('.eulerbeats-main.eulerbeats-is-playing').length !== 0;
```
alexesprit, line 6: Please don't use jQuery. It will be removed.

alexesprit (approved):

Issue #2829, opened 2021-04-22
Override general settings for specific connectors


RudeySH:
General settings like "Notify if track is not recognized" and "Force track recognition" are useful, but they would be even more useful if I could turn them on or off depending on the connector/website. For example, I'd like to disable "Notify if track is not recognized" on YouTube but enable it for most other websites. Another example, disabling "Force track recognition" everywhere except on sites that have better track info such as Bandcamp.

I'd like to copy the 4 checkboxes from the General settings and put them in the dialog that shows up when clicking on the gear icon for each supported website. The checkboxes would be in the "indeterminate" state by default, which means they follow the general settings. You can check any of the boxes per connector to override the general setting for that specific connector. That dialog has a reset button which is perfect for restoring the "indeterminate" state of the checkbox, in case you've overwritten the general settings for a connecter and want to undo it. It's important to note that the dialog would no longer only be used for custom URL patterns, so some of the text on the options page would need to be tweaked.

Another approach could be to replace the checkboxes with drop-down lists that have 3 options. Users are probably more familiair with how drop-down lists work, a checkbox in the "indeterminate" state might be too confusing. The 3 options would be something like this:

1. Unspecified / Follows general setting (this is the default)
2. Enabled  (corresponds to a checked box)
3. Disabled (corresponds to an unchecked box)

In both approaches the general settings themselves would remain untouched.

I want to attempt to make a PR for this myself but if anyone else wants to pick this up, that's fine too. But before I start working on this, I'd like to gauge interest from the maintainers, collaborators or other users. Does the proposal above sound good, or should it be implemented differently? I have some ideas for new general settings that I'd like to implement next, but the ability to have different settings per connector is a must before I want to propose new settings.

PR #2831, opened 2021-04-22
Add connector override options (#2829)

RudeySH:
I added a new type of options, connector override options, which allows users to enable or disable "Use now playing notifications", "Notify if track is not recognized", "Force track recognition" and "Scrobble podcasts" on specific websites. The checkboxes are in the "indeterminate" state by default, which means they're unset and the connector uses the global/general settings. You can check or uncheck these boxes to make the connector no longer follow use global setting. The reset button will restore the checkboxes to the indeterminate state, and will make it so the connector no longer overrides the general settings.

With this feature implemented, users can now for example enable force track recognition on all websites except YouTube, or disable unrecognized track notifications on all sites except Bandcamp (see screenshot below).

image

### DIFF
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -267,6 +267,10 @@
         "message": "Anders als andere Erweiterungen enthält diese keine blinkenden Banner oder hinterhältige Werbung.",
         "description": ""
     },
+    "customPatterns": {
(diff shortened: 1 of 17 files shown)


RudeySH:
Had to resubmit: #2832

PR #2854, opened 2021-05-10
Fix maloja duplicates

ano95:
Send timestamp with every request for Maloja.

Maloja assigns current server timestamp to scrobbles without `time`. That creates duplicated scrobbles when time is not in sync between client and server or when the connection is slow.

See also krateng/maloja#73


ano95:
Thank you for this information, I will remove now playing request from this PR later today.

ano95:
I am not exactly sure how to properly implement this, but this seems to work.

alexesprit (approved):

PR #2920, opened 2021-07-02
Add Thrill Jockey connector.

jaccarmac:
Thrill Jockey provides streaming of all their releases on their website. The
format is inconsistent between pages; For now, the connector works on album
pages. Track names are preceded by their number, which is easily trimmed by
looking for spaces. Track art has relative URL's, so `location.origin` must be
prepended manually.


jaccarmac:
The relative URL problem seems like it might be a case for `Util.normalizeUrl`, but I held off on tweaking that for now as its scope is limited at the moment.

jaccarmac:
Future directions could include additional cases for artist pages, where there is also a player (albeit without as much album data). There are also embedded Vimeo elements for music videos. There may be even more! I'll see as I explore the site.

alexesprit (approved):

PR #2940, opened 2021-07-26
Prefer YouTube video IDs on YouTube Music

LostLuma:

Prefers using video IDs instead of track metadata on YouTube music, as some videos (incorrectly) have the same metadata, which makes it impossible to locally store the correct metadata instead.

**Additional context**

As far as I'm aware this is not a breaking change, as the saved edits model also looks up saved information using metadata from the page if it can't be found right away, however I was unsure whether this should also include a check with all four base fields Song has, meaning I've left the file untouched for now.
https://github.com/web-scrobbler/web-scrobbler/blob/f5007fe38e1472b7f2074454761f8ccbe26a9cf9/src/core/background/storage/saved-edits.model.js#L23-L32

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -98,6 +98,14 @@ Connector.isPlaying = () => {
 	return Util.getAttrFromSelectors(playButtonSelector, 'd') === playingPath;
 };
 
+Connector.getUniqueID = () => {
+	const videoUrl = Util.getAttrFromSelectors('.yt-uix-sessionlink', 'href');
+
+	if (videoUrl) {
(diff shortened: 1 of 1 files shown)


inverse (approved):
Looks good to me 👍🏼

inverse:
Sorry for the slow action :/ Will be in the next release 👍🏼 

Regarding your other question - Feels like it could be extended to make that check, Not sure the rational for the way it is now.

LostLuma:
Thank you! I was just worried the PR would not be seen when closed, I understand not always being motivated to work on projects.

PR #2952, opened 2021-08-09
Add Yammat FM connector

JonatanPlesko:
I've added support for [Yammat FM](https://yammat.fm/).
I managed to build and test it on Firefox, and it works great there!

Yammat FM Firefox

I have a concern, though. I tried building and testing it out in Chrome, however, the extension in Chrome always says "Website not supported". It might be just a local issue with my development setup since I'm not familiar with extension development, but I tried both on a Windows and a Mac machine and couldn't get it to work on either in Chrome. Firefox works fine everywhere.

P.S. I know artist/title selectors seem incorrect, but they're actually swapped in the website 🤷


JonatanPlesko:
After further testing, I noticed another problem - scrobbling keeps going even after the radio is paused. As far as I can tell, play button has just the `.stream-controls__btn-play` class, while pause button has both that one and btn-pause: `.stream-controls__btn-play .btn-pause`. I'm not sure how am I supposed to use the selectors correctly here.

jaccarmac:
> I'm not sure how am I supposed to use the selectors correctly here.

Based on your description of the page, your code should only be picking up the pause button. But what's important to get pausing correct is what changes on the page when a song is paused. In this case, is one button or the other removed, or does an attribute (or property) on one of them toggle?

inverse:
@JonatanPlesko Need any help on this one? or what's left?

JonatanPlesko:
Yes, I still can't figure this out.
If I try with:
`Connector.pauseButtonSelector = '.stream-controls__btn-play .btn-pause';`
The extension doesn't even detect that something is playing at all.

If I use the current code:
`Connector.playButtonSelector = '.stream-controls__btn-play .btn-pause';`
playing is detected, but scrobbling continues even after the player is paused.

To answer @jaccarmac question, it's just an attribute being added when a song is paused (rather than toggled). When the player is playing, the button just has the `.stream-controls__btn-play` class. When it's paused, it has two classes, `.stream-controls__btn-play .btn-pause`.

jaccarmac:
Two things:

1. If the pause button has two classes on it, the selector should contain both classes back-to-back (`.stream-controls__btn-play.btn-pause`). Set up as you have it, with a space between the class names, I'm pretty sure it's looking for an element with `btn-pause` *inside* the element with `stream-controls__btn-play`.
2. If you have `playButtonSelector` and/or `pauseButtonSelector` defined, the default `isPlaying` implementation will only ever check for one element, with the play button taking priority. That's why looking for the play button never detects a pause: Both the pause and play buttons look identical to the connector.

https://github.com/web-scrobbler/web-scrobbler/blob/10c4b662f2663bf0705114c63be596bee3102c67/src/core/content/connector.js#L325-L358

I think that fixing your selector and defining *only* `pauseButtonSelector` should get the connector working correctly. If not, try defining a custom `isPlaying` or inserting some logging in the existing one.

Please forgive me if this is stuff you already know.

inverse:
@JonatanPlesko any progress on this?

JonatanPlesko:
Sorry, @inverse, I did not find the time after the initial struggles and then the PR fell off my radar.

I updated the PR with the fixed implementation. I've been scrobbling Yammat FM the entire day, with occasional pauses as well, and it seems there are no issues present. Scrobbling while playing works great, and during the pause, nothing is being scrobbled (as it was with the previous implementation).

Thanks a lot, @jaccarmac. I was not aware of either behavior, and your assumption that there should only be pauseButtonSelector defined with the fixed selector was correct.

As a sidenote - the issues I originally mentioned in the PR description about the add-on not working in Chrome are not present anymore. I tested the extension in both Chrome and Firefox, seems to work well.

inverse:
Hi @JonatanPlesko sorry for the delay in reviewing :+1:

PR #2960, opened 2021-08-17
Update detection of "various artists" albums in bandcamp connector

thomaslule:
Hi! 👋 

The bandcamp connector currently has false positives when deciding if an album is from various artists.

Examples (these are wrongly detected as "various artists albums" because of how the tracks are named):

https://teenagemenopause.bandcamp.com/album/viens-mourir
https://standard-in-fi.bandcamp.com/album/quattro

In this change I introduced a new rule: if every detected artist of an album has a name shorter than 3 characters, then we decide that those are not artist names.

This should cover the majority of current false positives with extremely few false negatives (if every artist in an album has a name shorter than 3 characters, it wont count as "various artists")


inverse (approved):
Nice one - thanks 👍🏼

Issue #2942, opened 2021-07-27
Add-on doesn't recognize when horizonte.cl is playing


altatransmision:
I already installed the latest extension and it's running ok in other applications on Opera -I tried with Mozilla too-, but the add-on doesn't recognize when I press play on horizonte.cl and, therefore, the scrobbler is not running in this site :(

![image](https://user-images.githubusercontent.com/66893491/127079566-336f0f27-0b3f-4f71-be69-1da5112c3324.png)

I appreciate your consideration by implementing the connector and, as soon as I can, I'm going to donate in honor of the great work you're all doing, folks.

Thanks in advance!

altatransmision:
Hi

mathwar:
There are multiple elements with `.fa-play` class on the page.

I can see in the browser console that `"isPlaying"` does not change it's state when the play button is clicked. With one extra class from the parent element added, for example: `Connector.playButtonSelector = '.btn_control_play .fa-play'`, then  `"isPlaying"` toggles between `false` and `true`, the extension icon turns green and the song info appears correctly.

But I couldn't get it to scrobble to Last.fm anyways. On my profile I can see the current song "Scrobbling now" at the top in Recent Tracks, but when the song is over and the next starts, the scrobble is not saved and only shows the new song as scrobbling now.

mathwar:
I see that the player is constantly reloading its content at short and random time intervals, thus sending the message `Empty status detected - resetting`. Immediately afterwards the same song is detected, sending the message `New song detected` again. The reset is performed several times during the song, resetting the `Song will be scrobbled in X seconds` counter, so it is never scrobbled.

Now, when the `duration` is not set by the `track.getinfo response`  and is therefore set to `30` as this is the default minimum duration, and if with any luck in this short time a reset is not triggered, then the scrobbled is sent.  

It seems to me that it is not a problem with the extension, but a design issue of the player.

inverse:
Will be in the next release 👍🏼

PR #2989, opened 2021-09-18
Fix horizonte.cl connector

mathwar:
There are multiple elements with `.fa-play` class on the page.
To correctly detect the play button status is necessary to add a class 
to make it a unique combination.

Resolves: #2942


inverse:
Thanks 👍🏼

PR #3017, opened 2021-10-11
Add Radio Rock&Pop Chile connector

mathwar:
I added the connector for Rock & Pop Radio Chile.
Also added its entry in connectors.js

R&P is located at 94.1 MHz on the FM dial in Santiago, Chile.


mathwar:
Screenshot from 2021-10-11 23-27-00

mathwar:
I have added "Chile" in the label because there are at least two other "Rock&Pop" radios that I know of: Argentina (FM95.9) and Paraguay (FM95.5)

PR #3048, opened 2021-11-10
Connectors, Yandex.Music: Return a list of artists instead of one

anton-martynov:


inverse:
Hi! thanks for contributing :) why would we want a list of artists returned in such instance?

anton-martynov:
> Hi! thanks for contributing :) why would we want a list of artists returned in such instance?

Full information about the artists. For example https://music.yandex.ru/album/2904571/track/24408035, there are 4 artists but current version of web-scrobbler scrobbles the first one only

inverse:
That URL lands in a 404 for me, I guess because I don't have an account. 

Okays so with your change when the track lists a set of artists it'll map it to a string of `artist-a, artist-b, artist-c` 

Won't this cause more problems since last.fm doesn't really support multiple artists?

anton-martynov:
> Won't this cause more problems since last.fm doesn't really support multiple artists?

Here are examples:
- last.fm: [My Woman by Lew Stone, The Monseigneur Band, Nat Gonella & Al Bowlly](https://www.last.fm/music/Lew+Stone,+The+Monseigneur+Band,+Nat+Gonella+&+Al+Bowlly/_/My+Woman)
- libre.fm: [My Woman by Lew Stone, The Monseigneur Band, Nat Gonella & Al Bowlly](https://libre.fm/artist/Lew+Stone%2C+The+Monseigneur+Band%2C+Nat+Gonella+%2526+Al+Bowlly/track/My+Woman)

anton-martynov:
I have created a new PR, https://github.com/web-scrobbler/web-scrobbler/pull/3079

PR #3055, opened 2021-11-18

mathwar:
seconds), so all songs will be sent at 240 seconds, the[ default maximum 
time](https://github.com/web-scrobbler/web-scrobbler/blob/9b5f5e55a5b68d33e379e016ecf91104bc713e71/src/core/background/util/util.js#L23). This affects songs that are less than 4 minutes long.


mathwar:
Two PR got mixed by mistake

PR #3064, opened 2021-11-28
Add Klassik Radio connector

mathwar:
Klassik Radio is a radio station in Germany. It specialises in classical 
music, Film music and Lounge music. The channel is receivable in over 
300 German cities via FM, throughout Germany via cable, and in Europe 
via satellite. It is also worldwide streamed on the internet.

### DIFF
--- a/src/connectors/horizonte.cl.js
+++ b/src/connectors/horizonte.cl.js
@@ -9,3 +9,7 @@ Connector.trackSelector = 'h1[class="line__clamp_2"]';
 Connector.playButtonSelector = '.button_control .fa-play';
 
 Connector.trackArtSelector = '.np__thumbnail img';
+
+Connector.currentTimeSelector = '.np__thumbnail .current_time:first-child';
+
+Connector.durationSelector = '.np__thumbnail .current_time:last-child';


src/connectors/horizonte.cl.js
```diff
@@ -9,7 +9,3 @@ Connector.trackSelector = 'h1[class="line__clamp_2"]';
 Connector.playButtonSelector = '.button_control .fa-play';
 
 Connector.trackArtSelector = '.np__thumbnail img';
-
-Connector.currentTimeSelector = '.np__thumbnail .current_time:first-child';
```
inverse, line 13: Looks like this came in from the other PR  - can we remove these changes?

mathwar: Yes, sure! I had trouble to keep two PR separated.

Those changes should be part of [fb7d8dc](https://github.com/web-scrobbler/web-scrobbler/pull/3064/commits/fb7d8dc38697a437a0fff44865d42c35c08cea0a) only for horizonte.cl

mathwar: those changes were reverted, now it's only Klassikradio.

inverse (approved):
Thanks 🙏🏼

PR #3065, opened 2021-11-29
Add short name selector as default for Tidal.

jaccarmac:
In #2789, the track selector was simplified to reflect Tidal's redesign. Now the variance in track names is back (long in the footer, short in the table). Instead of adding the option back, preserve the current behavior as far as possible. That is, use the title in the table if the correct tab is open, falling back to the footer. Annoyingly, the tab state is inconsistent between songs with lyrics and those without. The details tab is either `react-tabs-5` or `react-tabs-7`. That's the reason for the `#nowPlaying .react-tabs__tab-panel--selected` hack.


inverse (approved):
Feels like it's gonna break often with their changes but you know the service better than any 👍🏼

PR #3066, opened 2021-12-02
Select album artist on Tidal if possible.

jaccarmac:
This works if the current page matches the currently playing album.

### DIFF
--- a/src/connectors/tidal.js
+++ b/src/connectors/tidal.js
@@ -22,6 +22,15 @@ Connector.artistSelector = `${Connector.playerSelector} span.artist-link`;
 
 Connector.albumSelector = ['#nowPlaying div.react-tabs a[href^="/album/"]', `${Connector.playerSelector} a[href^="/album/"]`];
 
+Connector.getAlbumArtist = () => {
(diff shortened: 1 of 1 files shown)


inverse (approved):

jaccarmac:
I seem to be observing a bug with this feature wherein the album artist is sometimes not picked up after a track ends and the next begins. It's non-deterministic even then, and I haven't observed it many times.

jaccarmac:
The behavior seems to be a race condition on page load. If `getAlbumArtist` doesn't pick up an album artist during page load, the popup continues to show track information without album artist. However, `getAlbumArtist` continues to run as it is called by the extension. It doesn't affect the popup, the data echoed in the console, or the eventual scrobble.

The object echoed to the console on page load is printed a variable number of times, leading me to think this is a result of the `playerSelector` and extension observer interacting with Tidal's React updates. It might be a problem that the connector looks at data outside of `playerSelector` to determine track information.

jaccarmac:
@alexesprit @inverse I've gone through the debugger a few times, but am unfamiliar enough with the core that I got a bit confused. Eventually, album artist is getting picked up, but it doesn't propagate through to the state, seemingly because it doesn't get seen as a changed value. Any idea what could be going on there?

PR #3089, opened 2021-12-21

mathwar:
seconds), so all songs will be sent at 240 seconds, the default maximum
time. This affects songs that are less than 4 minutes long (many cases).

### DIFF
--- a/src/connectors/klassikradio.de.js
+++ b/src/connectors/klassikradio.de.js
@@ -1,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '#audioPlayer';
+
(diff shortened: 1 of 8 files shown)


src/connectors/klassikradio.de.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
```
inverse, line 1: Should these changes be here?

mathwar: They should not. I have a big problem with mixed commits. I'll do a cleanup and submit again. ;)

mathwar:
Closed for improvement.

PR #3094, opened 2022-01-03

mathwar:
seconds), so all songs will be sent at 240 seconds, the default maximum
time. This affects songs that are less than 4 minutes long (many cases).


inverse:
Thanks 🙏🏼

PR #3101, opened 2022-01-13
Filter out record side names from Bandcamp

### LINKED ISSUE #3093: Bandcamp connector - filter out record side names
Some records on bandcamp have the side and track number in the track name. This is not part of the track name, but is only used to indicate where on the record the track is located. Since they're not part of the track name, I want to scrobble the track without it.

It would be nice if the connector could detect the usage of these indicators and remove them from the track name.

Manually editing the track name is possible, but it requires some extra work.

**Additional context**
Some examples below:
image

image

Examples on Bandcamp: 
- https://zementrec.bandcamp.com/album/002
- https://hectoroaks.bandcamp.com/album/neanderthal-disco

Note that it is possible for bandcamp listings to go further then A and B in case of multiple records in 1 listing (for example C1, C2, D1 etc)

dyerai:
Checks for a record side prefix in the track title on the album page of the Bandcamp connector. If the prefix exists, it's filtered out of the track title string. This allows for more accurate scrobbles of Bandcamp tracks

**Additional context**
Closes #3093

### DIFF
--- a/src/connectors/bandcamp.js
+++ b/src/connectors/bandcamp.js
@@ -65,11 +65,15 @@ function initGenericProperties() {
 
 	Connector.getArtistTrack = () => {
 		const artist = Util.getTextFromSelectors(Connector.artistSelector);
-		const track = parseTrackTitle(Util.getTextFromSelectors(Connector.trackSelector));
+		let track = Util.getTextFromSelectors(Connector.trackSelector);
 
 		if (isArtistVarious(artist, track)) {
 			return Util.splitArtistTrack(track, SEPARATORS);
 		}
+		if (trackContainsRecordSide()) {
+			track = Util.removeRecordSide(track);
+		}
+
 		return { artist, track };
 	};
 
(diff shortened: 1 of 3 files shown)


jaccarmac:
Looks pretty good. I think it would be ideal if there were a check of all tracks on a page, similar to what happens just below your change: https://github.com/web-scrobbler/web-scrobbler/blob/95d657f91b0cd14b30167a1e82012ff337d4b8de/src/connectors/bandcamp.js#L70

src/core/content/util.js
```diff
@@ -110,6 +110,16 @@ const Util = {
 		return { artist, track };
 	},
 
+	/**
+	 * Remove record side from track title.
+	 * @param {String} str String contains track title
+	 * @return {String} String contains track title without record side
+	 */
+	removeRecordSide(str) {
```
inverse, line 118: Could we include a test case for this?

See https://github.com/web-scrobbler/w`eb-scrobbler/blob/3ca9d78037931c8d5517d08fde3b1f5d0330036d/tests/content/util.js#L23 How this is setup for `splitArtistTrack

inverse, line 118: Lets add a test case for this.

See: https://github.com/web-scrobbler/web-scrobbler/blob/3ca9d78037931c8d5517d08fde3b1f5d0330036d/tests/content/util.js#L23 for how we do it for `splitArtistTrack

src/connectors/bandcamp.js
```diff
@@ -258,6 +262,22 @@ function isArtistVarious(artist, track) {
 	return false;
 }
 
+function trackContainsRecordSide() {
+	const RECORD_SIDE_REGEX = /^[A-Z][1-9]\.? /;
```
inverse, line 266: This could come from `Util`  so we have a single regex definiton

src/core/content/util.js
dyerai: Added some test cases, let me know if you can think of anything more.

inverse (approved):
Amazing thanks 👍🏼

Issue #3102, opened 2022-01-14
Connector request: RefNet


jameschensmith:
**Details**

- **Description:** Free 24-hour Christian internet radio. Listen to trustworthy preaching & teaching from well-known Bible teachers, such as Alistair Begg, John MacArthur, Albert Mohler, John Piper, R.C. Sproul, & many more.
- **URL:** https://listen.refnet.fm

**Additional context**

This connector may be a bit challenging to get all the desired information. I did some initial analysis, and this is my observation. Artist & track can easily be queried from the marquee, but the rest of the information is found when you click on the "slot". This fetches from an endpoint `/music/{id}`, where `id` is some form of a dynamic numerical identifier for a grouping of tracks. The data returns is an HTML page, which is added to the DOM as a modal (removed if the modal is closed). So, one option could be to fetch & parse this information. Since it's HTML that's being returned, the parsing of this information could become pretty complex.

An alternative option which I found (would like some comments on whether this option is available to us or not) is that the site pulls all the data on initialization and places it into a `Cache` object. If we have access to this object, we can pull all the needed track information from it and not need to make any network calls. We could even fallback to just the information in the marquee if all else fails.

jameschensmith:
I'm open to working on this. I'll be spending some time on it tomorrow. Definitely would like some input on if it's okay for us to use the `Cache` object, though.

inverse:
Will be in the next release - thanks!

PR #3107, opened 2022-01-23
Add RefNet connector

jameschensmith:
**Summary**

Adds initial connector implementation for RefNet, which can scrobble artist & track from music.

**References**

- #3102


inverse:
Thanks 👍🏼

PR #3109, opened 2022-01-26
Fix Spotify's pause button selector.

### LINKED ISSUE #3105: Scrobbler cannot detect music playing on Spotify web
**Describe the bug**
When I am playing music on Spotify web, the scrobbler does not detect that it is playing music. 
Instead, the scrobbler looks like this: 
image
I suspect the scrobbler isn't picking up the music due to some UI changes on Spotify web.

**How to reproduce**
1. Go to play/open.spotify.com
2. Go to a song/album of your choice.
3. Press the play button.

**Expected behavior**
The scrobbler should detect Spotify web is playing music and should be scrobbling.

**Screenshots**
I threw in some screenshots.

**Debug logs**
I couldn't find debug logs, but I did take a look at the console and saw there was no info: 
image
Playing music on bandcamp, console messages did show up: 
image

-   OS: Mac, Windows 10 (reported by fillycolt)
-   Browser: Chrome Version 97.0.4692.71 (Official Build) (x86_64),  FF 96.0.2 (64-bit)
-   Extension version: 2.49.0

**Additional context**
N/A

jaccarmac:
Fixes #3105.

In #3035, the selector was changed to reflect a change in Spotify's UI which
combined the play and pause buttons. As far as I can tell, the SVG under that
button now varies only with regard to its `path`, so the `fill` selector no
longer works. The
shuffle.one (https://github.com/web-scrobbler/web-scrobbler/blob/db9bf606b52483bb13f16dc684e58f479f3373a3/src/connectors/shuffleone.js#L12)
and YouTube
Music (https://github.com/web-scrobbler/web-scrobbler/blob/db9bf606b52483bb13f16dc684e58f479f3373a3/src/connectors/youtube-music.js#L54)
connectors embed SVG paths, but end up using them in a custom `isPlaying`
implementation. Here, we embed the path to maintain the static
`pauseButtonSelector`.

### DIFF
--- a/src/connectors/spotify.js
+++ b/src/connectors/spotify.js
@@ -27,7 +27,7 @@ Connector.currentTimeSelector = `${playerBar} [data-testid=playback-position]`;
 
 Connector.durationSelector = `${playerBar} [data-testid=playback-duration]`;
 
-Connector.pauseButtonSelector = `${playerBar} [data-testid=control-button-playpause] [fill=none]`;
+Connector.pauseButtonSelector = `${playerBar} [data-testid=control-button-playpause] > svg > path[d="M2.7 1a.7.7 0 00-.7.7v12.6a.7.7 0 00.7.7h2.6a.7.7 0 00.7-.7V1.7a.7.7 0 00-.7-.7H2.7zm8 0a.7.7 0 00-.7.7v12.6a.7.7 0 00.7.7h2.6a.7.7 0 00.7-.7V1.7a.7.7 0 00-.7-.7h-2.6z"]`;
 
 Connector.applyFilter(MetadataFilter.getSpotifyFilter());
 
--- a/src/connectors/spotify.js
+++ b/src/connectors/spotify.js
@@ -3,6 +3,7 @@
 const playerBar = '.Root__now-playing-bar';
 
 const artistSelector = `${playerBar} [data-testid="context-item-info-artist"]`;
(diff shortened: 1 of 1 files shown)


jaccarmac:
I'd like to refactor this to at least break the selector into two parts. Is it worth it to create a custom `isPlaying` or is an interpolated selector good enough?

We could also match on the first few characters of `path[d]`: The play button starts with `M3`.

I haven't tested this widely so it might be more brittle than I think it is.

ahnesther:
What would the selector look like if broken into two parts?

jaccarmac:
```javascript
const playingPath = 'M2.7 1a.7.7 0 00-.7.7v12.6a.7.7 0 00.7.7h2.6a.7.7 0 00.7-.7V1.7a.7.7 0 00-.7-.7H2.7zm8 0a.7.7 0 00-.7.7v12.6a.7.7 0 00.7.7h2.6a.7.7 0 00.7-.7V1.7a.7.7 0 00-.7-.7h-2.6z'
Connector.pauseButtonSelector = `${playerBar} [data-testid=control-button-playpause] > svg > path[d="${playingPath}"]`;
```

or thereabouts.

ahnesther:
Hm, I see, thanks! That seems easier to manage than just having one long path in the Connector line. LGTM but since I'm a newb to the codebase who might be missing some details, I think another reviewer's eyes would be great here.

inverse:
Not sure of a better solution without using the aria labels which would require a bigger mapping of all languages. Let me know when this is ready @jaccarmac  if you wanna change the selector use a separate variable or not. Might make it easier for refactoring.

inverse (approved):

jaccarmac:
I think so. It should be ready to push out now; See if it smokes.

PR #3124, opened 2022-02-07
Add support for La Radio du bord de l'eau

fetzu:
Adds support for auborddeleau.radio ~~(main/home page only)~~ for both the player on the main page and the one on the "on air" page. Both players support cover/artwork.

**Additional context**
1. Had to use double quote symbol (") instead of single quote (') to get the label right; I hope that's ok.
2. ~~The "OnAir" page player seems to be using an iframe, and I have not gotten it to work yet.~~


fetzu:
I think I'm done ;)...

inverse (approved):
Nice one!

PR #3141, opened 2022-02-20
Added 256 & 512 squared extension icons/images

mountainash:
Created new high-res extension icon PNGs and updated manifest for cleaner "branding" on higher res monitors.

Compressed (meta data stripping) all icons (using [ImageOptim](https://github.com/ImageOptim/ImageOptim)) - there should be no image quality loss.
Screen Shot 2022-02-20 at 10 55 55 pm


src/icons/cover_art_unknown.svg
```diff
@@ -1,55 +1 @@
...
-         rdf:about=""><dc:format>image/svg+xml</dc:format><dc:type
-           rdf:resource="http://purl.org/dc/dcmitype/StillImage" /></cc:Work></rdf:RDF></metadata><defs
-     id="defs4145" /><sodipodi:namedview
-     pagecolor="#ffffff"
-     bordercolor="#666666"
-     borderopacity="1"
-     objecttolerance="10"
-     gridtolerance="10"
-     guidetolerance="10"
-     inkscape:pageopacity="0"
```
inverse, line 31: What was the reason for optimising these svgs? Should we maintain this meta data?

src/manifest.json
```diff
@@ -6,7 +6,9 @@
   "icons": {
     "16": "icons/icon_generic_16.png",
     "48": "icons/icon_chrome_48.png",
-    "128": "icons/icon_chrome_128.png"
+    "128": "icons/icon_chrome_128.png",
+    "256": "icons/icon_chrome_256.png",
```
inverse, line 10: Could you point me to the documentation where it points for sizes larger then 128?

inverse (commented):
Two comments 👍🏼

src/manifest.json
mountainash: https://developer.apple.com/documentation/safariservices/safari_web_extensions/creating_a_safari_web_extension It's in preparation.

src/icons/cover_art_unknown.svg
mountainash: Like _nearly_ everything "less is more"... less to download each time is better for everyone and the environment.

Lots of this meta data is just marketing fluff - the images are all so basic that any vector/svg reading app can re-open them and edit them (... should you not like the style of this question mark).

But these icons are "never" re-edited... they are sometimes "replaced". There's no need to add ~75% extra on most of the largest assets in this project for very little reason.

Issue #3142, opened 2022-02-21
Connector for fluxfm.de doesn't work anymore


fckawe:
The connector for the web radio on fluxfm.de ("Flux FM Berlin", already listed as supported website) does not work anymore. They have recently redesigned their website including the web player.
Any chance of getting this back to work?
Thanks in advance!

ichmoimeyo:
https://www.fluxfm.de/channels/f71292b4-14b1-4353-96b5-2114a0793334

Jazzradio Schwarzenstein not working.

PR #3145, opened 2022-02-22
Fixed the connector for FluxFM web radio.

fckawe:
The connector didn't work anymore after the website was redesigned.
Album info and track art aren't available anymore. Artist and title cannot be parse from separate elements.

**Additional context**
Fixes issue #3142.

### DIFF
--- a/src/connectors/fluxfm.js
+++ b/src/connectors/fluxfm.js
@@ -1,14 +1,19 @@
 'use strict';
 
-Connector.playerSelector = '.hero-player__headline';
+Connector.playerSelector = '.playbar';
 
 Connector.getArtistTrack = () => {
-	const artistTrack = Util.getTextFromSelectors('.hero-player__title');
+	const artistTrack = Util.getTextFromSelectors('.playbar__info__artist');
 	return Util.splitArtistTrack(artistTrack, null, { swap: true });
 };
 
 Connector.isPlaying = () => {
 	return Util.getAttrFromSelectors('rs-play-button > rs-button > button', 'title') === 'Pausieren';
+};
+
+Connector.isStateChangeAllowed = () => {
+	const artistTrack = Util.getTextFromSelectors('.hero-player__title');
+	return artistTrack.trim() !== 'Livestream - FluxFM';
 };
 
 Connector.onReady = Connector.onStateChanged;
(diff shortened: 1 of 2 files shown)


src/connectors/fluxfm.js
```diff
@@ -1,17 +1,14 @@
...
+Connector.playerSelector = '.hero-player__headline';
 
-Connector.artistSelector = '#track-artist';
-
-Connector.trackSelector = '#track-title';
-
-Connector.albumSelector = '#track-album';
-
-Connector.trackArtSelector = '#covercontainer img';
+Connector.getArtistTrack = () => {
```
inverse, line 5: Could be simplified to

```
Connector.artistTrackSelector = '.hero-player__title';
```

inverse (commented):
Looks good aside the one comment 👍🏼

fckawe:
Okay, I didn't notice that one in the documentation... :/ Thanks for the review!
I'm not sure how to change that now, since I'm not so familiar with the whole workflow. Do I have to do something at all?

fckawe:
Wait! I've made a mistake, sorry.
In my first version I took the artist - title element and separated it manually (via split). Then I've found the function to do the split artist and title.
The problem: The page is showing the title first and then the artist.
So, with that version artist and title are swapped! I didn't notice that because the music they played was quite unknown... :(
I guess I have to change it back to manually splitting...

fckawe:
Commited two more changes:

1. Using different selectors, now. That way the connector not only works with the live radio player but also with the "channels" they provide on the website.
2. Disabling scrobbling if track title is "Livestream" and artist is "FluxFM", which sometimes happens when no music is played and no upcoming track is given.

I think it works quite well, now.

src/connectors/fluxfm.js
```diff
@@ -1,17 +1,19 @@
...
+	const artistTrack = Util.getTextFromSelectors('.playbar__info__artist');
+	return Util.splitArtistTrack(artistTrack, null, { swap: true });
+};
 
 Connector.isPlaying = () => {
-	return $('#controls-play').hasClass('play');
+	return Util.getAttrFromSelectors('rs-play-button > rs-button > button', 'title') === 'Pausieren';
```
inverse, line 11:
```suggestion
	return Util.getAttrFromSelectors('.playbar rs-play-button > rs-button > button', 'title') === 'Pausieren';
```

inverse (changes requested):
Minor change but looks a lot better 👍🏼

inverse (approved):
Modified with changes after testing locally 👍🏼

Issue #2980, opened 2021-09-11
Jango.com's new layout cannot scrobble


effluviummist:
Hello, ever since jango.com has new layout, the web scrobbler cannot scrobble anymore

Opera Windows 10

effluviummist:
![jango.com issue](https://user-images.githubusercontent.com/90512000/132948487-9a7bf627-77a1-4f80-bf46-127e3791f71b.png)

MXB2001:
Same here. The old problem of not working until you reloaded doesn't apply anymore. Nothing makes it scrobble now. (Well I haven't tried magic incantations... ; - )

MXB2001:
2022 and jango still doesn't scrobble.

Issue #2951, opened 2021-08-07
jango.com not scrobbling


ormulum:
the extension icon is grey and reads "web scrobbler does not support this website"
firefox on ubuntu linux

ormulum:
it scrobbles normally  if i'm not signed in, but as soon as i'm signed in the icon greys and stops scrobbling

PR #3146, opened 2022-02-24
Fixed jango.com connector after layout changes

fckawe:
Fixed the connector for jango.com after layout changes on their website. Tested while being logged in to Jango (free account) as well as without being logged in.

**Additional context**
Fix was requested with issue #2980.


fckawe:
... there's another issue that probably described the same problem: #2951 (marked as outdated).

inverse (approved):
Thanks so much!

Issue #3060, opened 2021-11-22
open.fm - year in the album name


clr134:
Album name contains year in square brackets which causes incorrect scrobbles.

**How to reproduce**
1. Go to https://open.fm/stacja/impreza
2. Check album name.

**Expected behavior**
Year should be stripped out.

**Environment (please complete the following information)**
-   OS: Windows
-   Browser: Firefox
-   Extension version: 2.48.0

inverse:
Will be fixed in the next release! Thanks @fckawe

PR #3156, opened 2022-03-04
open.fm-connector: remove year from album name (issue #3060)

fckawe:
open.fm provides the release year within every album name. This change adds an album filter to the corresponding connector, which removes the trailing year.

**Additional context**
The bug was reported with issue #3060.


inverse:
Thanks!

Issue #2835, opened 2021-04-23
https://www.oxigenio.fm/


pedrojvaz:
**Provide information about website**
Radio station based in Lisbon, PT

There is a player on the website that scroll information about the current program and the music being played, in blue

![image](https://user-images.githubusercontent.com/4522620/115844563-3df04f80-a420-11eb-9bd2-1f69fedf510b.png)

**Additional context**
No account needed

fckawe:
I tried to build a connector for oxigenio.fm, here's the content so far:

```
'use strict';

Connector.playerSelector = '.wp-lunaradio.lunaaudioplayer';

Connector.artistTrackSelector = 'div[id^="wp"][id$="texttitle"] span:nth-child(1)';

Connector.playButtonSelector = 'div[id^="wp"][id$="buttonplay"]';

Connector.onReady = Connector.onStateChanged;
```

It seems to work pretty well. But quite often the player shows information that doesn't seem to be a "artist - track" combination, but rather the name of a show or something like that. A few minutes before it showed "hip-hop - instrumental" for example, or this:

![image](https://user-images.githubusercontent.com/1059389/157906410-2179d627-89ca-4ca9-ac9e-67868f374056.png)

So, the connector would scrobble these "titles", which imho is pretty nonsense. I don't have a clue how to prevent that, since it doesn't seem to follow any rules.

Ideas anyone?

inverse:
@fckawe thanks for looking! 


You could try adding something like

```js
Connector.isStateChangeAllowed = () => {
	return Util.getTextFromSelectors(Connector.artistTrackSelector).includes('-');
};
```

Which would restrict the scrobbling to when it matches `-` in the matcher.

fckawe:
@inverse Hmm, that wouldn't scrobble regular tracks either, would it?

The connector has to "understand" that "Oxigenio - Insistencia!" isn't a track to scrobble but "Nirvana - Lithium" is.

Okay, it could check if the artist equals "Oxigenio" and if so doesn't scrobble (assuming that "Insistencia!" is a show on that radio and so this could follow a general pattern like "Oxigenio - Show name").

But what about "Hip-hop - Instrumental", then? Doesn't seem to contain an artist at all and looks like some kind of "generic track name".

inverse:
@fckawe Ahh perhaps I misinterpreted the needs and lacking PT language skills. 

Yeah will be a tough one to implement is track names are not actual unless you can map a list of things to ignore.

[...]

PR #3169, opened 2022-03-13
Added a connector for portuguese radio Oxigenio FM

fckawe:
Added a connector for portuguese radio Oxigenio FM.

**Additional context**
The connector was requested and discussed with issue #2835.

### DIFF
--- a/src/connectors/oxigenio.fm.js
+++ b/src/connectors/oxigenio.fm.js
@@ -10,5 +10,3 @@ Connector.isStateChangeAllowed = () => {
 	const artistTrack = Util.getTextFromSelectors(Connector.artistTrackSelector);
 	return artistTrack.includes('-') && !artistTrack.match(/^\s*OXIGENIO - /);
 };
-
-Connector.onReady = Connector.onStateChanged;
--- a/src/connectors/oxigenio.fm.js
+++ b/src/connectors/oxigenio.fm.js
@@ -1,0 +1,14 @@
+'use strict';
+
+Connector.playerSelector = '.wp-lunaradio.lunaaudioplayer';
+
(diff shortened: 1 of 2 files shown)


src/connectors/oxigenio.fm.js
```diff
@@ -0,0 +1,14 @@
...
+Connector.artistTrackSelector = 'div[id^="wp"][id$="texttitle"] span:nth-child(1)';
+
+Connector.playButtonSelector = 'div[id^="wp"][id$="buttonplay"]';
+
+Connector.isStateChangeAllowed = () => {
+	const artistTrack = Util.getTextFromSelectors(Connector.artistTrackSelector);
+	return artistTrack.includes('-') && !artistTrack.match(/^\s*OXIGENIO - /);
+};
+
+Connector.onReady = Connector.onStateChanged;
```
inverse, line 14: I think we can drop this one

fckawe: Sorry, could you explain in detail what you mean?

inverse: My understanding of this onReady behaviour is when you need to override the default behaviour when the scripts  are injected.

See: https://github.com/inverse/web-scrobbler/blob/882fdba3702795d0a337a27323523cd81c147263/src/core/content/connector.js#L418

by default the state transitions are handled via the `playerSelector` which is watched for mutations.

fckawe: Okay, thanks for this explanation! I wasn't sure what's that line for, tbh.
I'll remove that, check if it works without it and if so, commit.

inverse: Yeah I think some other connectors also used this pattern when in reality they don't need to. Aside that the additions looks good. Thanks for your support 🙏🏼

inverse (approved):
Thanks!

PR #3175, opened 2022-03-17
Bandcamp fix albumArtist filter

### LINKED ISSUE #3165: some bandcamp albums don't scrobble
**Describe the bug**
Some bandcamp albums get the plugin indication that they're recognized and going to be scrobbled, but the scrobbles don't show up on last.fm.

**How to reproduce**
1. Go to https://houseofmythology.bandcamp.com/album/slava-ukraini (or another release with the same issue, but this is one of them) on Firefox 97.0.2 (I did not try in another browser or version)
2. Click on play and see the plugin indicator correctly identifying the tracks, just like it shows when the tracks get scrobble (including firing up the notification when a new track starts)
3. Watch the configured last.fm account and realize that the tracks don't show up there

**Expected behavior**
The tracks should be scrobbled.

**Screenshots**
I can make any screenshots you want/need if this description is not enough and you are not able to reproduce the problem.

**Debug logs**
I did not grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension), but if you can't reproduce it or in any case think that they'd be useful let me know, and I'll gladly get them.

-   OS: Ubuntu GNU/Linux
-   Browser: Firefox
-   Extension version: 2.55.0

inverse:
Resolves #3165

Last.fm API returns an error if the album artist contains "Various Artists" with more than just "Various Artists". 

This change introduces a scrobbler level filtering mechanism to hook into to apply filtering on songs.

@[fckawe](https://github.com/fckawe) can you test?

### DIFF
--- a/src/connectors/bandcamp.js
+++ b/src/connectors/bandcamp.js
@@ -17,6 +17,9 @@ let bandcampFilter = MetadataFilter.createFilter(
 		MetadataFilter.removeZeroWidth
 	)
 );
+bandcampFilter = bandcampFilter.extend(MetadataFilter.createFilter({
+	albumArtist: [cleanVariousArtists],
+}));
 
 setupConnector();
 
@@ -312,4 +315,8 @@ function getData(selector, attr) {
 	}
 
 	return {};
+}
+
+function cleanVariousArtists(text) {
+	return text.includes('Various Artists') ? 'Various Artists' : text;
 }
--- a/src/connectors/bandcamp.js
+++ b/src/connectors/bandcamp.js
@@ -17,9 +17,6 @@ let bandcampFilter = MetadataFilter.createFilter(
 		MetadataFilter.removeZeroWidth
 	)
 );
-bandcampFilter = bandcampFilter.extend(MetadataFilter.createFilter({
-	albumArtist: [cleanVariousArtists],
-}));
 
 setupConnector();
 
@@ -315,8 +312,4 @@ function getData(selector, attr) {
 	}
 
 	return {};
-}
-
-function cleanVariousArtists(text) {
-	return text.includes('Various Artists') ? 'Various Artists' : text;
 }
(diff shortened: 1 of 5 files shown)


fckawe:
Yes, I can confirm that this change solves the issue with the sampler album mentioned by the issue creator. Tested that and some other albums.

But: I think the issue at least applies to the bandcamp embedded connector as well. Shouldn't that also be fixed?
Basically the issue could apply to all connectors that give an album artist. And on the other hand the change is only necessary for LastFM (as far as we know).

inverse:
Valid point - will need to look to see how we can plug global filters per scrobbler

inverse:
@fckawe what about this approch?

fckawe:
> what about this approch?

@inverse Yep, looks good to me, as it covers both -- filter for all connectors but filter for last.fm only. I could run some tests later...

inverse:
Exactly - I allowed any scrobbler to plug in such filtering on the scrobbler level but only applied this rule on the last.fm one.

yayuyokitano:
Sorry for the late review, been busy with irl stuff

The approach looks reasonable, though I think adding to the comment in the base-scrobbler.js declaration to clarify that _applyFilter works by being overridden in the individual scrobblers is a good idea.

Other than that, as long as it passes tests it seems mergeable to me

fckawe:
> I could run some tests later...
Tested the changes with some albums on bandcamp (being connected to last.fm only, since I haven't an account on any other platform). Works!

inverse:
> Other than that, as long as it passes tests it seems mergeable to me

@yayuyokitano our test coverage isn't really that great :/ For some reason it's reporting 100% but in reality it's not... something to look into at some point.

I just pushed a change which refactors this logic a bit. By making all implementations call this filter making things a bit cleaner. 

What do you think?

PR #3185, opened 2022-03-22
Fix bug causing connectors to not be recognized

### LINKED ISSUE #3184: Randomly does not scrobble at all
This bug has been reported by a few people, and I have now reproduced it.

The bug seems to be related to other extensions. I reproduced with the following extensions:

1. Chrome Remote Desktop
2. Netflix Party is now Teleparty
3. BetterTTV

The issue started appearing at 2, 3 appears to have increased the frequency a bit.

When starting the browser, there will be around a 15-30% chance that web scrobbler just doesn't work.
On any and all websites it will display website not supported.
The only fix is seemingly to restart the browser.

Reproduced in chrome 99.0.4844.82 64bit win11. Has been reproduced in win10 as well (same ver).

Has also been reproduced all the way back to web scrobbler 2.49.0, and is also reproduced in current 2.57.0. It is not related to a recent change.

Weirdly, when multiple versions of web scrobbler are installed and activated, they will all work or not work at the same time, some external factor consistently causes this.

Still looking into what the problem could be, logs don't seem to give any hint.

yayuyokitano:
I don't know either. But it seems to work.
Fixes #3184

### DIFF
--- a/src/core/background/extension.js
+++ b/src/core/background/extension.js
@@ -90,12 +90,20 @@ define((require) => {
 				// Don't let the extension fail on Firefox for Android.
 			}
 
-			browser.tabs.onUpdated.addListener((_, changeInfo, tab) => {
+			browser.tabs.onUpdated.addListener(async (_, changeInfo, tab) => {
+
 				if (changeInfo.status !== 'complete') {
 					return;
 				}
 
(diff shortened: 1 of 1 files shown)


inverse:
Nice one 🎉

Issue #3095, opened 2022-01-04
Qobuz web player adds artist to the album tag


Alsweider:
Playing songs on [Qobuz web player](https://play.qobuz.com/), the artist name is always added to the title of the album in the scrobbler, so I have to fix it manually for every song to avoid scrobbling incorrect album names like "Metallica - Metallica".

Browser: Brave 1.33.106 Chromium: 96.0.4664.110
Web Scrobbler version: 2.48.0

![2022-01-04 06_34_03-Window](https://user-images.githubusercontent.com/30653982/148014614-f7a45586-580a-4985-b4a1-6f7239e9fab7.jpg)

justathoughtor2:
Same issue here. I infinitely prefer web scrobbler over Qobuz's built-in scrobbling capability, but this is a bit of a problem. I'll see if I can figure out a pull request. Should be simple enough to resolve.

justathoughtor2:
Pull request #3194 submitted. This should resolve it, unless I overlooked something. It definitely works on my end, and all the CI tests pass. Hopefully this will get resolved for all users soon!

inverse:
Will be in the next release! Thanks @justathoughtor2 for the fix

justathoughtor2:
> Will be in the next release! Thanks @justathoughtor2 for the fix

Of course! Thanks so much for merging so quickly!

PR #3194, opened 2022-03-27
Qobuz - select only album as album

justathoughtor2:
Qobuz seems to have made a change that results in a DOM that includes multiple textual components inside of the `.player__track-album` element. The key components are all included as `a` tags, and the second `a` tag is always the album (with the first being the artist, and the optional third being the playlist if present).

This commit makes a change to always select the second `a` tag within `.player__track-album` to ensure only the album is selected (till further changes are made on the Qobuz end).

**Additional context**
Resolves web-scrobbler/web-scrobbler#3095

### DIFF
--- a/src/connectors/qobuz.js
+++ b/src/connectors/qobuz.js
@@ -7,7 +7,7 @@ Connector.playerSelector = '.player';
 
 Connector.trackSelector = '.player__track-title';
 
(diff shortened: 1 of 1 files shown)


inverse (approved):
🙏🏼 thanks for helping with this one!

PR #3199, opened 2022-03-29
Update Apple Music domain matcher

leonjza:
Apple music used to live only on https://beta.music.apple.com. It can now also be accessed on https://music.apple.com. With the old matcher, `*://*.music.apple.com/*`, using Apple music at https://music.apple.com would fail to match because of the missing `.`.

This PR removes that `.` such that both `beta.music.apple.com` and `music.apple.com` matches.

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -1241,7 +1241,7 @@ const connectors = [{
 }, {
 	label: 'Apple Music',
 	matches: [
-		'*://music.apple.com/*',
+		'*://*music.apple.com/*',
 	],
 	js: 'connectors/apple-music.js',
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks

hans-juergen:
Maybe also offering the beta version as a supported site still makes sense, because it has been rewritten and improved by Apple according to some Reddit postings, so people might also want to scrobble with it:

https://www.reddit.com/r/AppleMusic/comments/v816up/am_beta_web_page_has_had_its_code_completely/
https://www.reddit.com/r/AppleMusic/comments/v7qf0a/apple_music_web_beta_suggesting_music_app_for/

Issue #2290, opened 2020-03-02
Option to only scrobble edited tracks (track whitelist).


RudeySH:
I'm using Web Scrobbler exclusively for scrobbling music on YouTube. I have a couple of playlists and I'm currently editing the tracks so that the metadata is correct. I have not enabled "Force track recognition" because I don't want messy scrobbles. I'd like to take it one step further and only scrobble tracks that I have edited. Only when the metadata of a track was edited by me, should it be scrobbled. When YouTube or any service plays a track that I have already edited before, it should scrobble automatically of course.

A checkbox option to enable the "Scrobble edited tracks only" behavior would help greatly. One thing to consider after implementing such checkbox, is how to deal with tracks that already have proper metadata and don't need to be edited (as far as the user is concerned). For those tracks, I'd imagine I would have to press the "edit" button, just to press "submit" without editing anything, just to get them into the edited tracks list,

Perhaps it would be even better to implement a whitelist option (and a blacklist if that sounds useful to you).  Adding tracks to the whitelist would lead to only those tracks being scrobbled, which is what I need. You could make it so that editing tracks would automatically add them to the whitelist.

alexesprit:
> One thing to consider after implementing such checkbox, is how to deal with tracks that already have proper metadata and don't need to be edited (as far as the user is concerned). For those tracks, I'd imagine I would have to press the "edit" button, just to press "submit" without editing anything, just to get them into the edited tracks list,

This feature could be implemented by marking all tracks as unknown, except that ones are edited by a user. So, you're almost right: you would just click on "Submit" button; you don't need to click on the "Edit" button, because the info popup displays edit controls for unknown tracks by default. Of course, this behavior is enabled, when the respective option is turned on.

> Perhaps it would be even better to implement a whitelist option (and a blacklist if that sounds useful to you). Adding tracks to the whitelist would lead to only those tracks being scrobbled, which is what I need. You could make it so that editing tracks would automatically add them to the whitelist.

I'm not sure if it's worth adding a whitelist, since it can be resolved by using current features.

yayuyokitano:
Added by #3222

PR #3211, opened 2022-04-05
Add option to scrobble edited tracks only

RudeySH:
I added an option to scrobble edited tracks only.

image

I think this will be very useful to users who want to scrobble music on YouTube without accidentally scrobbling videos that are recognized on Last.fm while not being music at all. I also think this is perfect for users who want full control over the metadata that is being scrobbled with.

When this option is enabled, an unedited song is treated as if it is unknown. This approach was suggested by @alexesprit in #2290.

**I need feedback** on the following points:

1. Currently, the new option overrules Force Track Recognition. With both Force Track Recognition and the new option enabled,  new tracks are never "recognized" until they have been edited. It doesn't really make sense to enable both options. Should the checkboxes be programmed so the user can't select both? Should we refactor these options into radio buttons instead (and migrate)?
2. I named the new option "Scrobble edited tracks only" but I don't know if this is the best name. Other candidates I could think of are "Only scrobble edited tracks", or "Recognize edited tracks only" (this one is technically more accurate to what this new option actually does).
3. Because unedited songs are treated as unknown, they will trigger the unrecognized song notification (if this notification enabled). I don't mind this, I think it makes sense the user is notified when they need to edit something. We could opt to add a separate unedited song notification to make the message more specific. However, doing this would result in more work: a new notification message plus translations, another new checkbox option to enable/disable unedited song notifications, and an additional ControllerMode (e.g. ControllerMode.Unedited so that we can differentiate from ControllerMode.Unknown).

I'm looking forward to hear feedback on this.

Currently, this PR contains 2 commits from another PR of mine, #3210. I intend to get that one merged first, then rebase this one.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* eol=lf
+* text=auto eol=lf
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "mocha-lcov-reporter": "1.3.0",
         "node-define": "0.1.1",
         "nyc": "15.1.0",
-        "prettier": "2.6.1",
+        "prettier": "2.6.2",
         "prettier-config-web-scrobbler": "0.1.0",
         "remark-preset-lint-recommended": "6.1.2",
         "stylelint": "14.6.1",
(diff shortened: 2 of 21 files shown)


RudeySH:
> Currently, this PR contains 2 commits from another PR of mine, #3210. I intend to get that one merged first, then rebase this one.

#3210 got merged, but rebasing was a mistake. I think the commit history of this PR is beyond repair. I made a new PR: #3222.

PR #3219, opened 2022-04-11
Add connector for burntable.com

hornbuckle:
New connector for burntable.com, a site where users can upload digital files of their vinyl records. Tested with my personal account and scrobbles are successful. Screenshot below.
burntablescrobble

NOTES:

- Connector is only functional on mixes via https://burntable.com/listen and user playlists.
- Album playback is limited to only 60-second increments due to licensing restrictions and uses a different player, so those plays are not captured by the connector.
- Additional logic required due to DOM structure and metadata elements not found inside the player wrapper prior to initial playback.
- Duration only available on the full-page player and not in the persistent player, which is being used for the connector and is initially hidden but appears when navigating the site during playback.

### DIFF
--- a/src/connectors/burntable.js
+++ b/src/connectors/burntable.js
@@ -1,0 +1,41 @@
+'use strict';
+
+// only works on mixes via https://burntable.com/listen or user playlists; album playback limited to 60-second increments
+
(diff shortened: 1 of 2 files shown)


inverse:
🙏🏼  Thanks

PR #3222, opened 2022-04-13
Add option to scrobble edited tracks only

RudeySH:
I added an option to scrobble edited tracks only.

image

I think this will be very useful to users who want to scrobble music on YouTube without accidentally scrobbling videos that are recognized on Last.fm while not being music at all. I also think this is perfect for users who want full control over the metadata that is being scrobbled with.

When this option is enabled, an unedited song is treated as if it is unknown. This approach was suggested by @alexesprit in #2290.

**I need feedback** on the following points:

1. Currently, the new option overrules Force Track Recognition. With both Force Track Recognition and the new option enabled,  new tracks are never "recognized" until they have been edited. It doesn't really make sense to enable both options. Should the checkboxes be programmed so the user can't select both? Should we refactor these options into radio buttons instead (and migrate)?
2. I named the new option "Scrobble edited tracks only" but I don't know if this is the best name. Other candidates I could think of are "Only scrobble edited tracks", or "Recognize edited tracks only" (this one is technically more accurate to what this new option actually does).
3. Because unedited songs are treated as unknown, they will trigger the unrecognized song notification (if this notification enabled). I don't mind this, I think it makes sense the user is notified when they need to edit something. We could opt to add a separate unedited song notification to make the message more specific. However, doing this would result in more work: a new notification message plus translations, another new checkbox option to enable/disable unedited song notifications, and an additional ControllerMode (e.g. ControllerMode.Unedited so that we can differentiate from ControllerMode.Unknown).

I'm looking forward to hear feedback on this.


inverse:
Thanks for providing this feature and detailed introduction into what it brings.

> Currently, the new option overrules Force Track Recognition. With both Force Track Recognition and the new option enabled, new tracks are never "recognized" until they have been edited. It doesn't really make sense to enable both options. Should the checkboxes be programmed so the user can't select both? Should we refactor these options into radio buttons instead (and migrate)?

Perhaps a radio group would be a good way to represent this as you get the benefit of the existing tool tips. Alternatively a drop down for this forced/controlled behaviour.

>  named the new option "Scrobble edited tracks only" but I don't know if this is the best name. Other candidates I could think of are "Only scrobble edited tracks", or "Recognize edited tracks only" (this one is technically more accurate to what this new option actually does).

I think what you have is descriptive enough :)

> Because unedited songs are treated as unknown, they will trigger the unrecognized song notification (if this notification enabled). I don't mind this, I think it makes sense the user is notified when they need to edit something. We could opt to add a separate unedited song notification to make the message more specific. However, doing this would result in more work: a new notification message plus translations, another new checkbox option to enable/disable unedited song notifications, and an additional ControllerMode (e.g. ControllerMode.Unedited so that we can differentiate from ControllerMode.Unknown).

We can iterate if we get confusion :)

RudeySH:
> Perhaps a radio group would be a good way to represent this as you get the benefit of the existing tool tips.

If were going with radio buttons, I suggest the following layout (WIP):

image

Some notes:
- The middle option is selected by default, because it's the default behavior of web scrobbler.
- With the intent to put related options closely together, the **Edited tracks** section is moved so that it comes directly after the new "Scrobble edited tracks only" option. The **YouTube** section is a bit further down now.
- The "Scrobble podcasts" option was moved because I did not want it to be between "Scrobble edited tracks only" and the **Edited tracks** section. I also did not want to put "Scrobble podcasts" between "Notify if track is not recognized" and the radio group, because those are also closely related options.

inverse:
Looks good but I feel like we should leave the options ordering alone for now.

What about moving those 3 radio buttons into their own section titled "Scrobble behaviour" or something? This could bit above or below the YouTube section.? What do you think?

RudeySH:
I've begun working on this PR again. Can this be reopened?

I've added the "Scrobble behavior" section as suggested. I left the ordering of the other sections/options as-is.

image

Is this better?

inverse:
Sounds better to me - Perhaps highlighting which one is the default one for newcomers.

RudeySH:
Just now, I have completed the radio button functionality. When you select a checkbox, the corresponding option is set to true, and the options of the other two radio buttons will be set to false.

I also resolved the merge conflicts that this PR had.

I tested all three radio button options, and they worked.

> Sounds better to me - Perhaps highlighting which one is the default one for newcomers.

I added "(default)" to the label of the radio button that is selected by default (the middle one). Is that what you wanted?

I think this PR is ready to be reviewed and merged!

inverse (approved):

inverse:
Sorry for the delay in replying - Thanks for the contribution!

src/core/background/pipeline/metadata.js
```diff
@@ -46,7 +46,10 @@ define((require) => {
 		}
 
 		const forceRecognize = await Options.getOption(Options.FORCE_RECOGNIZE, connector.id);
-		song.flags.isValid = isSongValid || forceRecognize;
+		const scrobbleEditedTracksOnly = await Options.getOption(Options.SCROBBLE_EDITED_TRACKS_ONLY, connector.id);
+
+		song.flags.isValid = (isSongValid || forceRecognize) &&
```
inverse, line 51: This condition is a bit difficult to understand and perhaps we could make it easier - but not a blocker.

PR #3230, opened 2022-04-22
(fix) update Idagio `isPlaying` selector

brodie-g:
Idagio connector was broken because the play/pause button changed position.
Looks like a 'shuffle' element was added to the playback controls. The play/pause button is now the third (3rd)
child.

image


inverse:
Thanks!

PR #3234, opened 2022-04-26
Add connector for Indonesian music archive Irama Nusantara

Slickerius:
New connector for the Indonesian music archive and streaming service iramanusantara.org. 
kisahpasarbaru

**Notes**
* The site does not currently keep track of which albums the tracks are from (playback is based on the album that the user is currently viewing), so album titles are fetched only if the track is being played directly from the album page.

### DIFF
--- a/src/connectors/iramanusantara.js
+++ b/src/connectors/iramanusantara.js
@@ -1,0 +1,14 @@
+'use strict';
+
+Connector.playerSelector = '#player';
+
+Connector.artistSelector = '#pl-info-text > h4';
+
+Connector.trackSelector = '#pl-info-text > h3';
+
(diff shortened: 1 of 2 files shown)


src/connectors/iramanusantara.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+Connector.playerSelector = '#player';
+
+Connector.artistSelector = '#pl-info-text > h4';
+
+Connector.trackSelector = '#pl-info-text > h3';
+
+// Album title is only available when track is played through the album page
```
inverse, line 9:
```suggestion
// Album title is only available when track is played through the album page
```

PR #3237, opened 2022-04-28
Minor CSS edit to fix small gap below image in popup

hornbuckle:
Added overflow: hidden to .main-container
Alternate solution is to add display: block to .album-art


inverse:
Thanks!

PR #3239, opened 2022-05-01
Added connector for Husk Recordings website audio player

hujiba:


src/connectors/huskrecordings.js
```diff
@@ -0,0 +1,14 @@
...
+Connector.playerSelector = '#webamp';
+
+Connector.trackSelector = '#title';
+
+Connector.artistSelector = '#artist';
+
+Connector.albumSelector = '#album';
+
+Connector.isPlaying = () => Util.hasElementClass(
+	['#metadata'],
```
inverse, line 12:
```suggestion
	'#metadata',
```

Could be simplified since we're only passing in a single selector

inverse (changes requested):
Minor change and we can merge :+1:

inverse (approved):

Issue #3249, opened 2022-05-09
Incorrect Bandcamp originUrl when scrobbling on the 'wishlist' page


MonkeyDo:
**Describe the bug**
Following #3236, when scrobbling from the bandcamp wishlist page, the first album on the page is selected for the `originUrl` link instead of the currently playing album.

**How to reproduce**
1. Go to your Bandcamp wishlist page (https://bandcamp.com/MY_USER_NAME/wishlist)
2. Click on any play icon on the page to play a track (except the first one in the list :) )
3. The scrobble's originUrl is set as the URL for the first album in the list

**Expected behavior**
The originUrl should be set to the currently playing album's URL rather than the first album in the list

**Screenshots**
<img width="1254" alt="image" src="https://user-images.githubusercontent.com/6179856/167443934-56ef4479-f5e0-4fcf-b347-f29f1070bfac.png">

**Debug logs**
See the `artist` (GOD IS AN ASTRONAUT) is not the one in the `originUrl` (merlin666; see screenshot: first album in the list is by 'Merlin')
```
Web Scrobbler: {
  "track": "Luminous Waves",
  "artist": "GOD IS AN ASTRONAUT",
  "album": "Ghost Tapes #10",
  "albumArtist": null,
  "uniqueID": null,
  "duration": 228,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://f4.bcbits.com/img/a1445989498_16.jpg",
  "isPodcast": false,
  "originUrl": "https://merlin666.bandcamp.com/album/the-wizard"
}
```

**Additional context**
The [querySelector](https://github.com/web-scrobbler/web-scrobbler/pull/3236/files#diff-c2bab7037f048bfd48172ff023ad63f8f071fb27c4de14f64f6c329f8f57760bR143) should include the class `.playing` which is added to the DOM element of the album being played: 
`document.querySelector('.playing .buy-now a').href`

PR #3250, opened 2022-05-09
Fix originUrl for Bandcamp collections

MonkeyDo:
Added the `.playing` class to the DOM element selector which targets the currently playing element

**Additional context**
Fixing issue #3249, more context on the issue description


MonkeyDo:
Sorry, I got a bit confused, this is not going to work properly.

This modification will make the wishlist page work, but it will break the collections page.
Both the wishlist and collections page use the same `initPropertiesForCollectionsPlayer` but each page requires a different DOM selector to target the right element.

Issue #3246, opened 2022-05-05
Incorrect Bandcamp originUrl when scrobbling on the Feed page


MonkeyDo:
**Describe the bug**
Following #3236 (originUrl for "collections" page), the same issue happens on other Bandcamp pages.
Each page seems to need a different DOM selector to get the appropriate album URL.

**How to reproduce**
1. Go to your Bandcamp feed page (https://bandcamp.com/MY_USER_NAME/feed)
2. Click on  any play icon on the page to play an album
3. The scrobble's originUrl is set as `https://bandcamp.com/MY_USER_NAME/feed`

**Expected behavior**
Ideally, the originUrl should reflect the URL of the album being played rather than the current page the user is on.

**Debug logs**
```
Web Scrobbler: {
  "track": "K.K. Gizzard",
  "artist": "par Louie Zong",
  "album": null,
  "albumArtist": null,
  "uniqueID": null,
  "duration": null,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://f4.bcbits.com/img/a3048034173_4.jpg",
  "isPodcast": false,
  "originUrl": "https://bandcamp.com/mr_monkey/feed?from=menubar"
}
```

**Environment (please complete the following information)**
-   OS: Mac OS
-   Browser: Brave
-   Extension version: 2.62.0

**Additional context**
The selector added in  #3236 only works for the "Collection" page (for example `https://bandcamp.com/mr_monkey`).
For the feed page specifically, the following selector targets the album currently being played (both in the main content of the page and the sidebar):
`document.querySelector('.playing .buy-now a').href`

PR #3253, opened 2022-05-09
Improve Bandcamp originUrl

### LINKED ISSUE #3251: Incorrect originUrl when scrobbling on the Bandcamp home page
**Describe the bug**

**How to reproduce**
1. Go to the [Bandcamp homepage](https://bandcamp.com/)
2. Click to play any playable element on the page
3. See the `originUrl` in the scrobble

**Expected behavior**
The originUrl should be a URL to the album being played rather than the generic homepage URL.

**Debug logs**
See the generic `originUrl`:
```
Web Scrobbler: {
  "track": "March of the Dead (Intro feat. John Carpenter and Cody Carpenter)",
  "artist": "DANCE WITH THE DEAD",
  "album": "Driven to Madness",
  "albumArtist": null,
  "uniqueID": null,
  "duration": null,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://f4.bcbits.com/img/a1386566724_2.jpg",
  "isPodcast": false,
  "originUrl": "https://bandcamp.com/?from=menubar_logo_logged_in"
}
```

-   OS:  Mac OS
-   Browser: Brave
-   Extension version: 2.62.0

**Additional context**
There are two types of playable elements on the Bandcamp home page.
One type is tied to a music player on the page while the other is not. So using the music player to get the album URL won't work for both.
However, the DOM elements all get the `.playing` css class added when they are being played (so that they can show a play/pause icon over the album art). This allows us to target whatever the currently playing item might be.

Albums in the "New and notable" section can be targeted with `document.querySelector('.playing + .item-title a').href` (note the  anchor element is a child of the `item-title` element):
image

On the other hand, in the "discover" section at the bottom of the page, the album links can be targeted with `document.querySelector('.playing + .item-title').href` (note the  `item-title` element itself is the anchor element).
image

inverse:
Resolves #3251  #3249 #3246

@MonkeyDo thanks for raising the tickets - this should fix two of the three!

### DIFF
--- a/src/connectors/bandcamp.js
+++ b/src/connectors/bandcamp.js
@@ -139,15 +139,7 @@ function initPropertiesForCollectionsPlayer() {
 
 	Connector.trackArtSelector = '.now-playing img';
 
-	Connector.getOriginUrl = () => {
-		const buyNowAnchor = document.querySelector('.buy-now a');
-		if (buyNowAnchor === null) {
-			Util.debugLog('Failed to resolve originUrl');
-			return document.location.href;
-		}
-
-		return buyNowAnchor.href.split('?')[0];
-	};
+	Connector.getOriginUrl = () => getOriginUrl('.buy-now a');
 }
 
 // https://bandcamp.com/%YOURNAME%/feed
@@ -163,6 +155,8 @@ function initPropertiesForFeedPlayer() {
 	Connector.trackArtSelector = '#track_play_waypoint img';
 
 	Connector.playButtonSelector = '#track_play_waypoint.playing';
+
+	Connector.getOriginUrl = () => getOriginUrl('.playing .buy-now a');
 
 	function removeByPrefix(text) {
 		return text.replace('by ', '');
@@ -213,6 +207,9 @@ function initPropertiesForHomePage() {
 
 		return null;
 	};
+
+
+	Connector.getOriginUrl = () => getOriginUrl('.playable.playing');
 }
 
 function isAlbumPage() {
@@ -230,7 +227,6 @@ function isFeedPage() {
 function isCollectionsPage() {
 	return document.querySelector('#carousel-player') !== null;
 }
-
 
 function getTrackNodes() {
 	let trackNodes = [];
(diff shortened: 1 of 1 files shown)


src/connectors/bandcamp.js
```diff
@@ -213,6 +207,9 @@ function initPropertiesForHomePage() {
 
 		return null;
 	};
+
+
+	Connector.getOriginUrl = () => getOriginUrl('.playable.playing');
```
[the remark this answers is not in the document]
inverse: I applied the change - let me know how it works?

image

 noticed this section doesn't respond to changes once initially played - but I think that's another issue unrelated to this.

src/connectors/bandcamp.js
```diff
@@ -139,15 +139,7 @@ function initPropertiesForCollectionsPlayer() {
...
-	Connector.getOriginUrl = () => {
-		const buyNowAnchor = document.querySelector('.buy-now a');
-		if (buyNowAnchor === null) {
-			Util.debugLog('Failed to resolve originUrl');
-			return document.location.href;
-		}
-
-		return buyNowAnchor.href.split('?')[0];
-	};
+	Connector.getOriginUrl = () => getOriginUrl('.buy-now a');
```
MonkeyDo, line 142: This should cover the 'collections' tab as well as the 'wishlist' tab (which are on the same page)

```suggestion
	Connector.getOriginUrl = () => getOriginUrl('.playing .collection-title-details a, .playing .buy-now a');
```

MonkeyDo (commented):
I've tested the PR locally, scrobbling on the home page now works as expected (apart from, as you said, an unrelated issue with the "new and notable" section).
So that's issue #3251 fixed!

One other issue now fixed with this PR is #3246 which isn't in the PR description.
Feed page working fine !

Regarding issue #3249, we have a similar situation to the home page where there are two different types of elements a user can play: one on the "collections" tab and one on the "wishlist" tab. (changing tabs doesn't trigger a page load so detecting which tab we're on would be tricky; currently the `isCollectionsPage` method returns true for either tab).

Considering the fallback selector idea seems to be working nicely, I'm suggesting another double-selector below to fix the wishlist tab.

Thanks again :)

src/connectors/bandcamp.js
inverse: Looks like we can even simplify it

inverse: Actually it didn't work and applied your full fix - thanks :pray:

MonkeyDo:
Thank you so much !

PR #3201, opened 2022-03-31
Add .gitattributes to prevent conversion to CRLF

RudeySH:
I've added a .gitattributes file with `* eol=lf` to force Git to normalize line endings to LF on checkin and prevent conversion to CRLF when files are checked out.

**Additional context**

The default config of Git for Windows is in conflict with web-scrobbler's lint.

- Git's "core.autocrlf" setting allows developers on Windows to automatically convert line endings to CRLF when checking out files, and normalizes line endings back to LF on checkin. While core.autocrlf is set to false by default, **the Git for Windows installer sets core.autocrlf to true by default**. Effectively, most developers on Windows will get CRLF line endings when checking out files.

- CRLF line endings lead to warnings in web-scrobbler's lint (specifically, [lintspaces](https://github.com/schorfES/node-lintspaces)), and converts CRLF back to LF.

As a result of of using both core.autocrlf=true and lintspaces, checking out web-scrobbler on Windows and then running lint will instantly result in hundreds of thousands of warnings, and hundreds of modified files.

<details>
<summary>View output</summary>

```
$ npm run lint

> web-scrobbler@2.57.2 lint
> grunt lint

Running "eslint:target" (eslint) task

Running "jsonlint:src" (jsonlint) task
>> 11 files lint free.

Running "lintspaces:src" (lintspaces) task
☓ package-lock.json
  L1: Incorrect end of line character(s) found.
  L2: Incorrect end of line character(s) found.
  L3: Incorrect end of line character(s) found.
  L4: Incorrect end of line character(s) found.
  L5: Incorrect end of line character(s) found.
  L6: Incorrect end of line character(s) found.
  L7: Incorrect end of line character(s) found.
  L8: Incorrect end of line character(s) found.
  L9: Incorrect end of line character(s) found.
  L10: Incorrect end of line character(s) found.
  L11: Incorrect end of line character(s) found.
  L12: Incorrect end of line character(s) found.
  L13: Incorrect end of line character(s) found.
  L14: Incorrect end of line character(s) found.
  L15: Incorrect end of line character(s) found.
  L16: Incorrect end of line character(s) found.
  L17: Incorrect end of line character(s) found.
  L18: Incorrect end of line character(s) found.
  L19: Incorrect end of line character(s) found.
  L20: Incorrect end of line character(s) found.
  L21: Incorrect end of line character(s) found.
  L22: Incorrect end of line character(s) found.
  L23: Incorrect end of line character(s) found.
  L24: Incorrect end of line character(s) found.
[...]

PR #3263, opened 2022-05-18
Fix .gitattributes to only change EOL to LF in code files

eugenesvk:
Limit end of line normalization to code files (`.css` `.html` `.js` `.json`) so that `src\icons\*.png` files are not viewed as changed even when they aren't

**Additional context**
https://github.com/web-scrobbler/web-scrobbler/pull/3201 added 
> a .gitattributes file with * eol=lf to force Git to normalize line endings to LF on checkin and prevent conversion to CRLF when files are checked out.

However, on Windows even with `[core] autocrlf=false` git config this leads to the `src\icons\*.png` being viewed as changed even though they aren't, thus, e.g., preventing switching branches
This PR implements the more limited option mentioned in https://github.com/web-scrobbler/web-scrobbler/pull/3201#issuecomment-1085445586 by applying the EOL normalization to code file extensions only (`.css` `.html` `.js` `.json`)


RudeySH:
After #3201 got merged, I also ran into this issue with .png files. I've been meaning to fix it via #3222 (alongside a new feature I'm working on), but that PR isn't going to be merged any time soon (the new feature isn't done). In that PR, I fixed the .gitattributes file by adding `text=auto` like so:

```
* text=auto eol=lf
```

I **strongly** prefer `text=auto` over limiting the .gitattributes file to .css, .html, .js and .json only, which is what your PR achieves. My reasoning for this is as follows. While these 4 file extensions are the only file types that lint converts the line endings for (via lintspaces), web-scrobbler also uses another tool that converts the line endings for all text files: EditorConfig. The .editorconfig file in this repository contains the following:

```
[*]
end_of_line = lf
```

This will convert line endings to LF when saving any text file in a text editor that supports EditorConfig. This includes files that arent .css, html, .js and .json, such as .md and .yml files.

By using `text=auto`, Git will only convert the line endings for files that it considers to be text files, which I can only assume will closely match what EditorConfig considers to be text files as well.

@eugenesvk Do you agree? If so, should I make a new PR for `text=auto` or will you update this PR?

eugenesvk:
Ok, I've updated the PR with your broader text-based suggestion

PR #3275, opened 2022-06-06
Update donation button in the extension options

### LINKED ISSUE #3274: Support button in the options page still leads to Ko-fi
**Describe the bug**
Recently we moved to OpenCollective to receive donations, but we still use Ko-fi button in the extension settings.

We should replace it with OpenCollective button.

alexesprit:
Replace Ko-fi support button with OpenCollective support button.

Closes #3274.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -326,6 +326,10 @@
         "message": "The extension is free, and will always be free. But if you want to support development of the extension, you can use the button below.",
         "description": ""
     },
+    "donationButtonLabel": {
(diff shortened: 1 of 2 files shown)


alexesprit:
As an alternative option we could use OpenCollective's buttons, but I found no way to customize them to use brand color, so I decided to do not use the buttons.

What's your thoughts on this?

<details>
  <summary><b>OpenCollective buttons</b> (Click to expand)</summary>
Blue button

White button
</details>

inverse (approved):

PR #3278, opened 2022-06-09
Send more additional information to listenbrainz

badlandspray:
As per API [spec](https://listenbrainz.readthedocs.io/en/latest/users/json.html#payload-json-details), send `duration` and `spotify_id` in `additional_info`

### DIFF
--- a/src/core/background/scrobbler/listenbrainz-scrobbler.js
+++ b/src/core/background/scrobbler/listenbrainz-scrobbler.js
@@ -284,6 +284,14 @@ define((require) => {
 				trackMeta.additional_info.release_artist_name = song.getAlbumArtist();
 			}
 
+			if (song.getUniqueId() && song.metadata.label === "Spotify") {
+				trackMeta.additional_info.spotify_id = song.getUniqueId();
+			}
+
+			if (song.getDuration()) {
+				trackMeta.additional_info.duration = song.getDuration();
+			}
+
 			return trackMeta;
 		}
 	}
--- a/src/core/background/scrobbler/listenbrainz-scrobbler.js
+++ b/src/core/background/scrobbler/listenbrainz-scrobbler.js
@@ -284,7 +284,7 @@ define((require) => {
 				trackMeta.additional_info.release_artist_name = song.getAlbumArtist();
 			}
 
-			if (song.getUniqueId() && song.metadata.label === "Spotify") {
+			if (song.getUniqueId() && song.metadata.label === 'Spotify') {
 				trackMeta.additional_info.spotify_id = song.getUniqueId();
 			}
 
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   "eslintConfig": {
     "extends": "web-scrobbler",
     "parserOptions": {
-      "sourceType": "script"
+      "sourceType": "script",
+      "ecmaVersion": 2020
     }
   },
   "prettier": "prettier-config-web-scrobbler",
--- a/src/connectors/spotify.js
+++ b/src/connectors/spotify.js
@@ -100,7 +100,7 @@ function getTrackUri() {
 
 	const url = new URL(contextLinkEl);
 	const trackUri = url.searchParams.get('uri');
-	const trackId = trackUri.split(':')?.[2];
+	const trackId = trackUri?.split(':')?.[2];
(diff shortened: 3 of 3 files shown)


jaccarmac:
I'm not a user of the Listenbrainz feature and largely unfamiliar with the core of the codebase, but this looks pretty reasonable.

inverse:
Thanks for contributing @badlandspray and thanks for the review @jaccarmac 

We're lucky enough that we have engineers from Listenbrainz who are active in this community.

@MonkeyDo Would you be able to check this one?

MonkeyDo:
Absolutely! I'm AFK this coming week but count on me once I'm back.

badlandspray:
@MonkeyDo Would you able to review it now?

src/core/background/scrobbler/listenbrainz-scrobbler.js
```diff
@@ -284,6 +284,14 @@ define((require) => {
 				trackMeta.additional_info.release_artist_name = song.getAlbumArtist();
 			}
 
+			if (song.getUniqueId() && song.metadata.label === 'Spotify') {
+				trackMeta.additional_info.spotify_id = song.getUniqueId();
```
inverse, line 288: This would be something like

```
spotify:track:0VfQeEvwYx4NXuGXkbWgUb
```

@MonkeyDo is that correct? 

The spec outlines

>  The Spotify track URL associated with this recording. e.g.: http://open.spotify.com/track/1rrgWMXGCGHru5bIRxGFV0

src/core/background/scrobbler/listenbrainz-scrobbler.js
```diff
@@ -284,6 +284,14 @@ define((require) => {
...
+			}
+
+			if (song.getDuration()) {
+				trackMeta.additional_info.duration = song.getDuration();
```
MonkeyDo, line 292: We now [clarified](https://listenbrainz.readthedocs.io/en/latest/users/json.html#payload-json-details) that we accept either `duration` or `duration_ms`, the former for durations in seconds and the latter (which I would recommend) for durations in milliseconds

src/core/background/scrobbler/listenbrainz-scrobbler.js
MonkeyDo, line 288: We have improved the metadata section of the listen and made them a bit more useful (see reference [here](https://listenbrainz.readthedocs.io/en/latest/users/json.html#brainzplayer-on-the-listenbrainz-website-playing-a-video-from-spotify))

The field `track_metadata.additional_info.origin_url` should also be added, in this case with the same value as the `spotify_id` field

MonkeyDo (commented):
Sorry about the long delay!
We've been finalizing some changes to the suggested structure of the track metadata payload (https://github.com/metabrainz/listenbrainz-server/pull/1607, https://github.com/metabrainz/listenbrainz-server/pull/1832), I figured it would be annoying to validate the PR to then propose some extra changes !

src/core/background/scrobbler/listenbrainz-scrobbler.js
MonkeyDo: Indeed the `spotify_id` and the `origin_url` fields should both be a fully-qualified rather than the `spotify:track:…` URN

src/core/background/scrobbler/listenbrainz-scrobbler.js
inverse: What's the rational for using `ms`? internally the connector uses seconds for duration so we would have the same precision if we multiplied this value up?

badlandspray:
Looks like the release been cut, I was hoping to get this in the release. Have to wait for one more month for the next release?

inverse:
@badlandspray You hopefully won't have to wait that long - releases are pretty automated so would be just a case of getting an approval and running a few commands.

src/core/background/scrobbler/listenbrainz-scrobbler.js
MonkeyDo: In that case `duration` is the correct one, sorry if I wasn't clear.
I just wasn't sure of the internal implementation and wanted to make sure the right field is used.

inverse (approved):

Issue #2020, opened 2019-07-25
Add Nugs.net support


aybesea:
**Provide information about website**
Nugs.net is a streaming service that covers about 100 jam bands, notably Dead & Co, Widespread Panic, Pearl Jam, Gov't Mule and Umphrey's McGee. This is very closely related to the LivePhish request that I put in earlier. Granted, it is a separate service, but both are run by Nugs and both use extremely similar software apps.

**Additional context**
A free trial is available at www.nugs.net

bsutich:
I have an account for the site, and would love to help in any testing of the scrobbling tool. Currently using macOS. Can test on Safari, Chrome, Firefox, etc.

circlecube:
I set up a PR to add support for nugs.net scrobbling, hopefully we can get it merged in soon!

inverse:
Should be in the recently published release. 

Thanks @circlecube

PR #3294, opened 2022-07-01
Feature/add nugs connector

circlecube:
Adding scrobbler connector for nugs.net player.

**Additional context**
I realized my scrobbled had gone way down since I have been listening on nugs more often lately. I didn't realize it wasn't supported and found a ticket (https://github.com/web-scrobbler/web-scrobbler/issues/2020) requesting support already, so I figured I try my hand and a connector for it. This is my first contribution to web-scrobbler.

### DIFF
--- a/src/connectors/nugs.js
+++ b/src/connectors/nugs.js
@@ -2,7 +2,7 @@
 
 Connector.playerSelector = 'div[app-player] .player-wrapper';
 
-Connector.playButtonSelector = '.play-toggle';
+Connector.pauseButtonSelector = '.player-controls .pause-toggle';
 
 Connector.artistSelector = '.player-info .artist-name';
 
--- a/src/connectors/nugs.js
+++ b/src/connectors/nugs.js
@@ -1,0 +1,19 @@
+'use strict';
+
+Connector.playerSelector = 'div[app-player] .player-wrapper';
+
+Connector.playButtonSelector = '.play-toggle';
+
+Connector.pauseButtonSelector = '.pause-toggle';
+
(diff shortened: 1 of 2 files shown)


yayuyokitano:
Thank you for your contribution!

Unable to test this as I do not have an account, but it looks sound so I will trust that the selectors are correct

src/connectors/nugs.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+Connector.playerSelector = 'div[app-player] .player-wrapper';
+
+Connector.playButtonSelector = '.play-toggle';
```
yayuyokitano, line 5: Actually, spoke too soon here.

Here, you should avoid using both playbuttonselector and pausebuttonselector. Please remove one of the two.

As soon as that's done this can be merged I think.

circlecube: Ah, I wondered about that. Is one preferred over the other? It's the same button but when playing the pause class toggles, but the okay class remains all the time.

yayuyokitano: As far as I am aware, there is no preference of one to the other.
The reason both exist is in case only one of the two can be used on a site, using both at once is redundant.
Just verify that it still works with the one you stick with and it should be good.

circlecube: I first removed the pause, but after testing realized it wasn't picking up anymore, so swapped it back for the pause button and I'm seeing the scrobbled. I'll consider this resolved.

inverse (approved):
LGTM :+1:

PR #3295, opened 2022-07-01
Update playlist-randomizer connector

### LINKED ISSUE #3293: playlist-randomizer selectors need updating
**Describe the bug**
Some time ago, the dev of playlist-randomizer pushed an update that breaks some of the selectors on this connector.

The simple one is that `.MuiAvatar-circle` changed to `.MuiAvatar-circular` for the trackArtSelector, but the other issue is that the channel name is now getting appended to the track name:

**Screenshots**
In the playlist:

In web scrobbler:

This one should be track "Cosmo Funk" with artist "Snail's House"

And in cases where the artist name isn't in the video title, it's using the channel name for the track name:

This one should be track "Vision" with artist "Pop Up!"

And without any dashes in either the title or channel name, it puts both the title and channel name for both the track and artist:

This one should be track "POGO Live 18.0.1.19" with artist "Pogo"

Californ1a:
Updated the playlist-randomizer connector to work since the dev changed some styles. The "www." was also dropped from the URL, so I added the version without it.

I'm not sure if there's something in the Util to better address "Artist - Topic" (provided by youtube) channel names, it looked like the other connectors were getting around that by using the video description rather than the channel name, which isn't available on this site.

One small bug I'm not sure how to fix, if the page is refreshed during play (or the user goes directly to a playlist link rather than via the homepage), the first song isn't scrobbled but all the rest will be.

**Additional context**
Fixes #3293

### DIFF
--- a/src/connectors/playlist-randomizer.js
+++ b/src/connectors/playlist-randomizer.js
@@ -23,3 +23,4 @@ Connector.getArtistTrack = () => {
 };
 
 Connector.applyFilter(MetadataFilter.getYoutubeFilter());
+Connector.onReady = Connector.onStateChanged;
--- a/src/connectors/playlist-randomizer.js
+++ b/src/connectors/playlist-randomizer.js
@@ -2,9 +2,20 @@
 
 Connector.playerSelector = '#root';
 Connector.trackArtSelector = '.Mui-selected .MuiListItemAvatar-root .MuiAvatar-circular .MuiAvatar-img';
-Connector.playButtonSelector = 'button[title="Play"]';
 
 const trackSelector = '.Mui-selected .MuiListItemText-root .MuiListItemText-primary';
+
+Connector.isPlaying = () => {
+	const playButton = document.querySelector('button[title="Play"]');
+	const tooltip = document.querySelector('.MuiTooltip-popper');
+	const tooltipText = tooltip ? tooltip.innerText : '';
+
+	if (playButton || tooltipText.includes('Play')) {
+		return false;
+	}
+
+	return true;
+};
 
 Connector.getArtistTrack = () => {
 	const currentTrack = document.querySelector(trackSelector);
@@ -13,8 +24,7 @@ Connector.getArtistTrack = () => {
 	}
 	let { artist, track } = Util.processYtVideoTitle(currentTrack.firstChild.nodeValue);
 
-	// Set to some default information that we have (probably "Song Title Song Artist" with a space)
-	// so that the user can edit the info in the extension
+	// Set to some default information that we have so that the user can edit the info in the extension
 	if (!artist) {
 		const regex = /^(.*) - Topic$/;
 		artist = currentTrack.lastChild.innerText.replace(regex, '$1');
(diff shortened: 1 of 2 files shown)


yayuyokitano:
Thank you for your contribution!

I think this way of getting around topic is reasonable enough personally.

As for your other issue, simply append `Connector.onReady = Connector.onStateChanged;` to the end of the connector, and it should work.

Californ1a:
> As for your other issue, simply append `Connector.onReady = Connector.onStateChanged;` to the end of the connector, and it should work.

Thanks, that did it.

Californ1a:
Would it be bad practice to hijack the react app's internal state? Replacing this:
```js
Connector.playButtonSelector = 'button[title="Play"]';
```
with
```js
Connector.isPlaying = () => {
  return document.querySelector("#root")._reactRootContainer._internalRoot.current.memoizedState.element.props.store.getState().player.isPlaying;
}
```
There's a tiny bug with using the play button selector where the `title` attribute gets removed from the button while you're hovering over it(and a separate tooltip element gets added to the body), so for a brief period of time between hovering over the play button and clicking on it, WS thinks the song is already playing. Probably no one would ever notice during actual usage, but it feels kind of wrong to try to access react state outside of react like that especially for such a small issue.

yayuyokitano:
> Would it be bad practice to hijack the react app's internal state? Replacing this:
> 
> ```js
> Connector.playButtonSelector = 'button[title="Play"]';
> ```
> 
> with
> 
> ```js
> Connector.isPlaying = () => {
>   return document.querySelector("#root")._reactRootContainer._internalRoot.current.memoizedState.element.props.store.getState().player.isPlaying;
> }
> ```
> 
> There's a tiny bug with using the play button selector where the `title` attribute gets removed from the button while you're hovering over it (and a separate tooltip element gets added to the body), so for a brief period of time between hovering over the play button and clicking on it, WS thinks the song is already playing. Probably no one would ever notice during actual usage, but hijacking the state would fix it, though it feels kind of wrong to try to access react state outside of react like that especially for such a small issue.

I don't think this is that bad, personally. It's not modifying the state, so I don't see how this could cause anything bad to happen.
Might be a bit unstable though, but that can be fixed if anything happens.

Californ1a:
As it turns out, `_reactRootContainer` isn't actually accessible from within the context of the content script, only within the page itself. So instead, I've just added a selector for the tooltip and returned false if either the play button or play tooltip are visible. It fixes hovering over the play button for a long duration counting as the song being played the entire time you're hovering over the button, though there is a brief period between hovering over the play button and the tooltip being added to the dom that gets counted as playing, but the mutation observer picks up on the tooltip being added and sets it back to false. That's probably good enough, so unless someone else wants to take a look at the play button hovering, then this is probably good to go.

Californ1a:
Is there something blocking this from being merged? I thought it was done 2 months ago.

inverse:
Sorry for leaving this one open for so long. 

I don't see any issues with it.

inverse (approved):

Issue #2943, opened 2021-07-29
Regex-based filters


awsms:
Recently, another great scrobbling service I use on my smartphone added a way to remove unwanted parts to scrobble thanks to regexes.
For example, removing ' (Album Version)', ' (Original Mix)' etc from the tracknames; or letting the user set custom filters for the album and artist names. As you imagine, manually editing every single 'Original Mix' is quite redundant.

I think it would be great if web-scrobbler could implement something similar.

jaccarmac:
Which scrobbler does this? (Interested both to see the implementation and for personal use.)

afoxie:
This would be an amazing feature, as it stands right now I have to manually edit (almsot) every YT Music track to remove the "(feat. xxxxx)

awsms:
> Which scrobbler does this? (Interested both to see the implementation and for personal use.)

It's https://github.com/kawaiiDango/pScrobbler

> This would be an amazing feature, as it stands right now I have to manually edit (almsot) every YT Music track to remove the "(feat. xxxxx)

Same here..

djmattyg007:
Doesn't web-scrobbler already do this? The key functionality was spun out into a library that you can install and use separately:

https://github.com/web-scrobbler/metadata-filter

I've also ported it to python:

https://github.com/djmattyg007/music-metadata-filter

flickur:
> Doesn't web-scrobbler already do this? The key functionality was spun out into a library that you can install and use separately:
> 
> https://github.com/web-scrobbler/metadata-filter
> 
> I've also ported it to python:
> 
> https://github.com/djmattyg007/music-metadata-filter


Can this be implemented in the scrobbler addon itself with an UI for easier use?

flickur:
Bumping this since I've been using pano scrobbler on my phone for some time now and the pattern filter is awesome. It would fix most of my problems with webscrobbler. Please make this happen.

djmattyg007:
It isn't clear exactly what you're asking for. The `metadata-filter` package is already integrated into `web-scrobbler`.

Are you after the ability to define custom patterns?

flickur:
Same thing OP is suggesting. Ability to set pattern filters that apply to all your scrobbles. There is no such function in the webscrobbler options, only the list of your edited single tracks. Here's an example from pano scrobbler where it gets rid of every "- radio edit" that appears in the track name.

[...]

PR #3306, opened 2022-07-17
Only scrobble one artist in YouTube Music

AvikRao:
Only scrobble the first artist in Youtube Music. Last.fm, the biggest music tracking service, only allows a single artist to be credited for each song. When the Web Scrobbler scrobbles multiple artists as a comma-separated string, it breaks compatibility with many of Last.fm's social features, such as identifying which songs and artists you share in common with other users. To maintain compatibility with Last.fm's single-artist paradigm for each song, the change made in this PR only scrobbles the first artist credited on a song, rather than comma-separating the entire list. 

**Additional context**
Example of Web Scrobbler scrobbling Phoenix by "League of Legends, Cailin Russo, Chrissy Costanza":
image
The actual song's entry in Last.fm, only by League of Legends:
image

By observing the global scrobble count, clearly the single-artist entry is the correct one. This is not a cherrypicked example - every example I have searched for with multiple artists follows the same pattern.

### DIFF
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -110,8 +110,7 @@ Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
 
 function getArtists() {
 	// FIXME Use Array.from after jQuery support will be removed
-	const artistElements = Util.queryElements(artistSelectors);
-	return artistElements && Util.joinArtists(artistElements.toArray());
+	return Util.queryElements(artistSelectors).toArray()[0]?.textContent;
 }
 
 function filterYoutubeIfNonAlbum(text) {
--- a/src/connectors/youtube-music.js
+++ b/src/connectors/youtube-music.js
@@ -79,14 +79,14 @@ Connector.getArtistTrack = () => {
 	let track;
 
 	if (hasVideoAlbum()) {
-		artist = getArtists();
+		artist = getArtist();
 		track = Util.getTextFromSelectors(trackSelector);
 	} else {
 		({ artist, track } = Util.processYtVideoTitle(
 			Util.getTextFromSelectors(trackSelector)
 		));
 		if (!artist) {
-			artist = getArtists();
+			artist = getArtist();
 		}
 	}
 	return { artist, track };
@@ -108,8 +108,7 @@ Connector.getUniqueID = () => {
 
 Connector.isScrobblingAllowed = () => !Util.isElementVisible(adSelector);
 
-function getArtists() {
-	// FIXME Use Array.from after jQuery support will be removed
+function getArtist() {
 	return Util.queryElements(artistSelectors).toArray()[0]?.textContent;
 }
 


inverse (changes requested):
Looks good except we should probably change the function name to `getArtist()` to better represent. Also I think FIXME comment is no-longer valid :)

jaccarmac:
> By observing the global scrobble count, clearly the single-artist entry is the correct one. This is not a cherrypicked example - every example I have searched for with multiple artists follows the same pattern.

Two notes:

1. I don't think this is correct. It's certainly true that, in some sense "Last.fm ... only allows a single artist to be credited". But tagging collaborative work with both artists is widespread and not strictly incorrect.
2. If we want to make this change, it seems better to do it for all connectors, as it's not a YouTube Music specific problem.

To be clear, I'm against it on general grounds, but think it's especially incorrect to implement only for YouTube Music. (I don't think this would be a good case for an option and would rather see something like #2943 to allow users to customize the extension behavior. However, I understand that that is a less attractive solution.)

jaccarmac (commented):
If this behavior changes, it should change globally and not only for YouTube Music. (Trivially, the YouTube connector should use the same rule.)

a-github-account:
I think this should be a togglable setting because other services that aren't Last.fm might utilize the multiple-artists approach.

inverse (approved):

inverse:
@jaccarmac do you have any further input on this one? I guess you're right that it should be made into something more generic and perhaps controlled like @a-github-account suggested.

jaccarmac:
Option or no, I maintain that the switch to single-artist behavior is incorrect if it happens only on YouTube Music. In the interest of the work that went into this PR, though, perhaps we could meet in the middle with an option which changes the behavior of YouTube *and* YouTube Music but no other connectors?

yayuyokitano:
Closing, as #3671 allows the user to achieve this if they want to in a controllable way that works across all connectors.

Lamby777:
> Closing, as #3671 allows the user to achieve this if they want to in a controllable way that works across all connectors.

How...? If you mean using regex to pick everything before the & sign, this ruins full artist names like `Bob Marley & The Wailers` or `King Gizzard & The Lizard Wizard` ([what did they do to the lizard??? D:](https://www.last.fm/music/King+Gizzard)). This problem is on the connector level, can't be fixed once the artists are combined...

I understand scrobbling the full list, and I used to believe this was the best option too, but this makes scrobbling certain artists that do features a lot (Gorillaz, Daft Punk, etc.) really skewed and pretty much pointless. Not to mention different platforms combining lists of artists differently ([here's](https://www.last.fm/music/Lotus+Juice,+Azumi+Takahashi,+&+ATLUS+Sound+Team/_/Mass+Destruction+-Reload-) a [game](https://www.last.fm/music/Lotus+Juice,+Azumi+Takahashi,+ATLUS+Sound+Team/_/Mass+Destruction+-Reload-) of [spot the difference](https://www.last.fm/music/Lotus+Juice,+Azumi+Takahashi+&+ATLUS+Sound+Team/_/Mass+Destruction+-Reload-), and of course [this one](https://www.last.fm/music/Lotus+Juice/_/Mass+Destruction+-Reload-) has actual comments). We should be given a setting to do what this PR does, not just assume & means another artist when the YT Music site cleanly separates multiple `<a>` tags for us when necessary. :/

PR #3307, opened 2022-07-21
Send album and album artist data to maloja

krkk:
notes:
- `albumartists` accepts a list of artists[^1]. However, this will make a list with a single item, with all the artists joined together. it feels ‘hacky’ but i think that’s the best we can get at the moment.
- i didn’t include the `duration` because from what i've seen, `song.getDuration()` uses a duration of a song instead of duration of how long were you’ve been listening to it, as the api suggests.

[^1]: majola api: https://github.com/krateng/maloja/blob/master/API.md#submitting-a-scrobble


inverse (approved):
Thanks :+1:

Issue #2016, opened 2019-07-17
Add support for LivePhish


aybesea:
**Provide information about website**
LivePhish is a paid service that streams all live Phish performances. A free trial is available at https://www.livephish.com/

Thank you!

aybesea:
I would be willing to pay for a month for a developer if they need it to get this working!

goldwave315:
this would be a great addition and much appreciated!

inverse:
It's in the next release - thanks @goldwave315

PR #3312, opened 2022-07-28
Feature/add livephish.com connector

goldwave315:
I added the scrobbler connector for plus.livephish.net

**Additional context**
I saw @inverse had added a connector for nugs.net, livephish.com is run by nugs.net so I thought it might be easy to duplicate his effort and add a connector for livephish.com and it was. There is also an existing support ticket for this [@2016](https://github.com/web-scrobbler/web-scrobbler/issues/2016).


alexgandy:
👍

PR #3315, opened 2022-08-03
Fix Tidal duration selector.

jaccarmac:


jaccarmac:
I'm not sure if this is a staged rollout. Anyone else using Tidal and seeing correct duration detection? (I was seeing scrobbles happen consistently at 30 seconds as the scrobbler could no longer find the song length.)

inverse:
Changes seem legit to me - Wonder if other Tidal users can comment... happy to merge and release either way

inverse (approved):

Issue #3321, opened 2022-08-10
Hoopla connector is broken


thomwim:
I'm trying to play the album "Chopped & Screwed" by Korn, which I borrowed from Hoopla. I have posted a picture of the track player.

https://www.hoopladigital.com/title/11250844

![image](https://user-images.githubusercontent.com/7088284/183870197-cd9ca7f3-9c3c-4215-9dec-5b8fecb586ee.png)

inverse:
@hornbuckle would you be able to take a look at this issue?

hornbuckle:
@inverse Sure. Looks like they may have updated their player, so I'll take a look and try to get it working again in the next few days.

inverse:
No rush - glad to have some help on this one :pray:

inverse:
Should be shortly available :+1:

Amraan-Amjad:
@hornbuckle @inverse 
Again, it is down 
I was trying to download the audiobook but it showed
![image](https://github.com/web-scrobbler/web-scrobbler/assets/121578485/7a427fe6-e5bb-4b70-bf5b-4c6bdc5caf0e)
Kindly update it

hornbuckle:
@Amraan-Amjad It does not support audiobook scrobbling; only music.

Amraan-Amjad:
> @Amraan-Amjad It does not support audiobook scrobbling; only music.

Could you add it please :)

hornbuckle:
@Amraan-Amjad No, it does not work. I tried it before.

PR #3207, opened 2022-04-05
Add connector for hoopladigital.com

hornbuckle:
New connector for hoopla, a service provided by local libraries to "borrow" and stream audio. Tested with my personal account and scrobbles are successful. Screenshot below.

hooplascrobble

Additional edit made to account for dash separator appearing in album title. Also tested with my personal account. Screenshot below.

hooplascrobbledash

Resolves request in issue #1636

PR #3325, opened 2022-08-12
Update connector for hoopladigital.com

hornbuckle:
Connector for hoopla updated to work with new version of player. Tested with my personal account and scrobbles are once again successful. Screenshot below.

Screen Shot 2022-08-12 at 12 03 23 PM

Update to #3207. Resolves issue #3321.

### DIFF
--- a/src/connectors/hoopladigital.js
+++ b/src/connectors/hoopladigital.js
@@ -1,24 +1,25 @@
 'use strict';
 
-Connector.playerSelector = '.duration-1000';
+Connector.playerSelector = '.duration-700.z-100'; // player is destroyed when album ends
 
-Connector.trackSelector = '.text-current > h2';
+Connector.playButtonSelector = '.duration-700.z-100 button[aria-label=Play]';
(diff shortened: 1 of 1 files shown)


hornbuckle:
New Issue:
Album text only appears in expanded player (shown in screenshot) and will not be captured unless player is expanded. I believe the expanded player was previously only hidden but still available in the DOM, but now it's added/removed when expanded/collapsed.

Continued Issue:
Player is destroyed when album ends and requires a refresh of the page to reinitiate the scrobbler. I had previously asked about this but never received a reply: https://github.com/web-scrobbler/web-scrobbler/discussions/3214
Would it be possible to add a MutationObserver to reinitiate the scrobbler? And if so, what would be the best way to do that? I couldn't find any other examples of this being done in a connector.

inverse:
Thanks for getting this fixed. I don't get emails for discussions so perhaps that's how I missed it. 

Let me get back to you on that discussion

PR #3328, opened 2022-08-13
Update selectors for hoopladigital.com connector

hornbuckle:
Minor edits to updates made in #3325.

- Adjusted `playSelector` to parent element because the actual selector of the player is destroyed when album playback ends. This allows the connector to pick up scrobbling again when the player returns to the DOM after playback is initiated again, which previously required a page refresh.
- Also removed player class that was added to `playButtonSelector`, `trackSelector`, and `artistSelector`, as it is not required for them to work. In case the player wrapper class changes again, which caused issue #3321, best case is that those three most important selectors will continue to work until the others can be updated in the connector.

### DIFF
--- a/src/connectors/hoopladigital.js
+++ b/src/connectors/hoopladigital.js
@@ -1,12 +1,12 @@
 'use strict';
 
-Connector.playerSelector = '.duration-700.z-100'; // player is destroyed when album ends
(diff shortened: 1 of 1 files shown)


inverse (approved):

inverse:
Thank - will publish the release :+1:

PR #3335, opened 2022-08-19
Update connector for freegalmusic.com

hornbuckle:
Updates made to Freegal connector for more accurate scrobbling information as provided within the limited, persistent player. Tested with my personal account and scrobbles are successful. Screenshots below.

Issues Resolved:

- Null track title, other error minor error handling with `isScrobblingAllowed`.
- Some of Freegal's metadata is very messy and includes a CSV-type list of contributors as the artist, which can include the actual artist, a songwriter, a producer, or any other contributor. This makes scrobbling very tedious for these albums, as it requires continued manual updating of the information. Typically, the first artist in the list is the correct and only performer of the song, so the rest are filtered out.
- Additionally, with this extended artist info being combined with the track title in the main player, sometimes the title can be truncated partially or completely. Previous version split the artist and title from one location, but this update pulls the title from the playlist on the right side instead.

The incorrect metadata mostly seems to occur when the commas are not followed by a space, so any correct multi-artist metadata will still scrobble as expected. (Note the title completely truncated after the artist.)

Additionally, a single artist with a comma in their name will also scrobble as expected.

And lastly, 10,000 Maniacs (or any other artist with a numerical comma, if there are any) will scrobble as expected.

Continued Issues:

- Still no album title or artwork provided within the persistent player. (As seen in the previous screenshot, the album title and artwork do not match the album that is actually playing.)
- Really long titles are still truncated in both locations within the player, so those will need to be manually updated because the full title simply isn't provided.

- Additionally, there may be some edge cases where the artist will still need to be manually updated, depending on how the incorrect metadata is provided. But I tried to test as thoroughly as possible with artists I could find available on the site.

Current connector (from #1607) is still technically working, so no rush on releasing this.

### DIFF
--- a/src/connectors/freegalmusic.js
+++ b/src/connectors/freegalmusic.js
@@ -2,10 +2,34 @@
 
 Connector.playerSelector = '#player-section';
 
-Connector.artistTrackSelector = '.fp-title';
+Connector.pauseButtonSelector = '.fp-playbtn[title=Pause]';
 
-Connector.isPlaying = () => $('.fp-playbtn').prop('title') === 'Pause';
+Connector.trackSelector = '.playlist-player ul > li.active-playlist-song';
+
+Connector.getArtist = () => {
+	const artistTrackText = Util.getTextFromSelectors('.fp-message');
+
+	if (artistTrackText) {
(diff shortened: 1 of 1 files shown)


inverse (approved):
Looks a lot more complete then the original implementation - thanks!

PR #3336, opened 2022-08-19
Update selector for hoopladigital.com connector

hornbuckle:
Minor update to replace `playButtonSelector` with `pauseButtonSelector` for better `isPlaying` handling.

Button is sometimes briefly replaced by "Loading…" text, which can momentarily throw off `isPlaying` status when monitoring Play instead of Pause.

This is obviously very minor and current connector is working, so no rush on releasing this.


inverse:
Thanks!

PR #3356, opened 2022-09-09
Add new connector for KJazz 88.1 / Fix connector for Jazz24

hornbuckle:
Added a new connector for KJazz 88.1, whose main site is https://kkjz.org/
Streaming player is at https://player.listenlive.co/67941 so JS connector file is listenlive.js
Tested and confirmed scrobbling.

While I was adding that, I noticed that the connector for Jazz24 was broken.
Their main site is https://www.jazz24.org/
Streaming player is now at https://v7player.wostreaming.net/854 so updated JS connector file to wostreaming.js
Tested and confirmed scrobbling. (I also avoided any reference to "v7" in an attempt to future proof any new versions, assuming they're not a complete rebuild.)

Since both sites use a 3rd party streaming platform, it is likely that other stations are also using them and can therefore reuse the JS connector files, which is why they are not named specifically for each station.

Combined both edits because they both required updating connectors.js


hornbuckle:
**NOTE:** Jazz24 also has podcasts built into the same player page, and I attempted to rework the connector JS to include both streaming radio and podcasts. However, the podcasts use a slightly different version of the player controls, and I could not successfully get the connector to toggle between the two. Is there a way to toggle Connector.isPodcast and use that to toggle selectors? Or is there a better way to manage that? Definitely open to suggestions for a future update. If I should open a separate issue or discussion for this, I'd be glad to do so.

hornbuckle:
Found a few more stations using WO Streaming Player and added them to connectors.js

hornbuckle:
I mistakenly pushed my changes to my master instead of a branch. Should I close and resubmit as a branch?

hornbuckle:
Closing to resubmit as a branch, as I also have additional pull requests of separate issues to submit.

inverse:
> 
I mistakenly pushed my changes to my master instead of a branch. Should I close and resubmit as a branch?

It doesn't actually matter to much for creating PR's - just that it's harder to keep your fork clean and synced with this repo so its best to branch :)

PR #3362, opened 2022-09-13
Add new connector for KJazz 88.1

hornbuckle:
Added a new connector for KJazz 88.1, whose main site is https://kkjz.org/

Streaming player is at https://player.listenlive.co/67941 so JS connector file is listenlive.js

Tested and confirmed scrobbling.

### DIFF
--- a/src/connectors/listenlive.js
+++ b/src/connectors/listenlive.js
@@ -1,0 +1,9 @@
+'use strict';
+
+Connector.playerSelector = '#nowPlaying';
+
+Connector.artistSelector = '.nowPlayingCard .song .artist';
+
+Connector.trackSelector = '.nowPlayingCard .song .title';
+
+Connector.isPlaying = () => Util.hasElementClass('#nowPlaying', 'hasSong');
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -859,6 +859,14 @@ const connectors = [{
 	js: 'connectors/jazz24.js',
 	id: 'jazz24',
 }, {
+	label: 'KJazz 88.1',
+	matches: [
+		'*://kkjz.org/',
+		'*://player.listenlive.co/67941',
+	],
+	js: 'connectors/listenlive.js',
+	id: 'kkjz',
+}, {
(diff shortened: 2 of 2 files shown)


jaccarmac (commented):
Similar comments to the other PR. Thanks for the submission, looks good.

Since this is the first ListenLive connector, I think it would be nice to name the connector "ListenLive" or at least add a fallthrough connector which matches on all other ListenLive URLs.

Have you had a chance to test it with any other stations using that site?

inverse (approved):
LGTM - I agree with @jaccarmac perhaps a catch all for non-maped listenlive radio stations could be tested... I wonder how that would work though :thinking:

hornbuckle:
Considering the same take on #3363, for a user who only sees what's in the extension popup, it would be clearer to show the site/station they're listening to instead of the name of the player.
I briefly tried to find some other sites using this same player, but the ones I found were either talk radio or not working. I'll try to find some others and add them later. But it seems that while adding a catch-all with a wildcarded ID at the end of the URL could potentially work, it would still be a better user experience to add sites/stations separately, use the site/station names, and avoid the ones that don't actually play music.

hornbuckle:
I don't know if this is something worth exploring, but I thought I'd mention… Perhaps another option for a new feature could be to allow setting `Connector.name` in the JS file. Then there could just be one connector with a wildcarded URL for these 3rd-party streaming platforms, but the name of the site/station could be pulled from an element within the player (if provided). This wouldn't always be available, of course, and perhaps we wouldn't want to open that up to be modified outside of connectors.js. But, in this case, there does seem to be a consistent element for Listenlive (`#sidebar h1`) that could be used.
I also got the idea looking at the Audacy player. We only have one connector for Audacy, but they have a lot of music stations: https://www.audacy.com/stations/category/music It's a bit different than these recent additions, as the Audacy site design makes it much clearer that you're listening via Audacy, so not having the individual station names is perhaps not as necessary. But their player provides the station name (`#stationName`), which could also be used—example: https://www.audacy.com/947thewave

PR #3363, opened 2022-09-13
Fix connector for Jazz24

hornbuckle:
While I was adding #3362, I noticed that the connector for Jazz24 was broken.

Their main site is https://www.jazz24.org/

Streaming player is now at https://v7player.wostreaming.net/854 so updated JS connector file to wostreaming.js

Tested and confirmed scrobbling. (I also avoided any reference to "v7" in an attempt to future proof any new versions, assuming they're not a complete rebuild.)

Also found a few more stations using the same WO Streaming Player and added them to connectors.js:


jaccarmac (approved):
Looks good!

I'm not a fan of the multiple connectors entries, but I see that it's consistent and can guess it's more user-friendly.

Any other maintainers want to chime in re: many connectors vs. many URLs?

inverse:
I agree it seems weird but I think for the end user the experience is better as it'll map to the connector - def the first time this extension had to consider this.

jaccarmac:
> def the first time this extension had to consider this.

There are actually a decent number of dupes:

```js
{
  'connectors/ambientsleepingpill.js',
  'connectors/jazzandrain.js',
  'connectors/radiotunes.js',
  'connectors/yandex-music.js',
  'connectors/radioultra.js',
  'connectors/radioplayer.js',
  'connectors/musickit.js',
  'connectors/radioco.js'
}```

(That comes from looking for all duplicates, where you can see which sites are more and less common.

```js
[
  'connectors/ambientsleepingpill.js',
  'connectors/jazzandrain.js',
  'connectors/jazzandrain.js',
  'connectors/radiotunes.js',
  'connectors/radiotunes.js',
  'connectors/radiotunes.js',
  'connectors/radiotunes.js',
  'connectors/radiotunes.js',
  'connectors/yandex-music.js',
  'connectors/radioultra.js',
  'connectors/radioultra.js',
  'connectors/radioultra.js',
  'connectors/radioplayer.js',
  'connectors/radioplayer.js',
  'connectors/musickit.js',
  'connectors/radioco.js'
]```

PR #3385, opened 2022-09-26
Refactor KJazz connector for general use as Listen Live

hornbuckle:
Update to #3362. Initiated discussion at #3366 prior to submitting this but didn't receive any replies.

In short, I ended up agreeing with @jaccarmac's comments as far as making this a more general connector. I found several more stations using Listen Live, and it just made sense to refactor the connector to work automatically for them instead of manually adding each one to connectors.js.

Also added support for artwork and album text, which were not used on KJazz, for any other stations that may use them.

Here are all of the stations using Listen Live that I have found so far, and this connector should now work for all of them:
https://player.listenlive.co/67941
https://player.listenlive.co/35541
https://player.listenlive.co/42241
https://player.listenlive.co/44461
https://player.listenlive.co/50151
https://player.listenlive.co/55891
https://player.listenlive.co/63411
https://player.listenlive.co/63431
https://player.listenlive.co/64101
https://player.listenlive.co/64111
https://player.listenlive.co/64121
https://player.listenlive.co/64131

### DIFF
--- a/src/connectors/listenlive.js
+++ b/src/connectors/listenlive.js
@@ -6,4 +6,26 @@ Connector.artistSelector = '.nowPlayingCard .song .artist';
 
 Connector.trackSelector = '.nowPlayingCard .song .title';
 
+Connector.getAlbum = () => {
+	const albumText = Util.getTextFromSelectors('.nowPlayingCard .song .album');
+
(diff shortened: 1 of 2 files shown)


inverse (approved):
Nice one :+1:

PR #3386, opened 2022-09-26
Refactor Jazz24 + others for general use via WO Streaming

hornbuckle:
Update to #3363. Initiated discussion at #3366 prior to submitting this but didn't receive any replies.

In short, I ended up agreeing with @jaccarmac's comments as far as making this a more general connector. Considering the number of stations using WO Streaming and potentially others that have yet to be found, it just made sense to refactor the connector to work automatically for them instead of manually adding each one to connectors.js.

I eventually want to figure out a way to allow support for podcasts, as are available on Jazz24, but wasn't able to manage the transition between the two. Will look into that more at another time.

Here are all of the stations using WO Streaming that have been found so far, and this connector should work for all of them:
https://v7player.wostreaming.net/854

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -851,61 +851,12 @@ const connectors = [{
 	js: 'connectors/radioultra.js',
 	id: 'radiojazzfm',
 }, {
-	label: 'WXBQ 96.9',
+	label: 'WO Streaming',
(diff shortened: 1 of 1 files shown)


inverse:
Thanks!

Issue #3364, opened 2022-09-13
[Issue] Connector for Radio France (fip) is not working


MisterPharaoh:
Is there any way to get this website to scrobble? 

https://www.radiofrance.fr/fip

hans-juergen:
I checked Fip a few days ago with their live radio, and I could scrobble from there, so maybe it is fixed.

inverse:
An improved scrobbler will be available in the next release

PR #3389, opened 2022-09-26
Refactor FIP connector as Radio France

hornbuckle:
I looked into this one after seeing issue #3364. While FIP did appear to be working, I also noticed that Radio France actually includes a few other stations, and all of them can be accessed via any page of the radiofrance.fr site. So, it seemed to make more sense to refactor FIP to be a general Radio France connector instead. I also made improvements to the connector code to ensure scrobbling is handled as accurately as possible.

The music stations on Radio France are:
https://www.radiofrance.fr/fip
https://www.radiofrance.fr/francemusique
https://www.radiofrance.fr/mouv

The site has various podcasts and news stories mixed in with the music, and everything uses the same player. While I was trying to figure out how to filter any non-music audio, I noticed that the player shows a "stop" icon during music and a "pause" icon during podcasts/stories. As such, the "stop" icon is used to indicate playing, and playing non-music will not register with the connector.


src/connectors/radiofrance.js
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	return null;
+};
+
+Connector.getTrackArt = () => {
+	const trackArtUrl = Util.extractImageUrlFromSelectors('#player .cover img');
+
+	if (trackArtUrl && !trackArtUrl.includes('.svg')) {
+		return trackArtUrl.replace(/(?<=\/)\d+x\d+(?=_)/g, '400x400'); // larger image path
```
inverse, line 28: Should we have a standard unit we aim to support? e.g. I saw in your prior one we resized to 300x300 which I copied in #3382

inverse (approved):
Nice one :hand:

src/connectors/radiofrance.js
hornbuckle: For each of one I've submitted so far, I used values that were found in larger versions of the artwork located outside of the player elsewhere on the page/site, just to keep consistent with what they were already using.
I believe this one and Mixcloud were flexible with the value, so I considered using 272x272 (double the 136x136 size in the extension popup for HD screens). But some, like hoopla, have to be specific values or the image URL is no longer valid.
In answer to your question, I'm not sure if we need to set a standard or not—as long as the values are larger than 272x272 (maybe round up to 300) to keep the image crisp in the popup?

PR #3394, opened 2022-09-27
Add connector for XRAY.FM

hornbuckle:
Most of their daytime programming appears to be talk radio—some of their own and some syndicated, so it didn't make sense to include recognition of that programming in the connector.

Connector will only work when artist and track info is provided, which is simultaneously populated on https://xray.fm/playlist (to get an idea of what times of the day they play music).

Set up to work on both https://xray.fm/ and https://www.xray.fm/ because a user could visit either one and won't be redirected.


inverse:
Thanks :+1:

Issue #3311, opened 2022-07-27
amazon music connector code needs updating


neavepaul:
**Describe the bug**
Tracks played on Amazon Music are not getting scrobbled.

**How to reproduce**
Steps to reproduce the behaviour:
1. Go to 'https://music.amazon.in/'
2. Click on 'My Soundtrack or any other track'
3. See the error

**Expected behaviour**
Songs should be scrobbled and added to my last.fm account as was in the past.

ctrueden:
I am also having the problem, where Amazon Music is being scrobbled. Web Scrobbler does detect that the site is supported, but after clicking play, it does not pick up on the change.

I have looked through the DOM while music playing, and compared against `src/connectors/amazon.js`, but all the selectors and properties (for the new player case) still match up with what's on the page. I do not understand why web-scrobbler cannot detect that music is playing.

Next, I followed the instructions on the "Debug the extension" wiki page.

<details><summary>Here is the log from a fresh Amazon Music page load</summary>

```
Injecting vendor/filter.js inject.js:52:12
Injecting vendor/jquery.min.js inject.js:52:12
Injecting vendor/browser-polyfill.min.js inject.js:52:12
Injecting core/content/util.js inject.js:52:12
Injecting core/content/reactor.js inject.js:52:12
Injecting core/content/connector.js inject.js:52:12
Injecting connectors/amazon.js inject.js:52:12
Injecting core/content/starter.js inject.js:52:12
Tab 51: Created controller for Amazon connector util.js:40:10
Tab 51: New song detected: {
  "track": "Invincible",
  "artist": "TOOL",
  "album": null,
  "albumArtist": null,
  "uniqueID": null,
  "duration": null,
  "currentTime": null,
  "isPlaying": true,
  "trackArt": "https://m.media-amazon.com/images/I/71HrT-cjr+L.jpg?isCloudQueue=true",
  "isPodcast": false,
  "originUrl": "https://music.amazon.com/artists/B003X7KSU4"
} util.js:40:10
Send "get info" request: 1 scrobble-service.js:97:12
Tab 51: isPlaying state changed to false util.js:40:10
Last.fm: track.getinfo response:
{
  "track": {
    "name": "Invincible",
    "url": "https://www.last.fm/music/Tool/_/Invincible",
    "duration": "0",
    "streamable": {
      "#text": "0",
      "fulltrack": "0"
    },
    "listeners": "114660",
    "playcount": "1058567",
    "artist": {
[...]

PR #3397, opened 2022-09-30
Update Amazon connector

hornbuckle:
Using the Amazon Music connector today and noticed it was not consistently registering that anything was playing and sometimes would even register as playing when not playing at all. (Behavior similar to issue #3311.)

Issue seemed to be with the play button selector that was used, as even with the last edit that was made, the selector was no longer unique. Updated as pause button selector (with a couple of other minor edits) and appears to be consistently recognizing play state as expected now. Also updated label to "Amazon Music" to be more descriptive of the service.

FWIW unsure if separate functions for old and new players are necessary anymore. Is an "old" player even available to users?

### DIFF
--- a/src/connectors/amazon.js
+++ b/src/connectors/amazon.js
@@ -28,29 +28,27 @@ function setupPropertiesForNewPlayer() {
 
 	Connector.getTrackInfo = () => {
 		const trackContainer = document.querySelector(trackContainerSelector);
-		if (trackContainer === null) {
-			return null;
+
+		if (trackContainer) {
+			const artist = trackContainer.getAttribute('secondary-text');
+			const track = trackContainer.getAttribute('primary-text');
+			const trackArt = trackContainer.getAttribute('image-src');
+
+			return { artist, track, trackArt };
 		}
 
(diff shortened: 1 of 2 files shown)


src/connectors/amazon.js
```diff
@@ -28,29 +28,27 @@ function setupPropertiesForNewPlayer() {
 
 	Connector.getTrackInfo = () => {
 		const trackContainer = document.querySelector(trackContainerSelector);
-		if (trackContainer === null) {
```
inverse, line 31: I prefer this style of returning early as it makes the happy path less nested

inverse (changes requested):
Aside a nit about nesting I think it's good to go - thanks :+1:

src/connectors/amazon.js
hornbuckle: Sure, I can revert that. Is that the approach that should be taken going forward when updating/adding connectors?

inverse: I find it often leads to cleaner and more readable code as when have multiple conditions the nesting can get excessive or the boolean logic complex.

inverse (approved):
Thanks

PR #3408, opened 2022-10-06
Update KCRW connector

hornbuckle:
Made some updates to the KCRW connector to make it a bit more robust. Updates include:

- Player selector to support both the main player and the mini popup player.
- Ensure support for any music stream, including live and archived playlists, and not scrobble breaks.
- Add support for Today's Top Tune, which is a daily single song that is available to stream for a month: https://www.kcrw.com/music/shows/todays-top-tune
- Concise label as just "KCRW," since that's how they're known and available streams are more than just their live FM programming.

### DIFF
--- a/src/connectors/kcrw.js
+++ b/src/connectors/kcrw.js
@@ -1,9 +1,30 @@
 'use strict';
 
-Connector.playerSelector = '#player';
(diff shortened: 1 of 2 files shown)


inverse:
Thanks!

Issue #3407, opened 2022-10-05
JazzGroove


Zindar222:
JazzGroove is a jazz radio site (https://jazzgroove.org/?channel=mix1) similar to Pandora. Is it possible to scrobble tracks from this site, I use it often for work to help with focus.

inverse:
Will be in the next release thanks to @hornbuckle

PR #3409, opened 2022-10-06
Add connector for The Jazz Groove

hornbuckle:
New connector as per request #3407. https://jazzgroove.org/
Pretty straightforward player, aside from a couple of things:

- Two selectors for artist and track because the classes change when the screen width is less than 900px.
- Station bumpers and other announcements also use the artist and track locations for messaging. Initially tried to filter out specific words/phrases but wording didn't seem consistent enough for that to be reliable. Eventually decided to filter messages out by checking for default station artwork, which only appears to be used for non-music messages; all songs use actual album artwork.

### DIFF
--- a/src/connectors/jazzgroove.js
+++ b/src/connectors/jazzgroove.js
@@ -1,0 +1,17 @@
+'use strict';
+
+Connector.playerSelector = '.jg-player';
+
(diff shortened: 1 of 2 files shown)


inverse:
Thanks :+1:

Issue #3414, opened 2022-10-10
EpidemicSound scrobbling not working


Arilozen:
EpidemicSound scrobbling isn't working - using Firefox if that helps. Everything else is scrobbling normally. I double checked to make sure that in the settings I had EpidemicSound set to scrobble yet the issue remains. Maybe they changed their site again? 

**How to reproduce**
1. Go to https://www.epidemicsound.com/music/featured/
2. Click on any song.
3. Check the Web Scrobbler addon to verify it isn't registering the song.

**Debug logs**
Some cookies are misusing the recommended “SameSite“ attribute 105
INS: content-ads.js loaded:  https://www.epidemicsound.com/ [content-scripts.js:1:108429](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
TSS: content-tss.js loaded:  https://www.epidemicsound.com/ [content-scripts.js:1:118223](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
INS: content-blocked-items.js loaded:  https://www.epidemicsound.com/ [content-scripts.js:1:138261](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
CONTENT_SHELL: Page allowed. Skipping shell injection blocks [content-scripts.js:1:137923](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
GET TAB ID RESPONSE:  
Object { tabId: 30 }
[content-scripts.js:1:119759](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
TSS: excluded result:  
Object { excluded: true }
[content-scripts.js:1:128944](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
TSS: Excluding content tss (trigger: send-mesage) [content-scripts.js:1:120296](moz-extension://11ff5be3-6505-4618-8315-a413f4195a42/content-scripts.js)
InstallTrigger is deprecated and will be removed in the future. [mainapp.bundle.js:2:324129](https://cdn.epidemicsound.com/player/20221010.87-0866fff409388670cc9a6db35aaa569de02a5f26/mainapp.bundle.js)
[...]

PR #3417, opened 2022-10-12
Fix Epidemic Sound connector

hornbuckle:
Fix for #3414.
I don't have an account with this site but tested using the URL provided: https://www.epidemicsound.com/music/featured/
Two selectors provided for trackSelector because its location changes when the screen width is less than 960px wide.

Unfortunately, the previous update to this connector did not use good selectors. Oddly, it also used attribute selectors, which were originally used to wildcard the classes, but still included the entire class with random strings.
When classes include random strings like that (seen most often on sites built with React), those characters will change whenever an update is pushed to the site—even if the actual markup remains the same. If a connector is submitted with random strings in the selectors, it's probably best to request changes to remove them before merging in order to prevent the connector from breaking too often.

### DIFF
--- a/src/connectors/epidemicsound.js
+++ b/src/connectors/epidemicsound.js
@@ -1,11 +1,13 @@
 'use strict';
 
-Connector.playerSelector = '[class="src-mainapp-player-components-___PlayerBar__player___ZJ8j7"]';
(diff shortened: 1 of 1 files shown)


inverse (approved):
:crossed_fingers: looks a lot less brittle

PR #3422, opened 2022-10-13
Fix interaction with sponsorblock completely

### LINKED ISSUE #3421: malfunction with sponsorblock
when im listening to songs (i.e. if yyou had my love by j-lo) which has auto-promotion, it scrobbles like this:
image

and if i try to rename it for the correct name, every single time i'm listening to other songs it will scrobble as the one i corrected

yayuyokitano:
Was unable to fully fix this previously, but thanks to change pushed to sponsorblock this can now be done.
Note that this requires the next version after 5.0.7 of sponsorblock to work, which is not uploaded to extension stores yet(!)
Merging this has to wait until the next version is in extension stores.

Fixes #3421


inverse (approved):
LGTM :+1: Should we hold off on this or do you think we should merge this or hold off until the release?

yayuyokitano:
I say probably wait – the existing solution at least _sometimes_ works with 5.0.7

inverse:
Looks like we need to wait for this release to be published :crossed_fingers: 

https://github.com/ajayyy/SponsorBlock/releases/tag/5.1.0

ajayyy:
That version has been published

inverse:
@ajayyy isn't it only a pre-release though? so not generally available unless you subscribe to the beta version?

ajayyy:
Sorry, it's written a little weirdly, but since 5.1.1 is "stable", it was published. That contains changes from both versions

inverse:
Thanks for clarifying!

PR #3424, opened 2022-10-15
QCIndie Connector

jbwharris:
QCIndie.com connector initial push

New connector for qcindie.com

**Additional context**
image

### DIFF
--- a/src/connectors/qcindie.js
+++ b/src/connectors/qcindie.js
@@ -1,7 +1,7 @@
 'use strict';
 // the artist and track info are inverted on this site, values for artist and track are intentionally flipped
 
-Connector.playerSelector = '#content.site-content';
-Connector.trackSelector = '#ReloadThis .col-sm-6:first-child p.artistinfo';
-Connector.artistSelector = '#ReloadThis .col-sm-6:first-child p.trackinfo';
-Connector.isPlaying = () => $('#wp1598381648smallbuttonplay').css('display') === 'none';
+Connector.playerSelector = '#content';
+Connector.artistSelector = '#ReloadThis > div:first-of-type .trackinfo';
+Connector.trackSelector = '#ReloadThis > div:first-of-type .artistinfo';
+Connector.isPlaying = () => Util.isElementVisible('.lunaaudioplayer [id$=smallbuttonpause]');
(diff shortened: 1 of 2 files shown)


hornbuckle:
The player on this site is a plugin: https://radioplayer.luna-universe.com/
The version being used on this doesn't provide any artist and track info while playing.
Additionally, those selectors do not look reliable.

hornbuckle:
Final review and approval is up to @inverse, of course, but if this site was going to be approved for inclusion, my unsolicited suggestions for better selectors and removal of jQuery would be as follows:

`Connector.playerSelector = '#content';`

`Connector.artistSelector = '#ReloadThis > div:first-of-type .trackinfo';`

`Connector.trackSelector = '#ReloadThis > div:first-of-type .artistinfo';`

`Connector.isPlaying = () => Util.isElementVisible('.lunaaudioplayer [id$=smallbuttonpause]');`

Haven't tested though, so please check if these provide accurate results.

jbwharris:
@hornbuckle I have actually used the connector over a long period of time and it is reliable. The logic being used here is that it's taking the first child of the Just Played column when it changes. It's not ideal, but it does work well. I did try using the song / artist output in the header, but it required parsing and splitting the output.

jbwharris:
@hornbuckle I'll give those a test right now

jbwharris:
@hornbuckle I just pushed those revised selectors over.

inverse (approved):
Thanks for the contribution + @hornbuckle for the review too :+1:

hornbuckle:
@jbwharris The "Just Played" heading is what threw me off initially, but I realized after listening for a few songs that the first div actually did contain the current song.

However, I missed the "Now Playing" at the top when reviewing earlier, which I see you mentioned in your comment after my suggested selectors.

There is a utility method to handle splitting the artist and track from a single selector, so you generally don't have to write that logic yourself. But there is also an an alternate option here, which may be less fragile in the long run, to replace `artistSelector` and `trackSelector` with the following, as it also handles the split:

`Connector.artistTrackSelector = '#stream_info_song';`

That may also require updating `playerSelector` to something higher on the page (such as `#page`) so that "Now Playing" included in the connector observer.

PR #3426, opened 2022-10-15
Update The Jazz Groove connector

hornbuckle:
Minor edit for clearer logic in connector for https://jazzgroove.org/

- Moved `isScrobblingAllowed` to `isTrackArtDefault`, since that is what it's actually checking, and then used `!isTrackArtDefault` for `isScrobblingAllowed`.

No rush to merge, as end behavior is the same as current version.


inverse:
:pray: thanks once again!

PR #3427, opened 2022-10-15
Update KCRW connector

hornbuckle:
Minor edits to connector for https://www.kcrw.com/

- Clearer setup of `MetadataFilter` with separate function
- Added `currentTime` and `duration` for Today's Top Tune so it doesn't scrobble too soon after it begins playing

No rush to merge, as end behavior is essentially the same as current version.

### DIFF
--- a/src/connectors/kcrw.js
+++ b/src/connectors/kcrw.js
@@ -16,8 +16,8 @@ Connector.getTrackInfo = () => {
 			artist: artistText,
 			track: trackText,
 		};
-	} 
-	
+	}
+
 	if (!Util.hasElementClass(metaSelector, 'music') && artistText && showNameText === 'Today\'s Top Tune') {
 		const artistTrackSplit = artistText.split(': ');
 		return {
--- a/src/connectors/kcrw.js
+++ b/src/connectors/kcrw.js
@@ -1,13 +1,12 @@
 'use strict';
 
-Connector.playerSelector = '.site-player'; // supports main and mini players
+const filter = MetadataFilter.createFilter({ track: removeFancyQuotes });
 
-const filter = MetadataFilter.createFilter({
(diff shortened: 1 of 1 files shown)


src/connectors/kcrw.js
```diff
@@ -17,14 +16,25 @@ Connector.getTrackInfo = () => {
...
 	return null;
 };
 
+Connector.isPlaying = () => Util.hasElementClass('button.play', 'active');
+
 Connector.applyFilter(filter);
 
-Connector.isPlaying = () => Util.hasElementClass('button.play', 'active');
+function removeFancyQuotes(text) {
+	return text.replace(/[‘’“”]/g, ''); // filter for Today's Top Tune
```
inverse, line 39: Wonder how useful this could be to moved into https://github.com/web-scrobbler/metadata-filter at somepoint? 

That lib needs some <3

inverse (approved):

inverse:
:pray: thanks once again!

src/connectors/kcrw.js
hornbuckle: Sure, I haven't looked into adding anything to that yet, but if you think it would be useful, I'll take a look.

jbwharris:
@hornbuckle I just noticed a bug with this one. In instances where a song has an apostrophe it's being stripped out.  

I can see the necessity for the Today's Top Tune formatting, but I suppose if that song were to have an apostrophe in the title it would encounter the same problem. 
image

But it's applying to regular tracks like this one as well which was played through the Eclectic 24 stream. 
image

hornbuckle:
@jbwharris Thanks for catching that. When I was testing out the connector, I was only encountering titles with a standard apostrophe, which is how they should be typed. I can probably update the filter to work better for Today's Top Tune, as it's intended. But then I wonder if it also needs to replace fancy apostrophes with standard ones when found within a title. I'll take a look again this week and work on a better filter.

PR #3428, opened 2022-10-15
Update anime(bits) connector

hornbuckle:
Updates to connector for https://radio.animebits.moe/

After submitting this connector, I realized there are also separate versions of the player available to a user for listening.

- The player is essentially the same, but a few classes are slightly different, so two selectors are now available for artist, track, and album.
- Also added a new filter to remove the playlist track number that gets prepended in the album player.
- Added check for default art.

Additional example URLs for testing:
https://radio.animebits.moe/player/hash/2d2258e32
https://radio.animebits.moe/player/album/hide+your+face
https://radio.animebits.moe/player/search/bump+of+chicken

No rush to merge, as main player at base URL is currently working.

### DIFF
--- a/src/connectors/animebits.js
+++ b/src/connectors/animebits.js
@@ -1,12 +1,14 @@
 'use strict';
 
+const filter = MetadataFilter.createFilter({ track: removePlaylistNumber });
+
 Connector.playerSelector = '#radio-left';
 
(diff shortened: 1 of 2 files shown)


inverse (approved):
:pray: thanks once again!

Issue #3344, opened 2022-08-30
Filtering station IDs so they don't get scrobbled between songs?


jbwharris:
I have been working on a connector for [Indie 102.3](https://indie.cpr.org/) and it does work, the problem is between songs they'll throw a tag in there for the station ID and the current DJ. Now I've been recently using the [WYEP](https://wyep.org/) connector which I see having a similar behaviour, but it doesn't seem to have this problem. And when I look at the code for this connector, there doesn't seem to be anything implicit that filters the output. Is there something within the Web Scrobbler that handles filtering, or am I just missing something here?

**Indie 102.3 page output**
![image](https://user-images.githubusercontent.com/2962327/187498430-ec5e4569-2a48-463f-9393-49ea00364892.png)
**Scrobbled station ID I just deleted**
![image](https://user-images.githubusercontent.com/2962327/187498756-67b264a8-57a8-43e8-bf8f-502517ec02c0.png)

**WYEP station ID output that doesn't get scrobbled**
![image](https://user-images.githubusercontent.com/2962327/187504951-4fd15dce-b313-47c4-a62f-aa4e1a0e9230.png)

inverse:
I don't know why that scrobbler would be working... perhaps it's due to the way it parses out the artist on `.` and if that doesnt match then nothing scrobbles.

For your case you could try something like...

```
Connector.isScrobblingAllowed = () => !document.querySelector('.qtmplayer__artist').textContent.startsWith('With ');
```

To block playback when artist matches `With `

jbwharris:
Thanks for pointing me in the right direction. I need to test it a bit more and see if there might be more conditions that I might need to hedge against, but it seems to be working so far.

PR #3429, opened 2022-10-16
Add connector for Colorado Public Radio

hornbuckle:
URL: https://www.cpr.org/
Listen Live links for both CPR Classical and Indie 102.3 provide artist and track info for scrobbling.

`isStateChangeAllowed` added to ensure scrobbling isn't temporarily reset to null in the middle of a song, as the track selector is always populated when a state change should occur.

### DIFF
--- a/src/connectors/cpr.js
+++ b/src/connectors/cpr.js
@@ -1,0 +1,11 @@
+'use strict';
+
+Connector.playerSelector = '[class*=audioPlayer__liveStream-]';
+
+Connector.artistSelector = `${Connector.playerSelector} [class^=audioPlayerTitle__artist-]`;
+
+Connector.trackSelector = `${Connector.playerSelector} [class^=liveStreamTitle__heading-]`;
+
+Connector.pauseButtonSelector = `${Connector.playerSelector} [class^=playPauseButton__icon-][class*=playPauseButton__pause-]`;
+
+Connector.isStateChangeAllowed = () => Connector.getTrack();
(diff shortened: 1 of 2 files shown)


inverse (approved):
:pray: thanks once again!

jbwharris:
@hornbuckle I had a crack at trying to get https://indie.cpr.org/ to work awhile back and it was a bit problematic with the station and host IDs. I'm not sure if that site can also be built into the scope of this connector. This is what I had for that page, but I found it a little unreliable for scrobbling. Here's what I had if you wanted to take a look. There was an issue I opened for this as well #3344 

```

'use strict';

Connector.isPlaying = () => $('i.qtmplayer__notification').is($("i.qtmplayer__notification.playing"));

Connector.playerSelector = 'div.qtmplayer__basic';
Connector.artistSelector = 'p.qtmplayer__artist';
Connector.trackSelector = 'p.qtmplayer__title';

Connector.isScrobblingAllowed = () => {
	const track = Connector.getTrack();
	return track && !track.startsWith('Indie 102.3');
};
```

hornbuckle:
@jbwharris Ah, yes, I remember seeing that issue. The main CPR site is a bit easier to manage, since the show info only shows up in the track selector and not both track and artist. However, I believe you were more or less on the right track (just be sure to use [Util methods](https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/content/util.js) instead of jQuery when you're working on connectors).
If you don't mind, I can work on a solution and spend a day or two with it to make sure all show info variations are included and submit a new connector. I believe a separate one would make sense in this case, since it's a different player design and a separate subdomain specifically for this station.

PR #3431, opened 2022-10-17
Update CPR connector

hornbuckle:
Follow-up to #3429.
URL: https://www.cpr.org/

Before this is released, I realized there was a bit more to be added—support for album and check for artist to scrobble.


inverse:
Thanks :pray:

PR #3433, opened 2022-10-18
Add connector for Indie 102.3

### LINKED ISSUE #3344: Filtering station IDs so they don't get scrobbled between songs?

hornbuckle:
New connector. Resolves #3344 (H/T @jbwharris)
URL: https://indie.cpr.org/

After spending some time with this site, it seems that "Indie 102.3" can appear in either the artist or track location, and that is when scrobbling should be prevented. Additional check for artist, which sometimes says "with DJ Name," in case the accompanying show does not include "Indie 102.3."
Adding as a separate connector from CPR, even though this station feed can also be accessed at https://www.cpr.org/indie/ , since it is a different player design and a separate subdomain.

### DIFF
--- a/src/connectors/cpr-indie.js
+++ b/src/connectors/cpr-indie.js
@@ -1,0 +1,18 @@
+'use strict';
+
+Connector.playerSelector = '#qtmplayer';
+
+Connector.artistSelector = '.qtmplayer__songdata .qtmplayer__artist';
+
+Connector.trackSelector = '.qtmplayer__songdata .qtmplayer__title';
+
+Connector.isPlaying = () => Util.hasElementClass('#qtmplayerNotif', 'playing');
+
+Connector.isScrobblingAllowed = () => {
+	const stationID = 'Indie 102.3';
+	return (
+		!Connector.getTrack().includes(stationID) &&
+		!Connector.getArtist().includes(stationID) &&
+		!Connector.getArtist().startsWith('with ')
+	);
+};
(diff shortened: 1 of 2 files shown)


jaccarmac (approved):

jaccarmac:
Code looks good, @jbwharris can you confirm this is functioning as expected?

jbwharris:
I believe is still has some of the same issues I had experienced before.

image

In this screenshot you can see that it missed 2 songs that played. It was working fine when I was sitting at my computer, but when it was just left for 10 minutes it just seemed to stop looking for changes. Also the song after this screenshot, it froze and wouldn't update, requiring a refresh. I find sometimes it'll just be snoozing in updating the page, so it'll be stuck on the last song, then you click and scroll and the UI will update.

hornbuckle:
Hey @jbwharris, I actually noticed this happening when I was working on it. Unfortunately, the problem is with the site not always refreshing the song in the player and not the extension itself. Unfortunately, there isn't much we can do on the extension side if the player doesn't provide the updated song info. Refreshing the page seems to get it going again, but eventually it still seems to stop refreshing the player again, too.
If you want to scrobble from that station for an extended period of time, the feed at https://www.cpr.org/indie/ seems to be more reliable.

PR #3438, opened 2022-10-24
Add connector for DKFM

hornbuckle:
URL: https://decayfm.com/

Coincidentally, this site uses the same player as the recently-added Indie 102.3 standalone site. However, it seems to be a slightly different version and requires a different check for both `isPlaying` and `isScrobblingAllowed`, so they can't share a connector script. This site doesn't seem to have an issue with refreshing the player information, as I have listened for an extended period of time with no issue of keeping up with the tracks played.

Adding label as "DKFM Shoegaze Radio," even though it's a bit long, as that is how they are listed everywhere.

DKFM also has their up-to-date play history on Last.fm: https://www.last.fm/user/DKFM_Shoegaze/library

### DIFF
--- a/src/connectors/decayfm.js
+++ b/src/connectors/decayfm.js
@@ -1,0 +1,15 @@
+'use strict';
+
+Connector.playerSelector = '#qtmplayer';
+
+Connector.artistSelector = '.qtmplayer__songdata .qtmplayer__artist .marquee';
+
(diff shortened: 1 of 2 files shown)


inverse:
Thanks once again :pray:

inverse:
@hornbuckle you on discord?

PR #3439, opened 2022-10-24
Add connector for Streemlion

hornbuckle:
This is a 3rd party player that I discovered while working on the connector for DKFM. Aside from their main stream, they have two additional streams available through Streemlion:

- DKFM Classic - https://listen.streemlion.com/radio_5fea66f604ac9
- DKFM EDGE - https://listen.streemlion.com/radio_5fea66f604ac2

I also found another stream using this player on nomadradio.eu: https://listen.streemlion.com/NomadRadio
That one uses a different player layout, but both versions work with the connector.

### DIFF
--- a/src/connectors/streemlion.js
+++ b/src/connectors/streemlion.js
@@ -1,0 +1,11 @@
+'use strict';
+
+Connector.playerSelector = '.player-wpr';
+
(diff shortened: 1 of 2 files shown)


inverse (approved):
Thanks once again :pray:  You're on a roll!

PR #3441, opened 2022-10-24
WXPN Connector

jbwharris:
New connector for xpn.org. Works for XPN and XPN2

**Additional context**
image

### DIFF
--- a/src/connectors/xpn.js
+++ b/src/connectors/xpn.js
@@ -1,0 +1,7 @@
+'use strict';
+
+Connector.playerSelector = '.sgplayer';
+Connector.trackSelector = '.sg-content-info .sg-title';
+Connector.artistSelector = '.sg-content-info .sg-artist';
+Connector.trackArtSelector = '#art';
+Connector.isPlaying = () => Util.getAttrFromSelectors('.sg-control-play a', 'title') === 'Pause';
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2217,12 +2217,12 @@ const connectors = [{
 	js: 'connectors/cpr-indie.js',
 	id: 'cpr-indie',
 }, {
-	label: 'WXPN',
+	label: 'XPN',
 	matches: [
 		'*://xpn.org/*',
 	],
 	js: 'connectors/xpn.js',
-	id: 'wxpn',
+	id: 'xpn',
 }];
 
 define(() => connectors);
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2215,7 +2215,7 @@ const connectors = [{
 		'*://xpn.org/*',
 	],
 	js: 'connectors/xpn.js',
-	id: 'WXPN',
+	id: 'wxpn',
 }];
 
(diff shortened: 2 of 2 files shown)


hornbuckle:
@jbwharris Just wanted to note that the `id` in connectors.js should be (lowercase) `xpn` and not (uppercase) `WXPN`.
Wiki: https://github.com/web-scrobbler/web-scrobbler/wiki/Connectors-development#how-to-assign-id-for-a-website

jbwharris:
@hornbuckle Good to know, just pushed that update.

hornbuckle:
@jbwharris Sorry to nitpick, but for consistency with existing connectors and per the wiki, I believe it should be `xpn` (to match the domain xpn.org and connector filename xpn.js). Only mentioning since it's not supposed to be changed once it's released, so just want to help ensure it's correct ahead of time.

jbwharris:
@hornbuckle It is inconsistent because the station name seems inconsistent. I think their official name is WXPN, but then their station branding they go by XPN. The DJs and ads will say XPN, but then the mentions and graphics on their site says WXPN. So I was trying to cover both names in the connector if that makes any sense. Which do you feel I should name it?

hornbuckle:
@jbwharris Sure, I understand it's inconsistent with this case. However, `WXPN` as the connector `label`, which you used, appears to be correct.

But the label and id will not necessarily match. For consistency with existing connectors and going by what the extension wiki says, there are two points to consider for the `id`:

1. [Link](https://github.com/web-scrobbler/web-scrobbler/wiki/Connectors-development#how-to-name-connector-file) "The connector filename should match the second-level (or lower-level) domain of the service, e.g. for youtube.com correct name is youtube.js." (So, xpn.js, which you used, is correct.)
2. [Link](https://github.com/web-scrobbler/web-scrobbler/wiki/Connectors-development#how-to-assign-id-for-a-website) "…the id property equals both the connector filename and the second-level domain of the website. … YouTube: connector filename is youtube.js and connector id is youtube (the connector filename and the second-level domain are equal)." (So, `xpn` as `id` would be correct.)

I know it's only a difference of one letter in this particular case, but just trying to clarify for this and future submissions. 🙂

hornbuckle:
@jbwharris This one may be a confusing one, I guess. 😕
`WXPN` as the label was previously correct.

jbwharris:
@hornbuckle So should I be changing everything to WXPN? I thought that it should be reflective of the domain, so I was changing it all to be the same.

hornbuckle:
@jbwharris Apologies for any confusion. I was trying to explain it as clearly as I could, with quotes from the wiki, but I guess I failed.

Here is what I was trying to get to:
```
	label: 'WXPN',
	matches: [
		'*://xpn.org/*',
	],
	js: 'connectors/xpn.js',
	id: 'xpn',
```
Note the domain, connector filename, and id are all the same. However, the label is different because that is how they actually type it on the site: https://xpn.org/about-us/

inverse (approved):
Thanks for the contribution!

Issue #3330, opened 2022-08-15
Frisky Radio


pedrojvaz:
Hello,

Could you support https://www.friskyradio.com (soon to be https://frisky.fm) online radio?
Current song/show is available at the bottom of the page.

Thanks

inverse:
Not sure how feasible that is - are tracks displayed? I couldn't see anything

pedrojvaz:
At the bottom it shows the current show being played, but not the track. Would this work?

jbwharris:
@pedrojvaz @inverse I was able to get a connector working for frisky.fm. The main problem I ran into was that their play/pause button was switched using xlink:href. I wasn't able to get it to pick up the play behaviour off this button, but I did get it to work based on whether the song/artist meta box existed. So if the song were paused or not, it wouldn't actually know since it isn't tied to that button. There's also a pretty big blind spot in not having a paid account, so I don't whether the behaviour might differ. It's somewhat difficult to test, since so many of the songs are intentionally long, and the free tier doesn't really have skip functionality.

inverse:
@jbwharris like it was an svg path? We have connectors that already use that if so

jbwharris:
@inverse I did try searching the codebase for something to use with the deprecated `<use xlink:href="#icon-play_arrow"></use>` in an SVG, but if there's connectors you could point me to I'll see if I can get it working.

inverse:
Will be in the next release!

PR #3444, opened 2022-10-27
frisky.fm connector

jbwharris:
New connector for frisky.fm

Adding new connector as requested in #3330. If you're using the site with only a free account you will only be able to test using the 4 radio streams. If someone with a frisky.fm account could test this more thoroughly it would be appreciated. I'm not sure if @pedrojvaz has a paid account to test.

**Additional context**
image

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -820,6 +820,7 @@ const connectors = [{
 	label: 'anime(bits)',
 	matches: [
 		'*://radio.animebits.moe/',
+		'*://radio.animebits.moe/player/*',
 	],
 	js: 'connectors/animebits.js',
 	id: 'animebits',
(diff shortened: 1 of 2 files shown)


hornbuckle:
@jbwharris Just wanted to point you in the direction of this recently updated connector for how to write selectors to leave off those last 5 random characters on the classes using attribute selectors: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/epidemicsound.js

jbwharris:
@hornbuckle Great suggestion, will do that moving forward. I just updated the connector to use a wildcard selector.

hornbuckle:
BTW I'm not familiar with this site, but it seems as if the label should be "FRISKY" since that is how it is consistently typed. Maybe the original requester can confirm.

jbwharris:
@pedrojvaz  that is the only site that's currently supported. I can look at the old site, but I figured if they're moving towards a new design and URL that I'd just create a connector for the new site since it won't have an expiry date.

src/core/connectors.js
```diff
@@ -2237,6 +2237,14 @@ const connectors = [{
 	],
 	js: 'connectors/xpn.js',
 	id: 'xpn',
+}, {
+	label: 'frisky.fm',
```
inverse, line 2241:
```suggestion
	label: 'FRISKY',
```

inverse (approved):

PR #3455, opened 2022-10-31
Update filters in KCRW connector

hornbuckle:
More robust filters for KCRW to fix the issue noted in [this comment](https://github.com/web-scrobbler/web-scrobbler/pull/3427#issuecomment-1296280680). (H/T @jbwharris)

In addition to ensuring the outer quotes are removed for Today's Top Tune, a filter has been added to replace "smart" quotes with plain quotes within any titles and artists.
Also added existing `removeVersion` and `removeCleanExplicit` filters.

P.S. @inverse Let me know if you still think either of these new filters are worth adding to the MetadataFilter module prior to merging this.
I could see the replacement one potentially being useful outside of this connector but not sure about the removal one.


inverse:
We could also PR that after and clean things up - That filter package needs some <3

PR #3459, opened 2022-11-05
Fix apple music connector

### LINKED ISSUE #3277: Add support for https://beta.music.apple.com.
It's the Apple Music Web Player's beta website.

### LINKED ISSUE #3442: Scrobbling not working on apple music website
**Describe the bug**
Scrobbling is no longer working on https://music.apple.com. This is what the extension says when playing music on the site: 
Screenshot 2022-10-24 154624

-   OS: Windows
-   Browser: Chrome
-   Extension version: 2.77.0

yayuyokitano:
Apple music pushed their beta to prod, this makes web scrobbler work with it again.
I don't have premium account, but this works with free at least.
A lot of shenanigans because of shadow root stuff, might be a little unstable.

Probably a good idea to publish on merge, as apple music kinda big.

Fixes #3442 
Fixes #3277


jaccarmac (approved):
LGTM.

When I read the issue comments, I thought there was a way to query into the shadow DOM. Thanks for putting this together.

Someone else should pull the publish trigger, as I have not tried it before.

yayuyokitano:
I released using the old system before, but didnt use the GH action yet.
Pretty sure it's just running the release extension workflow with no other caveats.
That said I'm not doing it without confirmation from inverse + knowing hes immediately available to fix any mess that may occur 😅

src/connectors/apple-music.js
```diff
@@ -1,22 +1,29 @@
...
-Connector.getTrackInfo = () => {
-	const artistAlbum = Util.getTextFromSelectors(artistAlbumSelector);
-	return Util.splitArtistAlbum(artistAlbum, [ARTISTALBUM_SEPARATOR]);
-};
+Connector.getTrack = () => queryShadowRoot('.lcd-meta__primary-wrapper .lcd-meta-line__fragment').textContent;
 
-Connector.isPlaying = () => {
-	return $(Connector.playerSelector).hasClass('is-playing');
-};
+Connector.isPlaying = () => document.querySelector('.playback-play__pause').getAttribute('aria-hidden') !== 'true';
```
hornbuckle, line 17: Would the following work here, to prevent an error when the page first loads?
`Util.getAttrFromSelectors('.playback-play__pause', 'aria-hidden') !== 'true';`

yayuyokitano: fixed, nice catch!

hornbuckle: Nice work on this. I had similar ideas to get it working and attempted updating the connector yesterday but quickly ran out of patience before I could get anything usable to submit. 😅
BTW I wonder if each occurrence of `textContent` needs additional checks, as the following error also appears on page load:
`Uncaught TypeError: Cannot read properties of null (reading 'textContent')`
But everything seems to work once playback is initiated.

yayuyokitano: yep, it indeed doesnt really affect anything, it just throws a little error when the element doesn't exist (which is when you dont want to scrobble anyway). Thanks to javascript being javascript this doesn't really affect anything else.
Regardless, I just pushed another commit with some optional chaining that eliminates the errors, changes no behavior afaik.

hornbuckle: Nice concise solution! 😎

inverse:
@yayuyokitano go for it :) with the new action you just need to select the release type as you would on cli.

should be fine :)

inverse (approved):

ericcirone:
Sees to be working now. Thank you!

drawcode:
Slick work with the shadowroot. I thought it would have to be the js solution (https://github.com/web-scrobbler/web-scrobbler/issues/3442#issuecomment-1291756221) and outside the connector/selectors. 

Thanks for fixing it, last.fm and libre.fm are probably wondering why no one scrobbled from apple for two weeks. Even if Apple changes things at least there is a way to get to quicker fix.

UnbreakCode:
> I thought it would have to be the js solution ([#3442 (comment)](https://github.com/web-scrobbler/web-scrobbler/issues/3442#issuecomment-1291756221)) and outside the connector/selectors.

it should be this solution because it's working even after changing their site

drawcode:
I did find a few holes in the shadowroot solution. It seems the shadowroot dom only updates when the page is active after a bit. It appears to not update the info to the display when the tab is not active or minimized.  So it works well when the window is focused, but can drop out if it is not. This must be an optimization that is just not refreshing those elements if not active so there is nothing to scrobble according to those elements.

PR #3471, opened 2022-11-12
Update Freegal Music connector

hornbuckle:
Initially updated to set up the original filter (see #3335 for more info) more clearly. Also realized it could use a couple of additional filters, whose concepts were borrowed from `metadata-filter` modules. Unfortunately, because of the possibly truncated song titles, I had to make modified versions of them just for this connector.
Also updated the connector label to Freegal Music, as that is the site's full name: https://www.freegalmusic.com/home

No rush to release this, as its functionality is essentially the same.

### DIFF
--- a/src/connectors/freegalmusic.js
+++ b/src/connectors/freegalmusic.js
@@ -1,35 +1,50 @@
 'use strict';
 
+const filter = MetadataFilter.createFilter({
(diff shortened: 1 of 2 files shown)


inverse (approved):
Thanks for the contribution!

PR #3472, opened 2022-11-12
Update Burntable connector

hornbuckle:
This is one of the first new connectors I submitted that required some additional logic. Well, I realized it could be done a bit better, as I was trying to cram everything into `getTrackInfo` without realizing I could use `getTrack` separately. Also updated for clearer filter setup.
URL: https://burntable.com/listen

No rush to release this, as its functionality is essentially the same.


inverse (approved):
Thanks for the contribution!

Issue #3260, opened 2022-05-16
Add. super45.fm


akrasiko:
Hi, 

I really enjoy Web Scrobbler and I would like to ask if the site super45.fm can be added to Web Srobbler. It's a cool online radio from Chile.

Thanks!

hornbuckle:
Looks like this site just uses a Radio.co embed, which is already supported. Click the "Pop out player" icon in the top right corner of the player, and it will scrobble with the extension.

akrasiko:
Hi David!

I hope this email finds you well. I tried what you said and Firefox
recognize the site and the song but it doesn't get scrobbled. When I click
on the Web Srobbler logo I got this message:

It looks like something went wrong while scrobbling the song. Open the
options page and try re-authenticating with the services. If
re-authentication doesn't help, and you still see this message, create a
new issue on the project page on GitHub.

Thanks for all your help and the amazing extension that Web Scrobbler is.
Warm regards!

hornbuckle:
I believe if you received that error, you wouldn't be able to scrobble from any site.
Were you able to re-authenticate your account(s) and try again?

akrasiko:
Now it's working perfectly.

Thanks!

PR #3474, opened 2022-11-13
WBRU connector

jbwharris:
New connector for WBRU.com. Functions for both the Indie and 360 Hip Hop feeds on the website. 

**Additional context**
image

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2293,6 +2293,15 @@ const connectors = [{
 	],
 	js: 'connectors/wsum.js',
 	id: 'wsum',
+}, {
+	label: 'WBRU',
+	matches: [
+		'*://*wbru.com/*',
+		'*://embed.radio.co/player/*',
+	],
+	js: 'connectors/wbru.js',
+	id: 'wbru',
+	allFrames: true,
 }];
 
 define(() => connectors);
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2299,7 +2299,7 @@ const connectors = [{
 		'*://*wbru.com/*',
 		'*://embed.radio.co/player/*',
 	],
-	js: 'connectors/wbru.js',
+	js: 'connectors/radioco.js',
 	id: 'wbru',
 	allFrames: true,
 }];
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2296,8 +2296,7 @@ const connectors = [{
 }, {
 	label: 'WBRU',
 	matches: [
-		'*://*wbru.com/*',
-		'*://embed.radio.co/player/*',
+		'*://www.wbru.com/*',
 	],
 	js: 'connectors/radioco.js',
 	id: 'wbru',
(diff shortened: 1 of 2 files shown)


hornbuckle:
There already is a radio.co connector: https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/radioco.js
So, instead of adding a separate JS file specifically for this site, support for WBRU could possibly be added in the same way it was added for 9128.live

jbwharris:
@hornbuckle That makes sense to me. So should I just remove the new connector PR and just add the entry into connectors.js?

hornbuckle:
I believe you could just keep this same PR open, delete wbru.js from your branch, and push a new update connectors.js

hornbuckle:
Actually, if you wanted to, you could also add super45.fm to connectors.js in the same way and resolve #3260. When I initially responded there, I didn't realize that radio.co support had first been added to support 9128.live and thought it was simply to support any embeds.
Not sure if @inverse would prefer separate PRs or not, but since this would only contain updates to connectors.js with both sites using radio.co, it seems like it would be appropriate to use this same PR since it is already open.

src/core/connectors.js
```diff
@@ -2299,7 +2299,7 @@ const connectors = [{
 		'*://*wbru.com/*',
```
hornbuckle, line 2299: This can just be `'*://www.wbru.com/*',`

src/core/connectors.js
```diff
@@ -2299,7 +2299,7 @@ const connectors = [{
...
 		'*://embed.radio.co/player/*',
```
hornbuckle, line 2300: This is no longer needed.

jbwharris: @hornbuckle would this work still? I'm not 100% on how `allFrames: true` works, but wouldn't that need to be added to the radio.co connector entry?

hornbuckle: As far as I am aware, `allFrames` injects the connector script into all iframes on the page, so it doesn't need a specific URL pattern to work.

jbwharris: @hornbuckle Just testing my own dev build, if I remove the WBRU entry in connectors.js it doesn't work. If I change the Radio.co entry to add WBRU as a URL and add the allFrame line it works. 

```
	label: 'Radio.co',
	matches: [
		'*://embed.radio.co/player/*',
		'*://*wbru.com/*',
	],
	js: 'connectors/radioco.js',
	id: 'radioco',
	allFrames: true,
```

Is this the approach we should go with, stacking the radio.co URLs in that entry instead of having individual connector entries? Or is it more valid to have individual entries per website that link to radioco.js? Is there analytics that track what connectors are being used, so it might be worthwhile identifying which site is doing the scrobbling?

hornbuckle: Oh, so, you don't need to make any changes to the existing Radio.co entry. That should stay as-is.

All you need to do is add the new WBRU entry to connectors.js, as follows, and you're done:
```
	label: 'WBRU',
	matches: [
		'*://www.wbru.com/*',
	],
	js: 'connectors/radioco.js',
	id: 'wbru',
	allFrames: true,
```

jbwharris: @hornbuckle Ok, that makes sense. I thought you were saying to get rid of it entirely, my bad.

hornbuckle: Ah, sorry, I see it was a bit confusing because my second comment included both lines when I only meant for it to refer to the line directly above it.

inverse (approved):
Thanks for the contribution!

PR #3475, opened 2022-11-13
Update QCIndie Connector

jbwharris:
QCIndie updated their web player to use a newer version of YesStreaming, so I cleaned this connector up a bit since it was a little rough previously with some of the selector logic. 

**Additional context**
image

### DIFF
--- a/src/connectors/qcindie.js
+++ b/src/connectors/qcindie.js
@@ -3,4 +3,4 @@
 Connector.playerSelector = '#content';
 Connector.artistSelector = '.p3-artistInfo';
 Connector.trackSelector = '.p3-title';
-Connector.isPlaying = () => { return Util.hasElementClass('#ppBtn', 'playing'); };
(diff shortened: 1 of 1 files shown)


hornbuckle:
Since this is still open, just wanted to mention that since the player is now self-contained, looks like `playerSelector` can be changed to `.p3-wrapper` so the extension doesn't need to observe the entire content wrapper anymore.

src/connectors/qcindie.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
-// the artist and track info are inverted on this site, values for artist and track are intentionally flipped
 
 Connector.playerSelector = '#content';
```
inverse, line 3:
```suggestion
Connector.playerSelector = '.p3-wrapper';
```

inverse (approved):
Thanks for the contribution and @hornbuckle for the feedback

PR #3488, opened 2022-11-24
Extend invidious to more domains

heyarne:
I extended the invidious connector to match on more domains. 

**Additional context**
The reason for this is that invidious is built with self-hosting in mind, so matching only on *.invidious.com will miss a lot of the instances. I took the list of domains at https://redirect.invidious.io/, included one very popular domain explicitly (yewtu.be) and saw that more than 90% of the remaining instances used `inv.` or `invidious.` subdomains. While this change won't match all instances of invidios, I feel like it's as good approximation for a domain-based heuristic.

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -1328,9 +1328,9 @@ const connectors = [{
 	label: 'Invidious',
 	matches: [
 		'*://*.invidio.us/*',
- 		'*://yewtu.be/*',
- 		'*://invidious.*/*',
- 		'*://inv.*/*',
+		'*://yewtu.be/*',
+		'*://invidious.*/*',
+		'*://inv.*/*',
 	],
 	js: 'connectors/invidious.js',
 	id: 'invidious',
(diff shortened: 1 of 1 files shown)


inverse:
Don't know this connector but how friendly are new instances being added/removed? Would it be too hard to keep updating? 

Alternatively one could map other domains to a connector themselves in the settings...

heyarne:
The appeal of invidious (or piped and other similar software) is that you can self-host it for privacy reasons. There are lists of public instances but they can frequently change. Some addons exist to dynamically redirect you to a list of known instances, which may also change often: https://github.com/libredirect/libredirect#readme

I did not know about the possibility to add other domains in the settings.

heyarne:
Any idea on how to proceed with this?

My intent was to figure out how to deal with what feels to me like an increasing amount of privacy-friendly frontends to popular services, a lot of them already included in this addon. They are by definition available across diverse, unpredictable URLs. I see how a URL pattern matching may cause both false positive and false negative matches, and is therefore not an ideal solution, but I still think it improves the status quo.

I did not know about the option to manually set urls before, and for me that works quite well. Maybe it's a good idea to more boldly mention that somewhere for other users of the services I mentioned?

PR #3503, opened 2022-12-05
Update emby.js

4piu:
* Update selectors to add support for Emby v4.7.10

### DIFF
--- a/src/connectors/emby.js
+++ b/src/connectors/emby.js
@@ -1,12 +1,22 @@
 'use strict';
 
-const trackArtSelector = '.nowPlayingBarInfoContainer .nowPlayingImage';
+const trackArtSelector = [
+	// Emby v4.7.10
+	'.nowPlayingBarInfoContainer .nowPlayingBarImage',
+	'.nowPlayingBarInfoContainer .nowPlayingImage'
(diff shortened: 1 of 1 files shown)


src/connectors/emby.js
```diff
@@ -1,37 +1,56 @@
...
 Connector.trackSelector = [
 	// Jellyfin v10.6.0 and newer
-	'.nowPlayingBarText div:nth-child(1) a',
+	'.nowPlayingBarText div:nth-child(1)',
 	// Emby or Jellyfin v10.5.0 and older
 	'.nowPlayingBarText .textActionButton',
 ];
 
-Connector.trackArtSelector = trackArtSelector;
+// Connector.albumSelector = [
```
inverse, line 24: Should this be commented out?

4piu: Should I remove this comment block? I added the album selector code but it does not work well because element not refreshed when invisible. I will also fix other styling issues.

inverse: Sorry missed this was out for a while - best to remove code that's not used to avoid confusion.

Aside this + lint issues looks good :+1:

PR #3518, opened 2022-12-14
Add Piped connector

### LINKED ISSUE #3241: Add support for scrobbling from Piped
Piped is an open-source alternative frontend for YouTube like Invidious https://github.com/TeamPiped/Piped

kgarner7:
This PR adds a connector for [Piped](https://github.com/TeamPiped/Piped). Resolves #3241

**Additional context**
I only include the official instance (piped.video). Like invidious, there are multiple instances (that change), so it would probably make more sense for people to add custom URLs.

### DIFF
--- a/src/connectors/piped.js
+++ b/src/connectors/piped.js
@@ -1,0 +1,23 @@
+'use strict';
+
+Connector.playerSelector = '#app > div';
+
+Connector.artistTrackSelector = 'div.w-full > div:nth-child(1) > div.font-bold.mt-2.text-2xl.break-words';
+
+Connector.getArtistTrack = () => {
+	const text = $(Connector.artistTrackSelector).text();
+	return Util.processYtVideoTitle(text);
+};
+
+Connector.getCurrentTime = () => $('.shaka-video').prop('currentTime');
+
+Connector.getDuration = () => $('.shaka-video').prop('duration');
+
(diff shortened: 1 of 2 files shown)


inverse (commented):
Thanks for taking the time to contribute! we're trying to phase out the usage of jQuery from connector development. Would it be possible to drop the query usage for vanilla JS or some of the other functions encapsulated in Utils?

kgarner7:
My bad! Thanks for the feedback.

src/connectors/piped.js
```diff
@@ -0,0 +1,26 @@
...
+Connector.artistTrackSelector =
+	'div.w-full > div:nth-child(1) > div.font-bold.mt-2.text-2xl.break-words';
+
+Connector.getArtistTrack = () => {
+	const text = Util.getTextFromSelectors(Connector.artistTrackSelector);
+	return Util.processYtVideoTitle(text);
+};
+
+Connector.getCurrentTime = () =>
+	document.querySelector('.shaka-video').currentTime;
```
inverse, line 14: This could be extracted into a common variable to avoid duplication and improve refactoring later - but it's not a deal breaker breaker :)

inverse (approved):
Thanks for the contribution! it'll be in the next release :)

Not your fault about the jQuery thing - we need to make that clearer

PR #3530, opened 2022-12-21
MyStreamPlayer connector

jbwharris:
This is a new connector I created initially for Bagel Radio. I couldn't find evidence that this streaming service was used by other sites, but wanted to cover my bases in case there were others out there. So it works for https://bagelradio.com and https://mystreamplayer.com URLs currently.

**Additional context**
image

### DIFF
--- a/src/core/connectors.js
+++ b/src/core/connectors.js
@@ -2321,6 +2321,7 @@ const connectors = [{
 	label: 'MyStreamPlayer',
 	matches: [
 		'*://*.mystreamplayer.com/*',
+		'*://*.bagelradio.com/*',
 	],
 	js: 'connectors/mystreamplayer.js',
 	id: 'mystreamplayer',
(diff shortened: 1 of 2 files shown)


src/core/connectors.js
```diff
@@ -2317,6 +2317,15 @@ const connectors = [{
 	],
 	js: 'connectors/uwu-radio.js',
 	id: 'uwu-radio',
+}, {
+	label: 'MyStreamPlayer',
```
inverse, line 2321: Would it be better to make this target specifically bagel radio using the mystreamplayer connector? 

I feel that would be nicer from a user perspective and then have a standalone mystreamone if needed.

cc @hornbuckle as you raised this kinda configuration before.

jbwharris: @inverse That seems reasonable, then it would have Bagel Radio by name in the supported sites too.

hornbuckle: Agreed. Similar to what was done for #3474 and #3477 (and even earlier on #2040).

jbwharris: Just split those off into separate entries

inverse (approved):
Thanks - much better - removed `allFrames` from MyStreamPlayer since we're running directly on that site.

PR #3535, opened 2023-01-01
Add amazing radio connector

kimonoki:
Added connector for amazing radio, including 
https://amazingradio.com/
https://amazingradio.us/

supports both radio and song play

image
image


inverse (approved):
Thanks for the contribution!

PR #3539, opened 2023-01-05
Enhance Synology Support

yefoenix:
Add support for some Synology Audiostation webpage without 5000 port.  

**Additional context**
Need the user open a new tab/window for Synology Audiostation.


inverse:
Thanks for contributing! wasn't aware you could do such things - interesting!

There are a few linting issues but after that we should be good to merge :pray:

PR #3548, opened 2023-01-17
Update Freegal Music connector for redesign

hornbuckle:
Freegal Music launched a redesign of their site this week. It's still using the same player, so the current connector almost works, but the player has been slightly improved and requires some updates to the connector.

- Artist and track now have separate selectors and are only visually truncated with CSS, which makes scrobbling the correct title much easier and has simplified the connector a bit.
- Unfortunately, the issue with extra artist metadata (as outlined in #3335) persists, so the filter to extract the first artist in those cases is still needed.
- Added filter to replace "smart" quotes with plain quotes, as I encountered some titles incorrectly using them since the last connector update.
- Track artwork has been added to the player, but album title is still missing (`alt` attribute on artwork always says "Album" instead of the album title).

### DIFF
--- a/src/connectors/freegalmusic.js
+++ b/src/connectors/freegalmusic.js
@@ -2,37 +2,22 @@
 
 const filter = MetadataFilter.createFilter({
 	artist: extractPrimaryArtist,
-	track: [removeParody, removeRemaster],
+	track: [removeParody, removeRemaster, replaceSmartQuotes],
 });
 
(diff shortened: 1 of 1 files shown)


hornbuckle:
A check for labels seems to have failed, but I don't see a way to add a label to this PR. Assuming that can be added upon review, as done previously?

inverse:
Thanks for the contribution!

PR #3550, opened 2023-01-18
Fix Audacy connector

### LINKED ISSUE #3320: Audacy has redesigned its site
A few weeks ago, Audacy changed the design of its site. The connector should be updated.

https://www.audacy.com/?action=AUTOPLAY_FULL&actionContentId=101-545

hornbuckle:
Resolves #3320. As noted, Audacy has deactivated its main player previously used by this connector.

Updated connector to scrobble from the mini player, which is thankfully still in the DOM when the full-screen player is activated.

Works for both broadcast stations and playlist stations:
- Broadcast station example: https://www.audacy.com/stations/947thewave
- Playlist station example: https://www.audacy.com/stations/princeandfriends

Selectors are odd because the site is built with React and uses randomly generated classes, which cannot be used because they will likely change whenever any site update is pushed.

### DIFF
--- a/src/connectors/audacy.js
+++ b/src/connectors/audacy.js
@@ -1,7 +1,37 @@
 'use strict';
 
-Connector.playerSelector = '#radioPlayer';
-Connector.artistSelector = '#showDescription';
-Connector.trackSelector = '#showName';
-Connector.trackArtSelector = '#poster';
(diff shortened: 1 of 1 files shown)


hornbuckle:
Still not sure what to do about the label check fail. 😕

inverse:
Thanks :pray:

PR #3561, opened 2023-01-30
Fix selector in Audacy connector

hornbuckle:
Audacy made another update to their player where they added more nested divs, which broke the last selector used for the play button in #3550. After looking into it again and testing, the connector works with a less-specific and hopefully less fragile selector—basically removing the divs between the player selector and the play button.

Tested with the same links from the previous PR as well as https://www.audacy.com/stations/metallicaradio which was mentioned in a comment this morning.

### DIFF
--- a/src/connectors/audacy.js
+++ b/src/connectors/audacy.js
@@ -24,7 +24,7 @@ Connector.trackArtSelector = `${audioPlayer} ${buttonFullscreen} img`;
 Connector.isTrackArtDefault = (url) => url.includes('base64');
 
 Connector.isPlaying = () => {
-	const buttonPlaying = Util.getTextFromSelectors(`${audioPlayer} > div:first-of-type > div:nth-of-type(2) > button:not([aria-label=Like]) svg title`);
+	const buttonPlaying = Util.getTextFromSelectors(`${audioPlayer} button:not([aria-label=Like]) svg title`);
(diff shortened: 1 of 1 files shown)


hans-juergen:
Thanks for fixing this connector👍 , currently I am listening and scrobbling successfully their Music Awards Radio featuring the Grammys 2023:

https://www.audacy.com/stations/musicawardsradio

hornbuckle:
You're welcome. Glad it's working out! 😄

hans-juergen:
Just found a little flaw though: if you click on the ^ button in the playbar to show the full page of the radio station, it will pause scrobbling until you close that full page view again:

https://www.audacy.com/music?action=AUTOPLAY_FULL&actionContentId=102-7546

Other than that it seems to work fine, related speech content podcasts for the Grammys are not scrobbled which is good.

I have written a related announcement on the Last.fm support forum in the meantime and added Audacy to the Internet Radio Scrobbling FAQ.

hornbuckle:
Thanks for noting this new issue. Audacy keeps making changes to their site, which, in addition to their markup, is making maintaining this connector difficult. When I initially rebuilt the connector in #3550, the mini player remained in the DOM when the full-screen player was activated, so it would continue scrobbling without interruption. However, it seems they have already changed this behavior and now remove the mini player when the full-screen player is active, so I will have to figure out a way to target both players and will look into this more soon.

Issue #3559, opened 2023-01-29
beatbump.ml not working


awsms:
ungoogled-chromium 109.0.5414.119 / web-scrobbler 2.85.2
https://beatbump.ml/release?type=MUSIC_PAGE_TYPE_ALBUM&id=MPREb_DBaqiU32Mgt

i can see in the console the extension is fetching fine the matadata.. but isPlaying is instantly set to false & then doesn't scrobble anything?
![Screenshot from 2023-01-29 15-47-55](https://user-images.githubusercontent.com/48278661/215334323-4cf5f5f4-0fb4-4833-bb6d-b667d5cfac8c.png)

awsms:
```
l.26	return document.querySelector('.player-btn.player-title svg use').href.baseVal === '/icons-9852f1b5.svg#play';
```
fixed it

jaccarmac:
Please feel free to open a pull request for that change!

awsms:
Done! https://github.com/web-scrobbler/web-scrobbler/pull/3567

inverse:
Will be in the next release - thanks @awsms

PR #3567, opened 2023-02-06
Fix beatbump's connector

awsms:
I changed the filename of the play/pause button so the extension knows the actual isPlaying state for https://beatbump.ml/.
I've also added a regex that matches anything after the last slash of the URL.

Fixes https://github.com/web-scrobbler/web-scrobbler/issues/3559

This is my first pull request on someone else's repo, hope everything is alright!

### DIFF
--- a/src/connectors/beatbump.js
+++ b/src/connectors/beatbump.js
@@ -23,5 +23,5 @@ Connector.getTrackArt = () => {
 
 
 Connector.isPlaying = () => {
-    return document.querySelector('.player-btn.player-title svg use').href.baseVal.replace(/.*\//, '') === 'icons-9852f1b5.svg#pause';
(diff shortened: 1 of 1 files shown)


inverse (commented):
First up thanks for taking the time to contribute. 

- Looks like there is a tiny linting issue that needs to be fixed before we can accept this - let me know if you need help resolving that? 

-  I tried out the change and it works fine - awesome! but I noticed with that site the track information doesnt change when changinge tracks? Did you notice that too?

starise:
I was about to report the issue with beatbump. Hope it'll be merged soon.

inverse (approved):
Fixed the linting issue :+1: 

Looks like the other issue I reported about the site player not updating is resolved too. Good job.

PR #3591, opened 2023-03-09
Add Gimme Metal to existing gimmeRadio connector

### LINKED ISSUE #3588: Gimme Radio Connector Does Not Work For Gimme Metal
Originally Gimme Radio was a single website www.gimmeradio.com.  It has since split into www.gimmecountry.com and www.gimmemetal.com.  There was a pull request to make the connector work for Gimme Country after the split (#2521) but does not work for gimmemetal.com

**How to reproduce**
1.  Go to www.gimmemetal.com
2.  Website not supported

**Expected behavior**
Connector to work the same as it does for gimmecountry.com and previously worked for gimmeradio.com

I'm assuming that gimmemetal.com just needs to be added to connectors.js?

image

image

mattcrav:
Added functionality to Gimme Radio connector so that it functions the same for www.gimmemetal.com as it does for  www.gimmecountry.com. 

- Adds gimmemetal.com urls to the existing gimmeradio connector.
- Adds 'Gimme Metal' to isScrobblingAllowed() function to ignore ads on Gimme Metal.  

Resolves #3588.

### DIFF
--- a/src/connectors/gimmeradio.js
+++ b/src/connectors/gimmeradio.js
@@ -11,6 +11,7 @@ Connector.onReady = Connector.onStateChanged;
 Connector.isScrobblingAllowed = () => {
 	return !(
 		Connector.getArtist().includes('Gimme Country') ||
-		Connector.getArtist().includes('Gimme Radio')
+		Connector.getArtist().includes('Gimme Radio') ||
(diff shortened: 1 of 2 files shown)


inverse (approved):
Thanks for the contribution!

PR #3594, opened 2023-03-16
Update KCRW connector for Today's Top Tune

hornbuckle:
Haven't checked KCRW's Top Tune in a bit, but apparently as of Feb 20 they began including the name of the show with the artist and track in the player. They also changed the separator between the artist and track from a colon to either a dash or en dash. Spacing after the show name and around the separator is also inconsistent.

Updated the connector to account for the above changes. Can be tested at https://www.kcrw.com/music/shows/todays-top-tune

No rush to release, as this only affects Today's Top Tune; other music streams are still working as expected.

### DIFF
--- a/src/connectors/kcrw.js
+++ b/src/connectors/kcrw.js
@@ -1,7 +1,10 @@
 'use strict';
 
 const filter = MetadataFilter.createFilter({
-	artist: replaceSmartQuotes,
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks for the contribution!

Issue #3568, opened 2023-02-09
Edge Upload Broken


inverse:
**Describe the bug**

Publishing to edge looks to be broken - I tried manually publishing via personal account and suspect it could be related to the manifest version being 2 when 3 is required.

When #3228 is resolved I suspect publishing will work.

Issue #3228, opened 2022-04-19
Migrate to V3 manifest


inverse:
We need to migrate to V3 manifest to avoid issues with Chrome.

See: https://developer.chrome.com/blog/mv2-transition/

We have until 2023 to act.

inverse:
@alexesprit  @yayuyokitano looks like we might need to migrate to the V3 setup before then. As the background workers section only takes a single script.

See: https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#manifest

yayuyokitano:
Working on a mini side project in yayuyokitano/Eggs-Enhancement-Suite
Getting a feel for manifest V3 and modern extension compilation

inverse:
Not stale

inverse:
Valid

Issue #2849, opened 2021-05-02
Supress scrobble while editing


myfonj:
**Problem:**
When manual edit takes longer and automatic scrobble moment occurs, unfinished edit is discarded and "wrong" automatic scrobble is sent.

**Solution I'd like**
Do not send automatic scrobble while edit mode is active.

**Alternatives I've considered**
For "fuzzy" recognized tracks do not auto scroble at all, just, open playing notification in edit mode and wait for manual confirmation, or just #1560.

**Additional context**
Sorry if dupe, haven't found prior report. And as always, thanks for all the ext!

yayuyokitano:
To be added in V3. Already working in the fork at least until you start playing a new song.

Issue #2428, opened 2020-05-28
Remove jQuery dependency


alexesprit:
## Current tasks

- [x] Migrate to Bootstrap 5
- [x] Rewrite UI with Vue.js
- [x] Rewrite connectors
- [x] Rewrite `Util` module with using DOM API only

PR #3605, opened 2023-03-24
V3

yayuyokitano:
Web Scrobbler 3.0.0.
Updates to manifest V3
Migrates to TypeScript/SolidJS
Moves from using RequireJS and Gruntfile to using Vite
Adds themes
Adds Safari support
Redesigns UI
Wait with scrobbling until user finishes editing
Removes dependencies (jquery, bootstrap)
Probably more that I forgot

This is still in a _very_ messy state, and I'm sure there are still bugs remaining in the code. However, it is now feature-complete, so I want it out there for review.

Not all tests were migrated, because not all test systems made sense with the new system, these should eventually be replaced, but I think we can do with merging first, and adding the tests later. Most tests are still functional and pass.

Publishing has yet to be done. However, there are build scripts set up so you can just run `npm run build:<target>` to build for chrome/firefox/safari. Note that safari build process requires bash and xcode (command line only, xcrun and xcodebuild), which would complicate any automated process.

Documentation has yet to be updated, in particular some wiki stuff probably needs some changing due to the changes to the structure of the extension.

Linting script also has to be setup, but that should be trivial, as linting has already more or less been setup, just need to call it in the package.json


scripts/make-list.js
inverse, line 1: I think we still need this and the other related script as Firefox required us to provide a list of 3rd party dependencies we bundled in. 

```
# Web Scrobbler

## Third-party projects

-   [Bootstrap v4.6.2][bootstrap]
-   [FontAwesome v6.3.0][@fortawesome/fontawesome-free]
-   [jQuery v3.6.4][jquery]
-   [MD5 v2.19.0][blueimp-md5]
-   [MetadataFilter v1.1.1][metadata-filter]
-   [RequireJS v2.3.6][requirejs]
-   [WebextPolyfill v0.10.0][webextension-polyfill]

[bootstrap]: https://github.com/twbs/bootstrap/releases/download/v4.6.2/bootstrap-4.6.2-dist.zip
[@fortawesome/fontawesome-free]: https://use.fontawesome.com/releases/v6.3.0/fontawesome-free-6.3.0-web.zip
[jquery]: https://code.jquery.com/jquery-3.6.4.min.js
[blueimp-md5]: https://github.com/blueimp/JavaScript-MD5/archive/v2.19.0.zip
[metadata-filter]: https://github.com/web-scrobbler/metadata-filter/releases/download/v1.1.1/metadata-filter-1.1.1.tgz
[requirejs]: https://github.com/requirejs/requirejs/archive/2.3.6.zip
[webextension-polyfill]: https://unpkg.com/browse/webextension-polyfill@0.10.0/dist/
```

Although perhaps they changed things now - especially since this code is bundled.

package.json
```diff
@@ -1,74 +1,68 @@
...
+    "test": "vitest",
+    "lint": "eslint . && stylelint \"**/*.scss\" && remark ."
+  },
   "repository": {
     "type": "git",
-    "url": "https://github.com/web-scrobbler/web-scrobbler.git"
+    "url": "git+https://github.com/web-scrobbler/web-scrobbler.git"
+  },
+  "bugs": {
+    "url": "https://github.com/yayuyokitano/plugin-loader/issues"
```
inverse, line 24:
```suggestion
    "url": "https://github.com/web-scrobbler/web-scrobbler/issues"
```

package.json
```diff
@@ -1,74 +1,68 @@
...
   },
+  "homepage": "https://github.com/yayuyokitano/plugin-loader#readme",
```
inverse, line 26:
```suggestion
  "homepage": "https://web-scrobbler.com/",
```

inverse (commented):
Awesome work so far - tried it out and it worked nicely :)  Great work!

I noticed the README also needs the developer section updated - I dont think I can change anything on your branch as it's not a branch in this repo.

Also is there a way to work on this in watch mode to have it automatically compling?

yayuyokitano:
> Awesome work so far - tried it out and it worked nicely :) Great work!
> 
> I noticed the README also needs the developer section updated - I dont think I can change anything on your branch as it's not a branch in this repo.
> 
> Also is there a way to work on this in watch mode to have it automatically compling?

In theory I think this should be simple, just didnt get around to it. I’ve got a few hours to waste at the airport right now, so I can look into it.

scripts/make-list.js
yayuyokitano: Yeah most of these are removed. I can set up a new file with md5, metadata-filter, webext-polyfill and solid i guess?

yayuyokitano:
Also i think maintainers of web scrobbler should be able to commit to my fork, but im not sure

inverse:
```
To github.com:yayuyokitano/web-scrobbler
 ! [remote rejected]   pr/yayuyokitano/3605 -> pr/yayuyokitano/3605 (permission denied)
error: failed to push some refs to 'github.com:yayuyokitano/web-scrobbler'
```

Alas not :( - I think only modifying via UI

yayuyokitano:
Gave you a full on collaborator invite

yayuyokitano:
Also, codacy is not liking this PR, I think it's too big for it to analyze. Just says failed to analyze in the details

inverse:
> 
Also, codacy is not liking this PR, I think it's too big for it to analyze. Just says failed to analyze in the details


Lets comment it out for now :) I dont think we should let that block us

inverse:
Disabled Codacy app from the repo

scripts/make-list.js
inverse: Yeah I remember Mozilla being really fussy with this before.

yayuyokitano:
Some shenanigans happened in gitlog but I added dev build options. Multiple caveats:

I dont think this will work for safari, but I don't know and don't care all that much.

The build is a _tad_ slow, often a couple seconds, so its not gonna be completely instant, but it's pretty quick.

Background and content script do not auto-reload, you have to manually reload the extension in browser. However, the files _do_ automatically recompile, so you don't have to rerun the command.

Connectors reload automatically, without even having to reload the extension.

Popup and options reload automatically as soon as you refresh the page.

inverse:
Sounds good :) Yeah - I was used to reloading in Firefox etc so that doesn't change anything! 

Awesome work :tada:

yayuyokitano:
Also just to pull your attention to this - the esling linting here is *completely* wrong, it must be reworked into the eslint package. We have an existing config in another repo, but it has many issues. First of all it is using JSdoc, this fork uses TSdoc instead. There is a package for it, you can see the eslintrc of this fork. Similarly, there is no linting around solidjs, and, while I think the vue rules probably include some around general JSX, it also includes other stuff.

inverse:
Ahhh I'm not too familiar with eslint - so would this be something we would want to enable granularity to bring up the quality again? 

Btw this PR also solves #3228

yayuyokitano:
> Ahhh I'm not too familiar with eslint - so would this be something we would want to enable granularity to bring up the quality again?
> 
> Btw this PR also solves #3228

Not entirely sure what you mean. To be clear there is *a* config here, and the code is linted according to it (with warnings but no errors), but the config doesn't completely match with what has been before. I think we'd need to add some of the rules to the eslint config repo, remove some, and we should be good. However, the eslint config is used by multiple repos I think, so gotta be a bit careful about how we change it.

So far the list of issues I know this solves is #1908 #2428 #2849 #3228 #3568

yayuyokitano:
Also it adds full fledged themes, which is a prerequisite that allows us to pretty easily solve #784 in the future

inverse:
Yeah I think we've got a good start here :) can always make the rules a bit stricter later on!

scripts/make-list.js
yayuyokitano: Regarding this, looking into it more I dont think this is needed anymore with bundling as long as the versions are pinned? Not completely sure, but I guess we can readd it if they reject the version. That said the versions are not currently pinned in my fork, so that has to be fixed. I'll be on that tomorrow probably.

We will have to provide source code with each release though (for them to bundle on their end and compare the diffs). Dunno how this is done through a CD process but I can't imagine they made it too hard.

inverse: Interesting - what has your experience been with publishing to their store been? 

I've only experienced this addon and they were a bit fussy. 

> We will have to provide source code with each release though (for them to bundle on their end and compare the diffs). Dunno how this is done through a CD process but I can't imagine they made it too hard.

Interesting - I guess we would just have to zip up the source dir and include that in the zip?

yayuyokitano: Unfortunately I have not yet uploaded an addon that uses bundling and stuff yet, so I can’t say for sure.

In the GUI for uploading there’s an entirely separate upload for source code, so I suspect there will probably be a separate field or function in any CD approach too.

yayuyokitano: Worst case scenario it gets rejected and we have to reupload

inverse:
> npm run dist:x now can be run to run a dist function that creates the distributable file for deployment, and the deploy.yml file has been updated to use it. The script content hasnt been filled in though. It's just compressing to a zip file though, isn't it?

Yeah it should just be zipping up the browser build directory into a zip

Wonder if we should isolate the commands and chain them in the npm scripts section for easier testing...

safari.sh
```diff
@@ -0,0 +1,5 @@
+rm -r build/safari
```
inverse, line 1:
```suggestion
#!/bin/bash

set -euo pipefail

rm -rf build/safari
```

Lets set the shell + enable script mode + use f to stop freak out when dir exists

http://redsymbol.net/articles/unofficial-bash-strict-mode/

build.ts
```diff
@@ -0,0 +1,78 @@
...
+			typeof err !== 'object' ||
+			!('code' in err) ||
+			err.code !== 'EEXIST'
+		) {
+			throw new Error('Something went wrong making build folder');
+		}
+	}
+}
+
+async function createDistributable() {
```
inverse, line 29:
```
import archiver from 'archiver';
import fs from 'fs';
import path from 'path';

async function createDistributable() {
	const curDir = path.resolve(path.dirname(''));
	const outputFile = `${curDir}/web-scrobbler-${process.env.BROWSER}.zip`;
	const output = fs.createWriteStream(outputFile);

	const archive = archiver('zip', {
		zlib: { level: 0 },
	});

	archive.pipe(output);

	archive.directory(`build/${process.env.BROWSER}`, false);
	archive.finalize();
}
```

I guess something like this would work  using the `archiver` package

build.ts
```diff
@@ -0,0 +1,78 @@
...
+	emptydirSync(folder);
+
+	const scripts = [
+		build(configs.buildStart),
+		build(configs.buildBackground),
+		build(configs.buildContent),
+	];
+	await Promise.all(scripts);
+
+	if (process.env.BROWSER === 'safari') {
```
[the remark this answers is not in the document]
yayuyokitano: Yes, the first safari build builds to a separate folder build/safariraw, which is then used by xcode to make the xcode project in build/safari, as safari extensions are actually native programs and xcode has a converter from webextension standard.

yayuyokitano:
> Wonder if we should isolate the commands and chain them in the npm scripts section for easier testing...

The options/popup config takes a massive shit when being run as a command rather than in code. I'm sure it can be fixed, but I'm not sure if it's worth the effort.

yayuyokitano:
dist now creates a zip file for both chrome and firefox. For dist:firefox, it also creates a zip with source code for verification.

yayuyokitano:
Not getting any diffs from running npm ci + npm run dist:firefox on the extracted src vs the original dist run. Within same computer though.

The actual zip files have different checksums though, but the extracted content seems to match.

scripts/create-distributable.ts
```diff
@@ -0,0 +1,67 @@
...
+
+	const output = fs.createWriteStream(outputFile);
+
+	const archive = archiver('zip', {
+		zlib: { level: 0 },
+	});
+
+	archive.pipe(output);
+
+	// TODO: make a glob that gets the top level files and doesnt nuke the system
```
inverse, line 22: What was the issue here?

yayuyokitano: I think it went recursive and made a huge archive that nearly crashed my computer

yayuyokitano:
> Looks good to me and we can improve stuff as we go! I guess the main issue now is how can we upload the src distribution for Firefox right?

The API has a field for it, but the action we are using hasnt seen an update in years and doesn’t support it. As an exercise I think I’ll just write a new action since it seems relatively simple and I haven’t done that before.

scripts/create-distributable.ts
yayuyokitano: Fixed it now that I had some fresh eyes

inverse (approved):
Think this puts us in a good starting point to iterate on!

yayuyokitano:
I think we can merge this then. Note that this PR is aimed at a separate branch for now, not master.

inverse:
We can change that by clicking edit in the PR - it says old commits may be removed... but there isnt much difference between the two bases just a few new changes.

https://github.com/web-scrobbler/web-scrobbler/compare/feature/manifestv3...master

PR #3616, opened 2023-04-01
Add connector for copyparty

9001:
This adds support for [copyparty](https://github.com/9001/copyparty), a self-hosted file sharing service. As most copyparty instances are for private use, I've included patterns for the public demo server and for the default port when running the server locally.

These changes will work on all copyparty versions starting from v1.6.11 onwards. If you'd like to test the changes, please drop by the [demo server](https://a.ocv.me/pub/demo/music/Ubiktune%20-%20SOUNDSHOCK%202%20-%20FM%20FUNK%20TERRROR!!/) :>

**Additional context**
The same metadata is available through the media session api; however, the album cover URL inside the mediasession tags has a login token included as a URL parameter. Since everything works as expected even without this token, it felt like a good idea to strip it away before passing the image URL to the extension.

### DIFF
--- a/src/connectors/copyparty.ts
+++ b/src/connectors/copyparty.ts
@@ -12,6 +12,6 @@ Connector.trackSelector = '#np_title';
 Connector.currentTimeSelector = '#np_pos';
 Connector.durationSelector = '#np_dur';
 
-this.getUniqueID = () => {
+Connector.getUniqueID = () => {
 	return Util.getTextFromSelectors('#np_url');
 };
--- a/src/connectors/copyparty.ts
+++ b/src/connectors/copyparty.ts
@@ -13,5 +13,5 @@ Connector.currentTimeSelector = '#np_pos';
 Connector.durationSelector = '#np_dur';
 
 this.getUniqueID = () => {
-	return document.getElementById('np_url').textContent;
+	return Util.getTextFromSelectors('#np_url');
 };
--- a/src/connectors/copyparty.ts
+++ b/src/connectors/copyparty.ts
@@ -1,0 +1,17 @@
+export {};
+
+Connector.playerSelector = '#np_inf';
+Connector.pauseButtonSelector = '#np_inf.playing';
+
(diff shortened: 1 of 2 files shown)


9001:
Nice, congratulations on v3 :tada:  can't wait to see the new settings UI!

I have rebased the branch on master; just let me know if it would be preferable to scrap the branch entirely and start afresh :>

src/connectors/copyparty.ts
```diff
@@ -0,0 +1,17 @@
...
+Connector.albumArtistSelector = '#np_circle';
+Connector.albumSelector = '#np_album';
+Connector.artistSelector = '#np_artist';
+Connector.trackSelector = '#np_title';
+
+Connector.currentTimeSelector = '#np_pos';
+Connector.durationSelector = '#np_dur';
+
+this.getUniqueID = () => {
+	return document.getElementById('np_url').textContent;
```
hornbuckle, line 16: Could this just use `Util.getTextFromSelectors('#np_url')` instead?

9001: Absolutely, wasn't aware of that helper. Will fix that 👍

src/connectors/copyparty.ts
```diff
@@ -0,0 +1,17 @@
...
+Connector.trackArtSelector = '#np_img';
+Connector.albumArtistSelector = '#np_circle';
+Connector.albumSelector = '#np_album';
+Connector.artistSelector = '#np_artist';
+Connector.trackSelector = '#np_title';
+
+Connector.currentTimeSelector = '#np_pos';
+Connector.durationSelector = '#np_dur';
+
+this.getUniqueID = () => {
```
hornbuckle, line 15: Sorry, missed this in my previous comment, but shouldn't this be `Connector.getUniqueID`?

9001: Hah, you're right! Guess that didn't do much huh...

PR #3620, opened 2023-04-03
V3

yayuyokitano:
Web Scrobbler 3.0.0.
Updates to manifest V3
Migrates to TypeScript/SolidJS
Moves from using RequireJS and Gruntfile to using Vite
Adds themes
Adds Safari support
Redesigns UI
Wait with scrobbling until user finishes editing
Removes dependencies (jquery, bootstrap)
Probably more that I forgot

This is still in a very messy state, and I'm sure there are still bugs remaining in the code. However, it is now feature-complete, so I want it out there for review.

Not all tests were migrated, because not all test systems made sense with the new system, these should eventually be replaced, but I think we can do with merging first, and adding the tests later. Most tests are still functional and pass.

Publishing has yet to be done. However, there are build scripts set up so you can just run npm run build:<target> to build for chrome/firefox/safari. Note that safari build process requires bash and xcode (command line only, xcrun and xcodebuild), which would complicate any automated process.

Documentation has yet to be updated, in particular some wiki stuff probably needs some changing due to the changes to the structure of the extension.

Linting script also has to be setup, but that should be trivial, as linting has already more or less been setup, just need to call it in the package.json


inverse (approved):

yayuyokitano:
Closes #2428 #2849 #3228 #3568

PR #3622, opened 2023-04-04
Add sourcemaps in dev mode

inverse:

Provide easier debugging in dev mode with sourcemap output

**Additional context**


yayuyokitano (approved):
LGTM, this does remind me that the connectors are *always* compiled with source maps. We probably want to change that too soon. A bit more complex since connectors are just based on the tsconfig.connectors.json file, which we cannot set dynamic values in, but nothing too complicated.

PR #3624, opened 2023-04-07
Overhaul icons

yayuyokitano:
Overhauls icons, mostly to work better with safari, which strongly prefers monochrome icons due to tinting shenanigans.
Completely reworks icons to use different shapes for all different action types, instead of relying exclusively on color to distinguish several action types.

Introduces canvas as a dev dependency, and a custom vite plugin to take svgs from src and convert and write the appropriate png files based on those svgs at build time.

This system is also used for the main extension icons. They end up looking a little different than before, and a little bigger in file size (compensated for by now only including one set in each distribution), but for the most part I think it is an acceptable difference. (Maybe with the exception of the chrome 48x48 icon which I would encourage taking an extra close look at. It is arguably the most important icon).

Also introduces new sizes, now using 16x16 and 32x32 as well for action icons, as 16-multiple icon sizes have been basically standard for a few years now AFAIK. Still also bundles 19x19 and 38x38, since the size difference is negligible. Also uses 96x96 main icons across all browsers, instead of just safari.

This PR probably needs some time to cook, maybe someone more knowledgeable than me on vector graphics can chip in. I don't think merging it as is is horrible either though.


manifest.config.ts
```diff
@@ -46,24 +38,31 @@ export const common: Manifest.WebExtensionManifest = {
 		page: 'src/ui/options/index.html',
 		open_in_tab: true,
 	},
+
+	icons: {
+		16: 'icons/icon_generic_16.png',
```
inverse, line 43: Why does this have a different name?

manifest.config.ts
```diff
@@ -86,17 +83,21 @@ export const firefoxManifest: Manifest.WebExtensionManifest = {
...
+ * Gets action with defaults for a browser
  *
- * @param browser - browser to get icons for
- * @returns object of icon paths
+ * @param browser - browser to get action for
+ * @returns manifest action object
  */
-function browserIcons(browser: string) {
+function getAction(browser?: string) {
+	const type = browser === 'safari' ? 'safari' : 'light';
```
inverse, line 92: This reads strange - so the type is either `safari` or `light`?

scripts/generate-icons.ts
```diff
@@ -0,0 +1,285 @@
...
+	},
+	'action_error.svg': {
+		light: '#fea800',
+		dark: '#fea800',
+	},
+	'action_ignored.svg': {
+		light: '#fea800',
+		dark: '#fea800',
+	},
+	'action_loading.svg': {
```
inverse, line 55: For me this one I couldn't see - it was the same colour as my browser .

image

inverse (commented):
Looks a lot different - but in a good way :) 

Left some comments!

manifest.config.ts
yayuyokitano: It uses a different icon. This was the case in v2 as well, where there is a generic icon (which is just red and white) for 16x16 due to it being so small. I dont know where this icon actually is used but I kept it to be safe

manifest.config.ts
yayuyokitano: Yes, I could probably name this better, basically this gets the type of the default icon. We cant know what theme the user has here so on the non safari browsers we simply have to pick either light or dark. It will change to the users selected theme once any navigation event happens that causes web scrobbler to refresh icon

scripts/generate-icons.ts
yayuyokitano: Interesting, that should be white since you have dark theme enabled. Could you try to repro and look for errors in the service worker?

yayuyokitano: I updated the theme fetching and it should be more reliable now I think.

manifest.config.ts
yayuyokitano: Updated this to be clearer

scripts/generate-icons.ts
inverse: Not sure what is up :/ Something with the transition with the action item causes the icon to go away for me until it's resolved :/ very odd.

src/core/background/action.ts
```diff
@@ -103,20 +101,45 @@ function getIconURL(url: string) {
...
+		const clonedSong = new ClonedSong(tab.song, -1);
 		songstr = `${clonedSong.getArtist()} - ${clonedSong.getTrack()}`;
 	}
 
 	browser.action.setIcon({
 		path: {
-			19: getIconURL(`icons/page_action_${mode.toLowerCase()}_19.png`),
-			38: getIconURL(`icons/page_action_${mode.toLowerCase()}_38.png`),
+			16: getIconURL(
+				`icons/action_${tab.mode.toLowerCase()}_16_${await getIconType()}.png`
```
inverse, line 114: Should we add like

`const iconType = await getIconType();` 

And reuse that to avoid having to call `getIconType` over and over?

scripts/generate-icons.ts
inverse: Okays - it's nothing to do with this.. It's some performance issue on my end - if I change `getIconType` to just return `dark` it performs as expected. something with `getBrowserPreferredTheme` is causing it to lag loading :/

inverse: i did - the loading icon never showed unless i made that function static value return. I'll record a video tonight.

yayuyokitano: Can you run the following functions in any tab (not service worker) and show results
`window.matchMedia('(prefers-color-scheme: dark)').matches`
`window.matchMedia('(prefers-color-scheme: light)').matches`

and then run the following function in service worker or extension settings/popup and show results:
`chrome.storage.local.get("StateManagement", console.log)`

src/core/background/action.ts
yayuyokitano: Done, also made some other changes to the same part in the spirit of DRY

scripts/generate-icons.ts
inverse:
```
window.matchMedia('(prefers-color-scheme: light)').matches
true
window.matchMedia('(prefers-color-scheme: dark)').matches
false
```

So I'm guessing I'm running in light mode by default. 

```
chrome.storage.local.get("StateManagement", console.log)
undefined
```


Here's a video of the behaviour on latest branch https://user-images.githubusercontent.com/713525/231571816-b82845cc-2220-4b71-b2c4-6b3d0282f269.mp4

yayuyokitano: Looked into it, turns out prefers-color-scheme just [straight up doesnt work in chrome since 3 years ago](https://bugs.chromium.org/p/chromium/issues/detail?id=998903) ;-;
Apparently a fix is being pushed to version 114, _maybe_ 113.

That sucks. We can't get the users preferred theme for a while then.

yayuyokitano: I guess i change the colors up so that we can afford dealing with chrome being terrible at following web standards, task for tomorrow though

inverse: Strange - I often wondered why my UI was dark yet was always served a light theme :D Probably some linux quirk too as I suspect it runs fine on other OS

yayuyokitano: Yeah based on the issue it seems to be a linux issue

inverse: Could be a tough one as Chrome supports the user customising the browser colour. 

I just changed it to something less dark and it works perfectly.

yayuyokitano: Think I have an idea to failsafe this, gonna be a slight pain to implement though but shouldnt be too bad

yayuyokitano: Alright so this now creates borders on the icons. It's not exactly pretty if your browser theme color is far off the expected color for your preferred theme, but it does completely eliminate any possibility of issues occurring, and is hardly visible with standard settings (other than linux chrome with dark theme until chrome 114). Safari icons unaffected by this change.

inverse: Seems legit to me from testing :)

scripts/generate-icons.ts
```diff
@@ -0,0 +1,343 @@
...
+import { createCanvas, loadImage } from 'canvas';
+import { resolve } from 'path';
+import { getBrowser, releaseTarget } from './util';
+import colorLog from './log';
+import { PluginOption } from 'vite';
+
+/**
+ * Input folder containing the original svg icons in the source.
+ */
+const input = resolve('', 'src', 'icons');
```
inverse, line 11: Are you sure the first empty `''` is needed - won't it resolve the same?

scripts/generate-icons.ts
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * Writes all the appropriate PNG files for a single monochrome svg.
+ *
+ * @param path - filename of svg.
+ */
+async function writeMonochromeIcon(
+	path: keyof typeof monochromeColors
+): Promise<void> {
+	if (releaseTarget === 'safari') {
```
inverse, line 215: We could have an enum/type that represents these release targets - but could be a future iteration :)

scripts/generate-icons.ts
```diff
@@ -0,0 +1,343 @@
...
+}
+
+/**
+ * Gets new PNG file name for misc icons
+ *
+ * @param path - filename of original svg.
+ * @param size - desired resolution of PNG.
+ * @returns PNG file name to output to.
+ */
+function getMiscOutputPath(path: string, size: number) {
```
inverse, line 267:
```suggestion
function getMiscOutputPath(path: string, size: number): string  {
```

inverse (approved):
Big improvements here :)

scripts/generate-icons.ts
yayuyokitano: You're right, fixed.

scripts/generate-icons.ts
yayuyokitano: Did it anyway

yayuyokitano:
I think this is probably good to go now. I'm still a little bothered by the chrome 48x48 icon though, it seems to pop a bit less than the old one.

inverse:
That's true - Do you think it's just the format of them? Like lacking the square shape they had before?

yayuyokitano:
the actual main icons are still square and completely unmodified from the svg used before, so that's not it. It looks to me like the antialiasing is somehow prioritizing white less than before maybe

inverse:
Could be - I introduced merge conflict merging dependabot changes :see_no_evil: sorry

yayuyokitano:
It's just package lock conflict so i guess we can just pick one and then regenerate lockfiles after

yayuyokitano:
That should do it

inverse (approved):

inverse:
```
malachi@pulsar web-scrobbler $ npm run build dev chrome

> web-scrobbler@3.0.0 build
> tsx build.ts "dev" "chrome"

Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.

If you are installing esbuild with npm, make sure that you don't specify the
"--no-optional" or "--omit=optional" flags. The "optionalDependencies" feature
of "package.json" is used by esbuild to install the correct binary executable
for your current platform.
```

Get this when running locally now :/ wonder what changed

yayuyokitano:
I added esbuild as a direct dev dependency, changed up the lockfile a fair bit. Does it work on your end now?

inverse:
Solved!

yayuyokitano:
You think we're good to go for merge? Also the major tag isn't going to accidentally create v4 right, I assume it still stays at 3 since its draft

inverse:
I think we should be good :) 

It'll end up in the v3 draft since that's not published yet  - we should think about a release I guess soon :)

yayuyokitano:
Still have a couple pretty big issues im getting somewhat sporadically that i would really prefer to manage to iron out first, but yeah hopefully we can get it out within a week maybe? Anyway merging

inverse:
image

Just installed on my work laptop and notice the border looks a bit strange :/

yayuyokitano:
> image
> 
> Just installed on my work laptop and notice the border looks a bit strange :/

See #3653 for an attempted improvement

PR #3647, opened 2023-04-15
Bring back edge publish

inverse:

Previously removed due to suspected v3 manifest requirement

**Additional context**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -88,7 +88,7 @@ jobs:
         run: npm ci
 
       - name: Create zipball
-        run: npx run dist chrome
+        run: npm run dist chrome
 
       - name: Get version from package
         id: get-package-version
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -67,3 +67,39 @@ jobs:
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '${{ steps.get-node-version.outputs.NODE_VERSION }}'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Create zipball
+        run: npx run dist chrome
```
yayuyokitano, line 91:
```suggestion
        run: npm run dist chrome
```

inverse: Good catch! I guess we should change the Chrome one too!

yayuyokitano: Ah yeah, missed that one

PR #3649, opened 2023-04-16
Add permission request to settings

yayuyokitano:
Adds a popup-esque thing on top of settings that asks the user to give host permissions if they didn't already. Helps a bit with UX, since this permission is no longer automatically granted in firefox/safari with manifest V3.

image

### DIFF
--- a/src/ui/options/components/permissions.tsx
+++ b/src/ui/options/components/permissions.tsx
@@ -3,6 +3,10 @@ import browser from 'webextension-polyfill';
 import styles from './components.module.scss';
 import { t } from '@/util/i18n';
 
+const desiredPermissions = {
+	origins: ['http://*/', 'https://*/'],
+};
+
 export default function Permissions() {
 	const [perms, setPerms] = createSignal(true);
 	hasPermissions(setPerms);
@@ -15,9 +19,7 @@ export default function Permissions() {
 					onClick={() => {
 						// This is inline and not async, as when it was not there were some issues with safari isTrusted
 						browser.permissions
-							.request({
-								origins: ['http://*/', 'https://*/'],
-							})
+							.request(desiredPermissions)
 							.then(setPerms);
 					}}
 				>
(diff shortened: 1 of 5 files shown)


yayuyokitano:
As some context on why I have made the decision I have made regarding how to trigger, browser.permissions.request must be triggered by a user action

src/ui/options/components/permissions.tsx
```diff
@@ -0,0 +1,37 @@
...
+import browser from 'webextension-polyfill';
+import styles from './components.module.scss';
+import { t } from '@/util/i18n';
+
+export default function Permissions() {
+	const [perms, setPerms] = createSignal(true);
+	hasPermissions(setPerms);
+
+	return (
+		<Show when={!perms()}>
```
inverse, line 11: How does this work? how does `perms()` know what perms are set?

src/ui/options/components/permissions.tsx
```diff
@@ -0,0 +1,37 @@
...
+			<div class={styles.permissionsPopup}>
+				<button
+					type="button"
+					onClick={() => {
+						// This is inline and not async, as when it was not there were some issues with safari isTrusted
+						browser.permissions
+							.request({
```
inverse, line 18: Would we extract the permissions into a variable for keeping things DRY?

yayuyokitano: Sure can, give me a bit

src/ui/options/components/permissions.tsx
yayuyokitano: perms is the getter function of a [signal](https://www.solidjs.com/tutorial/introduction_signals)

The signal is set to true by default, then the signal setter setPerms is uses to set it to false if perms are not granted, and set back to true if the user grants permission.

Signals allow the component to be reactive – solid only renders the component once, so if you just use a raw variable it will not change what is rendered when you change the variable value. Signals (and other similar solid utilities) tell the component to actually change what is rendered too.

src/ui/options/components/permissions.tsx
yayuyokitano: Done

src/ui/options/components/permissions.tsx
inverse: Thanks for the explanation thought it was something like that :)

inverse (approved):

PR #3651, opened 2023-04-16
Fix locking

yayuyokitano:
Issue in the locking algorithm caused actions to update slowly sometimes.
This solves the issue, in some cases speeding the extension updates up by several seconds.
Also does a little bit of general cleanup to state changing in background code.


inverse (approved):
Nice work :clap:

PR #3653, opened 2023-04-17
Use offset icons for border

yayuyokitano:
Should make the icon borders more uniform and a bit nicer

Old:
image

New:
image


inverse (approved):
Magic :+1:

PR #3654, opened 2023-04-17
Add safari build process to workflow

yayuyokitano:
Also migrates entire build workflow to macos, but has no impact on existing workflows.

EDIT: also makes the build process take significantly longer, but I don't think theres a way to avoid this

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,3 +46,9 @@ jobs:
       - name: Build the extension for Safari
         run: |
           npm run build safari
+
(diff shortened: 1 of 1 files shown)


inverse (approved):
Seems reasonable unless we want to go down a per job build :)

yayuyokitano:
> Seems reasonable unless we want to go down a per job build :)

I think it's probably a good idea to do this now given how long this is taking now. I think it is probably best done as a separate PR though.

PR #3655, opened 2023-04-17
Correct PR label workflow

yayuyokitano:
Noticed the array here is broken. I think this should fix it.


yayuyokitano:
@inverse output clearly shows the PR labels workflow isn't running the workflow from my branch (even though it claims to), any idea what to do?

yayuyokitano:
Ok I was able to actually test this using stupid means, it works as intended currently, ready to merge I think.

yayuyokitano:
This PR is failing PR Labels workflow because of the very bug it fixes – how ironic.

PR #3656, opened 2023-04-17
Preload options

yayuyokitano:
#3652 already fixed any speed issue there may have been due to not preloading, but left the transition to the options page looking pretty bad, with toggles flickering and select text popping in.

By preloading the options, it ensures the user's preferred settings are loaded before going to the page, ensuring no flickering.

Makes some additional changes to the loading of override options to reduce lag. (Does not render connector options until the user clicks to expand, did some modification to existing lazy loading too)

### DIFF
--- a/src/ui/options/components/options/connector-override.tsx
+++ b/src/ui/options/components/options/connector-override.tsx
@@ -41,15 +41,11 @@ const [overrideOptions, setOverrideOptions] = createResource(
 const [customPatternOptions, setCustomPatternOptions] = createResource(
 	customPatterns.get.bind(customPatterns)
 );
-console.log(options.loading);
 
 /**
  * Component that shows the override options for all connectors
  */
 export default function ConnectorOverrideOptions() {
-	console.log(options());
-	console.log(overrideOptions());
-	console.log(customPatternOptions());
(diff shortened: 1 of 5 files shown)


src/ui/options/components/options/connector-override.tsx
```diff
@@ -25,19 +32,24 @@ const customPatterns = BrowserStorage.getStorage(
...
-	const [options, setOptions] = createResource(
-		globalOptions.get.bind(globalOptions)
-	);
-	const [overrideOptions, setOverrideOptions] = createResource(
-		connectorOverrideOptions.get.bind(connectorOverrideOptions)
-	);
-	const [customPatternOptions, setCustomPatternOptions] = createResource(
-		customPatterns.get.bind(customPatterns)
-	);
+	console.log(options());
```
inverse, line 50: Should we remove these log lines?

inverse (approved):
LGTM aside the logging

src/ui/options/components/options/connector-override.tsx
yayuyokitano: whoops, forgot to clean those up

PR #3657, opened 2023-04-17
Add tx support + pull updated translations

inverse:

Got `tx` command line working to pull/push from the repo.

Perhaps should document the manual steps in https://github.com/web-scrobbler/web-scrobbler/wiki/Translate-the-extension

**Additional context**


.tx/config
```diff
@@ -0,0 +1,17 @@
...
+host = https://www.transifex.com
+
+[o:web-scrobbler:p:web-scrobbler:r:messages-json--master]
+file_filter   = src/_locales/<lang>/messages.json
+source_file   = src/_locales/en/messages.json
+type          = CHROME
+minimum_perc  = 0
+resource_name = messages.json
+
+[o:web-scrobbler:p:web-scrobbler:r:privacy-md--master]
```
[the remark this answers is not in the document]
yayuyokitano: I think it can probably stay, and we can rehandle it in the future

src/_locales/de/messages.json
```diff
@@ -23,6 +27,18 @@
         "message": "Benachrichtigung anzeigen wenn Titel nicht von der Erweiterung erkannt wird",
         "description": "Option title"
     },
+    "optionScrobblePodcasts": {
+        "message": "Scrobble podcasts",
```
inverse, line 31: Not a fan of this "default" behaviour to pre-fill missing translations with src lang as it might encourage people to modify on here vs on the Transflix UI

yayuyokitano: doesnt seem ideal yeah

src/_locales/de/messages.json
```diff
@@ -31,30 +47,42 @@
...
         "description": "Option title"
     },
     "optionScrobblePercentSuffix": {
-        "message": "der Songlänge",
+        "message": "of song duration",
         "description": "Option title"
     },
     "optionYtDesc": {
-        "message": "Wenn du beide Optionen deaktivierst, wird die Erweiterung YouTube videos aus allen Kategorien scrobblen.",
+        "message": "If you turn off both options, the extension will scrobble YouTube videos from all categories.",
```
yayuyokitano, line 79: It seems like it's killing a lot of existing translations (just one example). Is this not in transifex?

inverse: There is an `--mode` flag which has `onlytranslated` but that leaves the `message` field empty :/ 

```
     "optionScrobbleEditedTracksOnlyTitle": {
-        "message": "Only scrobble tracks if they are edited",
+        "message": "",
         "description": "Option title"
     },
```

inverse: Exactly! Feels like it would be more problematic :/

inverse: Updated with script as you suggested

.tx/config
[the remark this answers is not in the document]
inverse: Need to look into whats happening here, for some langs its pulling down in source lang :/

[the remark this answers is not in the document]
yayuyokitano: I checked on tx when you gave me admin and at least for the german stuff I found that the translations are not on transifex, probably added directly to repo at some point. I think this might be the case across the board but I didn’t check extensively

src/_locales/de/messages.json
```diff
@@ -31,30 +31,6 @@
...
-    },
-    "optionsHidden": {
-        "message": "Versteckt",
-        "description": "'Hidden' section"
-    },
-    "optionScrobblePercent": {
-        "message": "Scrobble bei",
-        "description": "Option title"
-    },
-    "optionScrobblePercentSuffix": {
```
inverse, line 50: Wonder why some of these are gone :/

yayuyokitano: Same reason as the english ones i think, directly committed at some point and not in transifex

scripts/translations-prepare.ts
```diff
@@ -0,0 +1,23 @@
...
+	let modified = false;
+	Object.keys(json).forEach((key) => {
+		if (json[key]['message'] === '') {
+			delete json[key];
+			modified = true;
+		}
+	});
+
+	if (modified) {
+		console.log(`Updating ${filePath}`);
```
yayuyokitano, line 20: Use the colorLog utility from log.ts for consistency

yayuyokitano (changes requested):

src/_locales/de/messages.json
inverse: Ahh it's not in the source - I guess someone modified manually - updated on transfix

package.json
```diff
@@ -52,6 +53,7 @@
     "eslint-plugin-solid": "0.12.1",
     "eslint-plugin-tsdoc": "0.2.17",
     "fs-extra": "11.1.1",
+    "glob": "^10.2.1",
```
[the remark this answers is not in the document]
inverse: Can you explain why? pin? is that for allowing dependabot to do its magic?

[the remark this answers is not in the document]
yayuyokitano: General good practice in case something happens, as having the ^ means auto-updating to newer non-major versions without any approval. Also repeatability of builds but that doesnt really matter for this dependency.

.tx/config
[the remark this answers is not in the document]
inverse: Fixed up - was missing some bits :/

src/_locales/de/messages.json
inverse: Fixed with the script

src/_locales/de/privacy.md
inverse, line 1: No idea why this file is being pulled there are zero translations here :/

src/_locales/zh/messages.json
```diff
@@ -279,10 +279,6 @@
         "message": "作者是谁？",
         "description": "Question #7"
     },
-    "faqAnswer7": {
-        "message": "本扩展的原作者是 David Šabata 。现在该扩展由 Web Scrobbler 团队开发和维护。",
```
yayuyokitano, line 283: Another translation that disappeared

yayuyokitano (changes requested):

src/_locales/zh/messages.json
inverse: Good catch :/ Fixed

yayuyokitano (approved):
Obviously theres still the issue of the "german" privacy.md that suddenly appeared, but i also dont think it really affects anything

inverse:
Yeah that one's an odd one :/ I still don't understand why that was pulled but yeah doesnt break anything :/

PR #3660, opened 2023-04-18
Split build workflow into multiple jobs

yayuyokitano:
Makes the build workflow not take quite as long (but still usually a bit longer than before safari build).

Not very DRY, but it does the trick.


inverse (approved):
LGTM :pray:

PR #3662, opened 2023-04-21
Fix context menu click event

yayuyokitano:
Issue caused context menu click event to be unreliable, this should fix it.


inverse (approved):
Nice one!

PR #3671, opened 2023-04-23
Add Regex/Bulk Edits

### LINKED ISSUE #1552: Custom filters
Add user created fillters

### LINKED ISSUE #1643: Scrobble filters (Remastered, Explicit)
On YouTube, the add-on has a filter that removes "official video" from the tags. Can you extend this feature into other websites?

Or remove the following case-sensitive phrases from track or albums: "explicit", "remastered" (with any year),...

### LINKED ISSUE #2157: Option to remember Artist/Album name correction
Sometimes Artist or Album name is misspelled on website (e.g. Yandex Music).  
Thus, I want to modify info for each and every song on the album currently playing.  
This is very annoying when I do some parallel activities. Because:
1. I have to modify the data for each song separately (e.g. when album has 15 songs - I have to do the same action 15 times!)
2. I have to switch to the tab where my music is playing (yep, 15 times! and 15 times back... - #1581 is already raised for this)
3. I have to be quick enough - since I can only modify the data during the 1st half of the song (this is a room for another improvement I have in mind - to be able to configure extension to send the scrobble only after song is finished, hence allowing to modify the data during the whole track length; _update_ - raised #2158 for this)

Add some option (checkbox?) to remember my Artist/Album name correction and auto-correct all the tracks the same way on those page.

### LINKED ISSUE #2588: Option for scrobbling only the first/main artist

### LINKED ISSUE #2608: Option to automatically convert latest comma in artists field into ampersand

### LINKED ISSUE #2943: Regex-based filters

### LINKED ISSUE #2948: Global correction (find and replace) for any word
I know global (user specific) corrections have been suggested before in some way for artists and albums, but given the solutions proposed for those, what I'm suggesting is a bit different.  A feature that works like "find and replace" for any word or set of characters in the artist or song name field (or album and album artist field) that you could set and and the scrobbler would apply the correction to everywhere where it finds the word/character pattern.

For example if I would want "Radio edit" to always be corrected to just "Remix", I could do that. This could also be used to permanently prevent known misspelled artist/song names to be scrobbled without having to always edit them manually.

This would solve multiple track correction-related problems people have and it would be for everyone to create their own custom corrections so it would only affect themselves, which I see as a big plus. The info from these corrections could be of course utilized with the scrobblers automatic corrections too.

### LINKED ISSUE #3046: Please include pattern edits with the extension!
The Web Scrobbler's built-in metadata filter doesn't automatically remove the " - Single", " - EP", " - Bonus" / "Bonus Tracks Version" and many other suffixes from the album field when scrobbling music using the Apple Music web player. It's too time consuming and cumbersome to have to do it manually everytime I play music. 

I want the functionality to be available in the extension to have the extension do this automatically or at least the ability to import the pattern edits (see the attached picture) from the scrobbler app on my phone to the Web Scrobbler extension.

Screenshot_20211105_151052_com arn scrobble

### LINKED ISSUE #3069: Create Sintax Rules

I'm a longtime last.fm user of many years. Right know I'm a Youtube Music user with the spanish language as the primary language of the site (I know I can change it to english). 
The problem is that Youtube Music is forcing the translation of TITLE ID of any song like this:

SONGTITLE (feat. ARTISTNAME) 

to

SONGTILE (con ARTISTNAME)

"featuring" to "con"

I would love to know if maybe you could add the option add a rule to whenever this happens we can force the "con" to "feat." so I can scrobble the same songs I have on my hard drive and sd card music collection.

thanks for your work

### LINKED ISSUE #3186: Automatically remove the " - Single" and " - EP" suffixes from scrobbles when playing music from the Apple Music web player.
The Web Scrobbler's built-in metadata filter doesn't automatically remove the " - Single", " - EP", and many other suffixes from the album field when scrobbling music using the Apple Music web player. It's too time-consuming and cumbersome to have to do it manually every time I play music.

I want the functionality to be available in the extension to have the extension do this automatically.

### LINKED ISSUE #3432: Fallback for scrobbling songs with multiple artists

I'm having trouble scrobbling songs from 電音部/DENONBU (https://music.youtube.com/channel/UCA22VRlaMmb2B1-ki2MLLBw) on YouTube Music. It always says "The track is not recognized" until I manually add it.
YouTube Music has 電音部 and the singers in the artist field and often has the composer in brackets in the title.
Using the built in scrobbling feature of Spotify it only uses the first artist (which is always 電音部). Additionally Spotify does not have the composer in brackets in the title. It's artist field usually contains "電音部, composer, singers".

Example: YouTube Music has "`DENONBU, Tama Kurogane (CV: Akina), Aki Shirokane (CV: Arisa Komiya) & Ginka Haijima (CV: Azuki Shibuya) - Azabu Autobahn (Prod. Kenmochi Hidefumi)`" and web-scrobbler tries to scrobble "`DENONBU, Tama Kurogane (CV: Akina), Aki Shirokane (CV: Arisa Komiya), Ginka Haijima (CV: Azuki Shibuya) - Azabu Autobahn (Prod. Kenmochi Hidefumi)`".
Spotify has "`電音部, 黒鉄たま (CV: 秋奈), 白金 煌 (CV: 小宮有紗), 灰島銀華 (CV: 澁谷梓希) - Azabu Autobahn`" and scrobbles "`電音部 - Azabu Autobahn`".

Looking at the links to test in the YouTube Music connector source code, the first one (https://music.youtube.com/playlist?list=OLAK5uy_kDEvxPASaVnoSjOZViKEn4S3iVaueN0UI) also fails. web-scrobbler tries to scrobble "`Mt. Eden, T-Mass - As One (feat. Aviella Winder)`". Last.fm has the following entries for it, in order of popularity:
- Mt. Eden - As One
- Mt Eden - As One
- Mt. Eden & T-Mass - As One (feat. Aviella Winder) 
- Mt Eden - As One (feat. Aviella Winder)
- Mt Eden, T-Mass feat. Aviella Winder - As One
- Mt. Eden, T-Mass feat. Aviella Winder - As One
- Mt. Eden - As One (feat. Aviella Winder)

I would want some kind of fallback, so it tries to scrobble with only the first artist (or maybe also with a & instead of a comma) rather than just failing (and maybe also without the feat. or Prod. part). 

#3306 proposes to always only scrobble one artist in YouTube Music.

**Additional context**
I think that trying to scrobble with all artists is good, but it often creates problems and I'd rather just have the first artist than no scrobble at all.
I don't think Force track recognition is a good option either for YouTube Music, because YouTube Music can contain videos which could then be wrongly scrobbled.

yayuyokitano:
BREAKING: upgrades required node version to 18
BREAKING: modifies some legacy pipeline behavior, potentially causing inconsisten behavior to v2, in particular with normalize pipeline.

Adds a regex/bulk editing system on top of the existing one, accessed through a "regex mode" button in the edit menu.

Some base rules of how this should work:
- On a song being played, web scrobbler first looks for a normal edit. If it finds one, it applies that, and stops there. No bulk edit will be applied, even if there is a match before or after applying the edit.
- If there is no edit for the particular song, web scrobbler starts to iterate over bulk edits, starting with the most recently made edit by the user, and applies only the most recent bulk edit that matches. Applying only the newest is a nice way for the user to be able to apply "overrides". So if you apply a blanket edit to delete starting from the first comma to only scrobble the first artist, you can easily apply a new bulk edit that affirms an artist that has comma in their name.
- A bulk edit is stored as an object with both a search and a replace property, both of which have all four track fields (track, artist, album, albumartist), which can be strings or null.
- Determining if a bulk edit should be applied is done by going through all the fields of the search property and testing against the respective song field. Null fields are automatically true, while non-null fields must match the entire song field (so the actual applied regex is `/^{searchField}$/`). All four fields must be true for the bulk edit to be applied.
- When a bulk edit is applied, it takes each song field where both search and replace are not null, and applies a replace using `/^{searchField}$/` to the replace field. Capture groups work.
- After first bulk edit is applied, web scrobbler searches for metadata, then applies bulk edit again, as there may have been a new album field added because of the metadata fetching. It’s a bit magic but it’s the best way I see.

The rules of this are admittedly a bit magic, but from my experience it is the ruleset that feels the most right to use. We might need to add an explainer somewhere in the extension though.

Adds some more locale stuff. Some of it is trivial for us to add ourselves based on existing properties, but most isn't.

Closes #2588 Closes #3432 Closes #2948 Closes #2943 Closes #2608 Closes #1552 Closes #3069 Closes #3046 Closes #1643 Closes #2157

Also adds two default filters that remove - EP and - Single at the end of album fields, applied by apple music (which is also commonly fetched for other connectors when no album can be fetched from the website itself).

Hence also closes #3186 

Any more issues this closes?

I think this PR needs a bit more love before being pushed so setting it to draft for now, but it's pretty close.


src/core/object/pipeline/pipeline.ts
```diff
@@ -4,13 +4,21 @@
...
 
 export default class Pipeline {
 	private song: Song | null = null;
-	private processors = [Normalize, UserInput, Metadata, CoverArtArchive];
+	private processors = [
+		Normalize,
+		UserInput,
+		RegexEdits,
+		Metadata,
+		RegexEdits, // Run regex edits again, as the regex edit might have caused an album to be found.
```
inverse, line 19: Why do we need to do it 2x vs just applying the regex edit at the end?

yayuyokitano: I'm not 100% sure just naively running entire regexedits pipeline twice is the optimal solution, but we do need some form of regex edit application both before and after for expected behavior I think.

yayuyokitano:
Updated the editing system to now edit each field once.
This is untested for now, and I need to make some more adjustments.

The new system means that any number of edits can be applied before/after the metadata fetching, but once a field has been modified by a regex edit, the same field cannot be modified again (the same field can be the subject of multiple searches, but not replaces)

This limitation carries over between the regex application before and after metadata fetching.

I think this behavior should feel a lot more consistent.

yayuyokitano:
Fixed up casing stuff i shouldve done in the first place.

Also changed up the previews to now show information prior to doing any regex edits instead of showing warning.
Edits from regular edits still show.
Any album that was fetched from the regex edited information will also still display.

yayuyokitano:
Note that this commit was breaking, and as of this commit the branch is incompatible with the previous commit of the branch. Remove and readd extension to browser to fix. This issue will not occur between this branch and V2, it should smoothly migrate.

inverse:
Not sure I am doing it properly but for example if I go to https://youngspice.bandcamp.com/album/basic-couch-formula

Play a song and edit

in artist field replace left `Young Spice` with `Spice` - just for illustration.

It doesn't seem to change anything for me :/ like still plays correctly. Am I doing something wrong?

src/ui/popup/regex.tsx
```diff
@@ -0,0 +1,188 @@
...
+				onInput={(e) =>
+					setSearches((prev) => ({
+						...prev,
+						[type]: e.currentTarget.value || null,
+					}))
+				}
+			/>
+			<Switch
+				fallback={
+					<Close
```
inverse, line 129: Should this unmatched state block saving?

yayuyokitano: I think it should probably just serve as a guide and not block anything personally.

If the user wants to save anyway probably just let them, the extension is pretty clear that the edit should be doing nothing to the current track.

yayuyokitano:
> Not sure I am doing it properly but for example if I go to https://youngspice.bandcamp.com/album/basic-couch-formula
> 
> Play a song and edit
> 
> in artist field replace left `Young Spice` with `Spice` - just for illustration.
> 
> It doesn't seem to change anything for me :/ like still plays correctly. Am I doing something wrong?

Found the issue, it's caused by an error from the track not existing in lastfm. I think maybe quite a lot of the pipeline actually breaks because of this, and it was only noticeable now because of the regex edits. Let me have a look into it.

yayuyokitano:
My suspicion was correct, all this time the normalize pipeline has literally not been running at all 🤦🏼‍♀️
Pretty sure this should be an issue in v2 too.
Let me try to patchwork this.

yayuyokitano:
Alright, added a new pipeline step at the very top that copies parsed data into processed before starting.

Some steps rely on processed data being there even if it is the same as parsed, including the regex edits.

I don't see anything in the code that should break from this change - with that said I don't guarantee anything, this should be viewed pretty critically as it's a relatively significant change to how the pipeline operates.

Anyway the bug you described should be fixed now (and the normalize pipeline should actually work now! On the negative side that might cause some differences in scrobbled tracks for users between v2 and v3, but it would be v3 working correctly)

This PR should now be considered breaking compared to v2, though other than normalization potentially messing things up a bit I don't think it should be noticeable for users.

src/core/object/pipeline/pipeline.ts
yayuyokitano: Gonna mark this one as resolved after the regex pipeline edits from "Apply edits once per field"

yayuyokitano:
Added tests now.

This uses the node fetch API, meaning required node version is bumped to node 18. It's LTS so I think it's fine.

Now that there are tests. I think I'm ok with undrafting this, though more tests might be needed.

inverse:
Nice improvements! It's def working for me now.

However I noticed this behaviour.

Play music and edit regex rule -> all works

How does one remove this rule? I tried changing it to match and replace with $1 which the preview it works but it seems to only scrobble the first change? 

Perhaps do you think we need some UI to see regex rules to remove them? like with edited tracks (just noticed we can't remove them either here too)

yayuyokitano:
> Nice improvements! It's def working for me now.
> 
> However I noticed this behaviour.
> 
> Play music and edit regex rule -> all works
> 
> How does one remove this rule? I tried changing it to match and replace with $1 which the preview it works but it seems to only scrobble the first change?
> 
> Perhaps do you think we need some UI to see regex rules to remove them? like with edited tracks (just noticed we can't remove them either here too)

The way it currently works is that there are four flags for each of the four fields indicating if they have been edited by a regex edit. Once a regex edit is applied to a field (So that an actual replace operation occurs, this would include just a $1 replace), that field can no longer be modified by any other regex edit. This allows you to use regex edits to override other regex edits (such as having a regex edit changing last comma to ampersand, and then having a separate rule telling it to be ignored for a specific artist with comma in the name). The most recently added regex edit will be applied first.

The full list of regex edits can be viewed and deleted in options menu, just like normal song edits. Should I add a link to the popup linking to the section of options? I don't see an elegant way of adding regex edit deletion within the popup itself.

inverse:
> The full list of regex edits can be viewed and deleted in options menu, just like normal song edits. Should I add a link to the popup linking to the section of options? I don't see an elegant way of adding regex edit deletion within the popup itself.

How did I miss that UI section in options :see_no_evil:  i think it's clear enough for now :+1:

yayuyokitano:
> > The full list of regex edits can be viewed and deleted in options menu, just like normal song edits. Should I add a link to the popup linking to the section of options? I don't see an elegant way of adding regex edit deletion within the popup itself.
> 
> How did I miss that UI section in options 🙈 i think it's clear enough for now 👍

I think it will become clearer once there is a separate entry in the sidebar for edits, which I want to do.

I want this to be merged first though in the interest of trying to keep the PR at least _somewhat_ remotely atomic.

src/core/object/pipeline/regex-edits.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export async function process(song: Song): Promise<void> {
+	// If song info was already corrected by user, skip this stage
+	if (song.flags.isCorrectedByUser) {
+		return;
+	}
+	try {
+		await RegexEdits.loadSongInfo(song);
+	} catch (e) {
+		// Do nothing
```
inverse, line 16: Should we log debug here to make debugging easier if something really went wrong?

inverse (approved):
Aside the comments seems reasonable :+1:

src/core/object/pipeline/regex-edits.ts
yayuyokitano: Not sure. To be honest I dont know how we would even get an error here.

But this pipeline is directly copied from pipeline/user-input, as the user-input function works basically the same due to regex edit class being directly based off the user edit model.

I assume the try catch is there for some reason, but there was no comment in the original telling me why, so I mindlessly copy pasted.

We could add some logging for sure, it shouldnt spam the user.

I’ll merge this as is for now and I can make a new PR adding some logging later, as due to the size of the changes on this PR this not being merged is blocking several future modifications needed before release.

RAM237:
Hey, awesome news! Which version it is supposed to be released with?  
As I understand it is not yet in 2.89.0

yayuyokitano:
It is scheduled for release in version 3.0.0, we’re working on getting it out now

maxelo:
love the new web scrobbler design, thanks for your work
I was very excited for adding regex rules
But I can't  get it to work
Can anyone help me?

help

search: "(\s?\(con\s)"

replace "(feat. "

example:

[https://regex101.com/r/wEwpHH](https://regex101.com/r/wEwpHH)

yayuyokitano:
> love the new web scrobbler design, thanks for your work I was very excited for adding regex rules But I can't get it to work Can anyone help me?
> 
> help
> 
> search: "(\s?(con\s)"
> 
> replace "(feat. "
> 
> example:
> 
> https://regex101.com/r/wEwpHH

Currently, regex demands full text matches. I plan to add some options that will allow you to tweak this behavior a bit and help make the behavior more clear to the user in the future.
So in this case, you might want something along the lines of search: `(.*)\(con\s(.*)` replace: `$1(feat. $2`

maxelo:
> > love the new web scrobbler design, thanks for your work I was very excited for adding regex rules But I can't get it to work Can anyone help me?
> > help
> > search: "(\s?(con\s)"
> > replace "(feat. "
> > example:
> > https://regex101.com/r/wEwpHH
> 
> Currently, regex demands full text matches. I plan to add some options that will allow you to tweak this behavior a bit and help make the behavior more clear to the user in the future. So in this case, you might want something along the lines of search: `(.*)\(con\s(.*)` replace: `$1(feat. $2`

it works =)

hope you can put this rules in the future, thanks for your work!

[https://github.com/web-scrobbler/web-scrobbler/issues/3069#issuecomment-1013927834](url)

maxelo:
I managed to improve the regex rule if you want to add it in the future:

[https://regex101.com/r/rz997G/1](url)

Regular Expression
`(.*)\(con\s(.*)()()(.*)\sy\s(.*)`
Substitution
`$1(feat. $2 and $6`

Example:
Get Lucky (**con** Pharrell Williams **y** Nile Rodgers)
Into:
Get Lucky (**feat.** Pharrell Williams **and** Nile Rodgers)

thanks for your work

PR #3689, opened 2023-05-02
Fix Connector API handling of undefined/null values

yayuyokitano:
Some functions in connector do not handle undefined/null values correctly.
This goes a long way to fixing that, and adjusts connectors accordingly.

I think once these three current PRs are merged, I want to do one last PR changing the sidebar organization in the options (I think there will be a lot of merge conflicts if this is done before current PRs are merged and I can't be bothered) and then I think we might be ready to release!

### DIFF
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -308,7 +308,7 @@ export default class BaseConnector {
 	 * @returns Check result
 	 */
 	public isTrackArtDefault: (
-		trackArtUrl?: string
+		trackArtUrl?: string | null | undefined
 	) => boolean | null | undefined = () => false;
 
 	/**
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -330,7 +330,7 @@ export default class BaseConnector {
 	 *
 	 * @returns True if state change is allowed; false otherwise
 	 */
-	public isScrobblingAllowed: () => boolean | null = () => true;
+	public isScrobblingAllowed: () => boolean | null | undefined = () => true;
 
 	/**
 	 * Function that will be called when the connector is injected and
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -153,7 +153,7 @@ export default class BaseConnector {
 	 *
 	 * @returns Song artist
 	 */
-	public getArtist: () => string | null;
+	public getArtist: () => string | null | undefined;
 
 	/**
 	 * Default implementation of track name lookup by selector.
@@ -162,7 +162,7 @@ export default class BaseConnector {
 	 *
 	 * @returns Song title
 	 */
-	public getTrack: () => string | null;
+	public getTrack: () => string | null | undefined;
 
 	/**
 	 * Default implementation of album name lookup by selector.
@@ -171,7 +171,7 @@ export default class BaseConnector {
 	 *
 	 * @returns Song album
 	 */
-	public getAlbum: () => string | null;
+	public getAlbum: () => string | null | undefined;
 
 	/**
 	 * Default implementation of album artist name lookup by selector.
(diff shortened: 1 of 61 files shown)


src/core/content/util.ts
```diff
@@ -73,7 +73,7 @@ export const defaultSeparators = [
  * @param str - Time-string in h:m:s format
  * @returns Seconds
  */
-export function stringToSeconds(str: string | null): number {
+export function stringToSeconds(str?: string | null): number {
```
inverse, line 76: Isn't `str?` flagging the parameter as optional? In what instance would we not pass either string or null here?

src/core/content/connector.ts
```diff
@@ -462,7 +464,7 @@ export default class BaseConnector {
 	 *
 	 * @returns The source URL
 	 */
-	public getOriginUrl: () => string | null = () => {
+	public getOriginUrl: () => string | null | undefined = () => {
```
inverse, line 467: When would `document.location.href` return anything other than string?

yayuyokitano: getOriginUrl is overrideable, and several connectors override it. document.location.href is always a string.

src/core/content/util.ts
yayuyokitano: You’re completely right, I’ll get rid of this.

src/core/content/connector.ts
inverse: Oh true! Need more :coffee:

src/core/content/connector.ts
```diff
@@ -307,7 +307,9 @@ export default class BaseConnector {
 	 * @param trackArtUrl - Track art URL
 	 * @returns Check result
 	 */
-	public isTrackArtDefault: (trackArtUrl?: string) => boolean = () => false;
+	public isTrackArtDefault: (
```
inverse, line 310: Do we really want to do this? Feels like this function should always return bool

src/core/content/connector.ts
```diff
@@ -279,15 +279,15 @@ export default class BaseConnector {
...
+	public isPlaying: () => boolean | null | undefined;
 
 	/**
 	 * Default implementation to check whether a podcast is playing. Only has an
 	 * effect if the user has opted to disable podcast scrobbling.
 	 *
 	 * @returns True if the current track is a podcast; false otherwise
 	 */
-	public isPodcast: () => boolean = () => false;
+	public isPodcast: () => boolean | null = () => false;
```
inverse, line 290: While I get the selector ones could potentially return undefined should we really be allowing these `is()` ones that seem boolean return anything other than boolean values? feels like we should strive to have a strict interface here to make our implementation a lot niceer

yayuyokitano: The philosophy of web scrobbler v2 was that the connector functions all accepted undefined and null values in order to remove any checking functionality from connectors and put it into the core. This was not just an accident, but not doing null checks is explicitly stated as good practice in our wiki.

The core already handles these cases as v2 did (interpreting nullish value as false), and having a strict implementation vs a lenient implementation actually makes no difference to the core code itself, other than the declared type.

There are quite a few cases where these changes do remove a nullish coalescing operation from connector code (especially with isTrackArtDefault).

We could move to make it strict, but it would be a departure from the way the API has worked in the past and would probably just cause complexity without any big benefit except the core feeling cleaner.

inverse: Gotcha - yeah it does make the connector implementation feel cleaner without having to be explicit.

inverse (approved):
LGTM although the interface feels weird but your points are totally valid :)

yayuyokitano:
Yeah it's a little odd – why i didn't do this in the initial v3 implementation.

btw I wonder if we're good to release now. I can't think of anything more at this point (though im sure there are some bugs hiding)

inverse:
Feels like we're in a very good shape with a lot of nice additions from you ;)

PR #3692, opened 2023-05-02
dont output "expected a number or a string"

inverse:

Seems to work just fine without this when testing  - should we even be logging this? Feels like it's causing a bit of noise in the console of each page? 

**Additional context**


yayuyokitano:
Yep, I think I talked about wanting to remove this.

I’ll have to have a closer look at the v2 code before approving, as I think this is an important function to get right

yayuyokitano (approved):
LGTM

PR #3693, opened 2023-05-03
Reimplement tests

yayuyokitano:
Some tests were removed in v3 due to the complexity of reimplementing them.
#3671 removed a fair amount of that complexity, so I went back and reimplemented all of it.

Some of this is VERY cursed because these tests were not written with typescript in mind, but they seem to work.


inverse (approved):
Nice one!

PR #3694, opened 2023-05-03
Reorganize options

yayuyokitano:
Splits the old options tab into three different tabs, options, edits, and connectors.
When even one of our main maintainers can't find the edits section (https://github.com/web-scrobbler/web-scrobbler/pull/3671#issuecomment-1529133733) we have a UX problem.

image

Also makes the sidebar scrollable, as at this point that could become relevant with some small window sizes.
Most actually used window sizes should still not experience having to scroll the sidebar though.


inverse (approved):
So much nicer!

PR #3698, opened 2023-05-05
Fix edit import/exports

yayuyokitano:
- Changes the export method to be compatible with safari.
- Moves to using refs for imports, fixing an accidental ID-related bug in the process.
- Live updates the content of the modals to show imported edits without refreshing the page.


inverse (commented):
LGTM :+1:

inverse (approved):

PR #3700, opened 2023-05-05
Add links to now playing popup

yayuyokitano:
Adds links to the now playing popup for track, artist, album, album artist, playcount (leading to track in users library).

Also does a bit of general split up of the now playing component.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -205,3 +205,69 @@ export function areAllResults<T>(results: T[], result: T): boolean {
 export function capitalizeFirstLetter(text: string): string {
 	return text[0].toUpperCase() + text.slice(1);
 }
+
+/**
+ * Create a URL to an artist page on Last.fm.
+ * @param artist - Artist name
+ * @returns URL to the artist page
+ */
+export function createArtistURL(artist: string | null | undefined): string {
+	if (!artist) {
+		return '';
+	}
+	return `https://www.last.fm/music/${encodeURIComponent(artist)}`;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


inverse (approved):
LGTM - nice one!

src/util/util.ts
```diff
@@ -205,3 +205,69 @@ export function areAllResults<T>(results: T[], result: T): boolean {
...
+/**
+ * Create a URL to an artist page on Last.fm.
+ * @param artist - Artist name
+ * @returns URL to the artist page
+ */
+export function createArtistURL(artist: string | null | undefined): string {
+	if (!artist) {
+		return '';
+	}
+	return `https://www.last.fm/music/${encodeURIComponent(artist)}`;
```
inverse, line 218: I'm guessing we don't need to handle localisation here as lastfm should redirect?

yayuyokitano: ah thats a good spot actually. No lastfm does not redirect.

yayuyokitano: Looking at v2 - it seems web scrobbler never localized urls.
Since the main purpose here is to restore functionality i think localization of URLs should be a separate PR.
Did some simplification regardless, and fixed one pretty huge bug i found that i introduced in #3671 that caused applying an edit to not actually exit the edit state.

yayuyokitano:
Also just to make it clear, album and album artist are not grabbed from metadata because it is unreliable method of getting link for those fields. Track and artist are reliable.

yayuyokitano:
Reverted the link change – reverting this allows us to change the baseurl based on localization more easily later

PR #3714, opened 2023-05-08
Make options page more responsive, improve building

yayuyokitano:
Implements a bottom context menu in place of sidebar on low widths for better experience. Still not 100% ideal but pretty close.

Example:
image

Also adds a dev mode to safari build.

Also adds new build target as discussed.

### DIFF
--- a/build.ts
+++ b/build.ts
@@ -1,7 +1,7 @@
 import { build } from 'vite';
 import * as configs from './vite.configs';
 import { mkdir } from 'fs/promises';
-import { getBrowser, releaseTarget, releaseType } from './scripts/util';
+import { getBrowser, isDev, releaseTarget, releaseType } from './scripts/util';
 import colorLog from 'scripts/log';
 import { exec } from 'child_process';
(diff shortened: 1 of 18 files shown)


build.ts
```diff
@@ -40,20 +40,32 @@ async function main() {
 	];
 	await Promise.all(scripts);
 
+	if (isDev() && releaseTarget === 'safari') {
+		await new Promise((resolve) => {
```
inverse, line 44: Why do we need to have this sleep?

inverse (approved):
LGTM - nice one :+1:

build.ts
yayuyokitano: The dev build does not await the completion of the build before continuing, the sleep allows the initial build processes to finish up before creating the xcode project.

There is probably a better way here long term, but this did nicely for development on my end for the time being.

inverse: Ahhh is that because of the watch mode? Yeah we should probably think of a better way longer term - but right now it'll do :)

PR #3725, opened 2023-05-15
Make popup responsive

yayuyokitano:
This is kinda patchworky as some major redesigning had to be done, but from my testing it seems to work without issue, without affecting desktop.

Note that this adds _a lot_ of new items to locale, as existing entries text are too long to be practical for the buttons.


inverse (approved):
LGTM

PR #3752, opened 2023-05-25
fix scrobble not working when play a playlist

youthlin:

Since there is a replayDetectionTimer, So when the song played end, it would be set to replay, and call the `pause()` method. So that `this.pausedOn` will not be null.
Then will play the next song on the playlist, and call the `start()` method. However, when `update()` called, `this.pausedOn == null` is false, so `this.setTrigger` is not called. This make the 2nd and the rest in the playlist would NOT be scrobbled.

So when `start`  called, we should set `this.pausedOn` to `null`~

**Additional context**

### DIFF
--- a/src/core/object/timer.ts
+++ b/src/core/object/timer.ts
@@ -30,6 +30,7 @@ export default class Timer {
 	start(cb:()=>void):void {
 		this.reset();
 		this.startedOn = now();
+		this.pausedOn = null;
 		this.callback = cb;
 	}
 


yayuyokitano:
Normally the connector should not be pausing as it goes to a new song so I think this bug might only be happening with a handful of connectors.

This fix should make the core more robust, and I don’t see any situation where it can break anything, so LGTM, but I want another reviewer to look at it too in case I missed some obscure gotcha.

youthlin:
> Normally the connector should not be pausing as it goes to a new song so I think this bug might only be happening with a handful of connectors.
> 
> This fix should make the core more robust, and I don’t see any situation where it can break anything, so LGTM, but I want another reviewer to look at it too in case I missed some obscure gotcha.

Yes, the 163-music connector watch the `a[data-action="play"]` element, when the played song changed, the button would changed from `data-action="pause"`(when play) to `data-action="play"`(when switch a song) and to `data-action="pause"`(when the new song played)
image

inverse (approved):
Can't think of any edge cases :/

PR #3757, opened 2023-05-25
Conditionally render different options in safari

yayuyokitano:
Hopefully a decent compromise with apple's policies.
Release ready with this merge methinks.

Introduces ifdef to the repo, and an extension to highlight them.
It is worth noting that the conditional compile plugin used is quite small and potentially unstable, but has worked just fine for me https://github.com/KeJunMao/vite-plugin-conditional-compile


inverse (approved):
Not the biggest fan of this as it adds conditional behaviour which will be harder to dev with but I see your reasoning :)

inverse (approved):
Not the biggest fan of this as it adds conditional behaviour which will be harder to dev with but I see your reasoning :)

inverse (approved):
Not the biggest fan of this as it adds conditional behaviour which will be harder to dev with but I see your reasoning :)

PR #3759, opened 2023-05-26
Update logos

yayuyokitano:
EDIT: Found desired sizes from chrome web store, adjusted universal icon accordingly

Also changes safari icon to not have weird clippings on ios


yayuyokitano:
Ok now it should actually be good

inverse (approved):

PR #3776, opened 2023-05-27
Add tooltips for common errors

yayuyokitano:
Adds a tooltip for service error informing about adblock things
Adds a tooltip for unsupported site informing about needing to refresh after update

This is definitely not optimal and we should make better long term fixes, but I think it's an ok temporary bandaid while we focus on more pressing issues


inverse:
Feels like a nice work around for now

PR #3797, opened 2023-05-28
Don't fetch edits on unsupported sites

yayuyokitano:
Should reduce memory usage a lot for people with many tabs and many saved edits.

EDIT: also checks for frames when allframes is false now, shouldve been added a long time ago

### DIFF
--- a/src/core/content/main.ts
+++ b/src/core/content/main.ts
@@ -27,7 +27,13 @@ async function main() {
  */
 async function fetchConnector(): Promise<void> {
 	const connector = await getConnectorByUrl(window.location.href);
+	console.log('bbbbb', connector, window.location.href);
 	if (!connector) {
+		return;
+	}
+
+	// Don't run the connector in frames if it's not allowed to run in frames
+	if (window !== top && !connector.allFrames) {
 		return;
 	}
 
--- a/src/core/content/main.ts
+++ b/src/core/content/main.ts
@@ -27,7 +27,6 @@ async function main() {
  */
 async function fetchConnector(): Promise<void> {
 	const connector = await getConnectorByUrl(window.location.href);
-	console.log('bbbbb', connector, window.location.href);
 	if (!connector) {
 		return;
 	}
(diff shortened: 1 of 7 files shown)


src/core/content/main.ts
```diff
@@ -25,15 +27,23 @@ async function main() {
  */
 async function fetchConnector(): Promise<void> {
 	const connector = await getConnectorByUrl(window.location.href);
+	console.log('bbbbb', connector, window.location.href);
```
inverse, line 30:
```suggestion
```

Lets remove this :)

src/core/content/main.ts
```diff
@@ -25,15 +27,23 @@ async function main() {
...
 	if (!connector) {
 		return;
 	}
 
+	// Don't run the connector in frames if it's not allowed to run in frames
+	if (window !== top && !connector.allFrames) {
```
inverse, line 36: What does the first conditional do? Not familiar with this :)

src/core/storage/saved-edits.ts
```diff
@@ -12,8 +12,8 @@ class SavedEditsImpl extends SavedEditsModel {
 		BrowserStorage.LOCAL_CACHE
 	);
 
-	constructor() {
-		super(BrowserStorage.getLocalStorage(BrowserStorage.LOCAL_CACHE));
+	public init() {
+		this._init(BrowserStorage.getLocalStorage(BrowserStorage.LOCAL_CACHE));
```
inverse, line 16: Should we protect if it's already been init?

inverse (approved):

src/core/content/main.ts
yayuyokitano: window === top if content script is not running in a frame/iframe

src/core/storage/saved-edits.ts
yayuyokitano: There shouldn't be any harm to running init twice except that you initialize the storage twice.
Though it is only ever run at the very beginning.
_init is protected.

Issue #3789, opened 2023-05-28
High CPU usage at idle and continuously looping fetch for extensions::setIcon


Infinitay:
**Describe the bug**
After a fresh restart of my PC and Chrome, I immediately noticed my computers fans spinning faster. I took at look at task manager and Chrome was the culprit. Taking a look at Chrome's task manager, The Web Scrobbler extension was using >= 100% of my CPU usage. At the time I opened a new tab and opened the developer console because at first it didn't explicitly say "Extension: Web Scrobbler", it only showed a service worker and pointed to chrome-extension:// with Web Scrobbler's ID (hhinaapppaileiechjoiifaancjggfjm). Immediately I noticed lots of logs to the console pointing to it's storage variable with settings and other data related to the extension. 

Furthermore, when expecting the service worker for the extension, I noticed there was an endless loop of a fetch call within `extensions::setIcon` @L10 within the `loadImagePathForServiceWorker` function.

**How to reproduce**
I'm not sure to be honest. I logged on to my PC for the first time today and relaunched my Chrome session for the previous night

**Expected behavior**
1. No high CPU usage
2. No endless fetch loop

**Screenshots**
![image](https://github.com/web-scrobbler/web-scrobbler/assets/6964154/dc90f86b-1844-48fc-8fe2-d6ba75c3517f)
![image](https://github.com/web-scrobbler/web-scrobbler/assets/6964154/06800a06-034d-4bd2-9307-2b3989b14673)
![image](https://github.com/web-scrobbler/web-scrobbler/assets/6964154/7ace28d4-0b76-455c-99d5-f229fa6cf0ae)

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   OS: Windows 10
-   Browser: Chrome, Version 113.0.5672.127 (Official Build) (64-bit)
-   Extension version: 3.0.5

**Additional context**

**Potentially resolved itself?**
As I was getting my environment data to submit this bug report, the endless loop of fetching the icon and calling of `setIcon` was stopped. This happened immediately after I went on the extensions page and right once I clicked on the "Accounts" tab. The odd thing is that I was already on the extension options page and clicked thru all of the option menus **except** for that Account tab. The moment I did my fans slowed down and when I saw Chrome's task manager the extension's CPU usage dropped.

[...]

PR #3801, opened 2023-05-28
Improve service worker performance with many tabs

yayuyokitano:
Maximum priority issue, should be a hotfix ASAP.

Goes a long way to fixing #3789 and might even close it entirely.
Reduces CPU usage of service worker DRASTICALLY, also reduces RAM usage quite a bit.
V3 service worker now typically uses less RAM than V2 background page did, and hardly any more CPU it seems.

Mostly impacts the code determining which tab to use for setting action icon/page.
Seems stable on my end at least, but limited amounts of testing can be done in this short amount of time.


inverse (approved):
LGTM - awesome work :pray:

PR #3805, opened 2023-05-28
Fix edit delay check

yayuyokitano:
If you opened edit dialog, started editing, then closed the popup it would cause scrobble not to go through.

### DIFF
--- a/src/core/object/controller/controller.ts
+++ b/src/core/object/controller/controller.ts
@@ -130,6 +130,8 @@ export default class Controller {
 				type: 'setEditState',
 				fn: (isEditing) => {
 					this.isEditing = isEditing;
+					this.eventEmitter.emit('updateEditStatus', isEditing);
+
 					if (isEditing) {
 						clearTimeout(this.setNotEditingTimeout);
 						this.setNotEditingTimeout = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/core/object/controller/controller.ts
```diff
@@ -127,6 +130,13 @@ export default class Controller {
 				type: 'setEditState',
 				fn: (isEditing) => {
 					this.isEditing = isEditing;
+					if (isEditing) {
+						clearTimeout(this.setNotEditingTimeout);
+						this.setNotEditingTimeout = setTimeout(() => {
+							this.isEditing = false;
+							this.eventEmitter.emit('updateEditStatus', false);
+						}, 5000);
+					}
```
inverse, line 139: Should it have a `return`?  Otherwise we will call `this.eventEmitter.emit('updateEditStatus', isEditing);` again? or is that not a problem?

inverse, line 139: Shoudl we return here? or is it okays that we call `this.eventEmitter.emit('updateEditStatus', isEditing);`?

inverse (approved):
LGTM aside one question

src/core/object/controller/controller.ts
yayuyokitano: It is not only ok, it should be called. I'll bump the line up to make it a bit clearer

yayuyokitano: Basically the emitter outside the timeout will instantly set controller to not be in edit mode if the user cancels the edit within the popup or finishes the edit, while the timeout takes care of the case where the user just clicks outside the popup while editing

Issue #3812, opened 2023-05-29
[AccuRadio] - Web Scrobbler doesn't stop scrobbling tracks with long titles


Rostov84:
**Describe the bug**

In general, Web Scrobbler properly scrobbles tracks from all Accu Radio channels. But after updating the extension to version 3.0.9
for tracks from channels in the "classical music" category (https://www.accuradio.com/classical/) Web Scrobbler starts to scrobble (green note icon appears), but does not complete the scrobble process (green tick icon does not appear) . I suppose this is due to the song titles being too long. For this reason, such songs are not listed on the last.fm list.

**Environment (please complete the following information)**
-   OS: Windows 7 Home Premium SP1 (x64)
-   Browser: Google Chrome 109.0.5414.120 (x64)
-   Extension version: 3.0.9

yayuyokitano:
Hi, could you give a specific example of a track that does not work? I do know that there is a limit of 500 bytes per field from last.fm's side, idk if you're hitting that though.

Rostov84:
For example:

Amadeus Quartet - String Quartet in D major, KV 499 "Hoffmeister": 4th movement

inverse:
This connector currently doesnt handle the duration of tracks. Could it be something to do with this not existing causing scrobble issues?

yayuyokitano:
Very possible. The name sure isn't long enough to be causing any issues with lastfm

inverse:
I just patched the connector to pick up this information and it scrobbles correctly. 

We should def handle this case - wonder what changed compared to v2 for this :/

yayuyokitano:
Think I found the issue, some excessively aggressive type checking

Rostov84:
I just noticed that the same problem also applies to stations:

https://www.di.fm/
https://www.rockradio.com/
https://www.jazzradio.com/
https://www.classicalradio.com/
https://www.radiotunes.com/
https://www.zenradio.com/

However, while on classic Accu Radio channels the problem applies to all songs, on the above-mentioned channels scrobbling of only selected songs is not completed.

yayuyokitano:
I suspect these might all be fixed by #3815

inverse:
Published a release just now - it's pending review from Google

PR #3814, opened 2023-05-29
Handle duration in accuradio

inverse:
Fix for #3812

But we should handle missing duration information :)

### DIFF
--- a/src/connectors/accuradio.ts
+++ b/src/connectors/accuradio.ts
@@ -20,3 +20,30 @@ function getScrollableText(selector: string) {
 		Util.getTextFromSelectors(selector)
 	);
 }
+
+function getDuration() {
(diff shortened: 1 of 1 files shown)


src/connectors/accuradio.ts
```diff
@@ -20,3 +20,30 @@ function getScrollableText(selector: string) {
 		Util.getTextFromSelectors(selector)
 	);
 }
+
+function getDurationInfo() {
```
yayuyokitano, line 24: Are you sure you cant just use Connector.timeInfoSelector = "#progressWrapper" here?

inverse: How did I forget about this - :see_no_evil:

inverse: I thought we had something like this!

PR #3815, opened 2023-05-29
Fix handling of tracks without duration

yayuyokitano:
Loosens some typechecks to reimplement defaults

### DIFF
--- a/src/core/object/timer.ts
+++ b/src/core/object/timer.ts
@@ -2,7 +2,7 @@
  * Timer object.
  */
 export default class Timer {
-	private targetSeconds: number | null = null;
+	private targetSeconds: number | null | undefined = null;
 	private startedOn: number | null = null;
 	private pausedOn: number | null = null;
 	private spentPaused = 0;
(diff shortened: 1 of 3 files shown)


src/core/object/timer.ts
```diff
@@ -78,12 +82,15 @@ export default class Timer {
...
 	 * @param seconds - Seconds
 	 */
-	update(seconds: number | null): void {
+	update(seconds: number | null | undefined): void {
 		// only if timer was started
 		if (this.startedOn !== null) {
 			this.targetSeconds = seconds;
 
-			if (this.targetSeconds !== null) {
+			if (
```
inverse, line 90: won't this change mean that when there is no duration we would just fall into clearTrigger?

yayuyokitano: Yes. However, crucially, by this point getSecondsToScrobble has been called.
getSecondsToScrobble sets the seconds to default scrobble time if it is null.
So really, it should be impossible for this to actually be null, just typescript type checking combined with javascript jank moment.

inverse (approved):

inverse (approved):

PR #3816, opened 2023-05-29
Fix Apple Music connector

### LINKED ISSUE #3808: Apple Music connector not working
Have never GitHub before so this is a first -

**Describe the bug**
Web Scrobbler won't recognise tracks via Apple Music, tracks not appearing/notifying when playing, not being logged on LastFM account.

**Expected behavior**
Tracks recognised and scrobbled, notification to pop up with song playing. 

-   OS: Windows
-   Browser: Chrome
-   Extension version: 113.0.5672.127

image

yayuyokitano:
Seems like the old musickit script didnt handle initialization correctly – sometimes MusicKit can be initialized without the instance being initialized yet.
This introduces a system that polls for it until it is actually initialized.
It seems to consistently work now where the old system failed.

Closes #3808

### DIFF
--- a/src/connectors/musickit-dom-inject.ts
+++ b/src/connectors/musickit-dom-inject.ts
@@ -3,55 +3,70 @@
  * global. It listens to events directly from the player and propagates these
  * events to the extension via 'postMessage'.
  */
-if ('MusicKit' in window && window.MusicKit) {
-	musicKitAddEventListeners();
-} else {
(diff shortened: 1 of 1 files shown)


inverse (approved):
Can't verify but changes seem legit :+1:

inverse (approved):
Can't verify but changes seem legit :+1:

PR #3817, opened 2023-05-29
Reimplement ifdefs

yayuyokitano:
Reduces the amount of logs considerably. Also potentially avoids some future firefox bugs.


inverse (approved):
Nice one!

PR #3820, opened 2023-05-30
Push content scripts on reload

### LINKED ISSUE #3551: Adding zvuk.com
Is it possible to add connector for https://zvuk.com?

Now the extension says that the site is not supported although "СберЗвук" is advertised as available in the desription of Web Scrobbler.

"СберЗвук" is the previous name for zvuk.com. They also changed the domain.

yayuyokitano:
This makes it so that whenever the extension reloads/is updated, it pushes content script into all pages that have connectors.
This prevents the user from having to refresh the page after every update.
~~There is some risk of double scrobbles occurring the first time this is pushed since previously scrobbles were pushed from the content script, but it seems like (unintentionally) this already errored before, causing scrobbles to not be submitted after an update, so I don't think it should be an issue.~~ EDIT 2: Not an issue anymore

~~Note: this seems to occasionally fail as the PR currently stands, with connector not being properly loaded in. I am not sure what is causing this, but it seems to only affect loading on extension reload, and not loading in normal use. Still worrying though.
Happens approximately 1/4 - 1/10 reloads for me depending on connector. Seems to happen the most on eggs.mu out of the websites I tested.~~ EDIT 2: Seems to be fixed

Also note: this would require us to add a new permission to the extension. This will cause a popup on the user side I think.

EDIT: Closes #3551 

EDIT 3: Eggs connector still seems to be buggy after reload especially on firefox, but it does scrobble.


yayuyokitano:
Closes #3551

src/core/background/inject.ts
```diff
@@ -0,0 +1,75 @@
...
+	 * As scripts are always invalidated on reload, and this only runs on install, there is no need.
+	 */
+
+	try {
+		const script = 'content/main.js';
+		browser.scripting.executeScript({
+			target: { tabId },
+			files: [script],
+		});
+	} catch (err) {
```
inverse, line 53: Do we even care for catching? What about letting it be caught in the outer loop?

inverse (approved):
This is magic!

Did some minimal testing in Chrome with Bandcamp and it worked flawlessly!


Might be worth getting @jaccarmac taking a look too as there is quite a bit of change going on here.

src/core/background/inject.ts
```diff
@@ -0,0 +1,75 @@
...
+		if (err instanceof Error) {
```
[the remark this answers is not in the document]
yayuyokitano: Because try catch is not typed (a function could theoretically throw anything and typescript isn't that smart), typescript requires us to validate it.

src/core/background/inject.ts
yayuyokitano: It should be fine to just not catch. I think I copy pasted this specific part from V2 and modified it for working with manifest V3 and web scrobbler V3 (which is much simpler). There might have been additional logic here that has simply been stripped. Removing.

lbrueckner:
https://github.com/web-scrobbler/web-scrobbler/blob/9c7ec746eaaa57a90a106488d01e0ae257336888/src/connectors/deezer-dom-inject.ts#L62

This causes an error `Uncaught TypeError: currentMedia is null` when loading the extension initially because at that point window.dzPlayer.getCurrentSong() returns null.

This additional checks prevent the error while initializing on reload still seem to work:
```
modified   src/connectors/deezer-dom-inject.ts
@@ -123,7 +123,7 @@ if ('cleanup' in window && typeof window.cleanup === 'function') {
 		const currentMedia = player.getCurrentSong();
 
 		// Radio stations don't provide track info
-		if (currentMedia.EXTERNAL) {
+		if (currentMedia === null  ||  currentMedia.EXTERNAL) {
 			return null;
 		}
 
@@ -197,6 +197,9 @@ if ('cleanup' in window && typeof window.cleanup === 'function') {
 		const currentMedia = (
 			window.dzPlayer as { getCurrentSong: () => Media }
 		).getCurrentSong();
+                if (currentMedia === null) {
+                        return null;
+                }
 		return currentMedia.__TYPE__ === 'episode';
 	}

```

PR #3825, opened 2023-05-31
Tidy listenbrainz scrobbler

inverse:

- Remove unused code
- Make FF case conditional (there was a console  error on chrome when I wasn't logged in)

**Additional context**

### DIFF
--- a/src/core/scrobbler/listenbrainz/listenbrainz-scrobbler.ts
+++ b/src/core/scrobbler/listenbrainz/listenbrainz-scrobbler.ts
@@ -335,7 +335,7 @@ export default class ListenBrainzScrobbler extends BaseScrobbler<'ListenBrainz'>
 		// NOTE: Use 'same-origin' credentials to fix login on Firefox ESR 60.
 		const promise = fetch(url, {
 			method: 'GET',
-			// #v-ifndef VITE_FIREFOX
+			// #v-ifdef VITE_FIREFOX
 			credentials: 'same-origin',
 			// #v-endif
 		});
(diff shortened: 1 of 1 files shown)


src/core/scrobbler/listenbrainz/listenbrainz-scrobbler.ts
```diff
@@ -21,8 +21,6 @@ const apiUrl = `${baseUrl}/submit-listens`;
 export default class ListenBrainzScrobbler extends BaseScrobbler<'ListenBrainz'> {
 	public userApiUrl!: string;
 	public userToken!: string;
-	/** Can't find where this value is being set from */
-	private authUrl!: string;
```
inverse, line 25: I think it was changed in https://github.com/web-scrobbler/web-scrobbler/commit/9f885e4c4d31d03e64b497b68f9b3a3e901ea1e2#diff-e2b6cac4e0d320435974b23d0ebc5174fb22df607c8def762f90dd7eb1004830L272 and never cleaned up

src/core/scrobbler/listenbrainz/listenbrainz-scrobbler.ts
```diff
@@ -346,7 +335,9 @@ export default class ListenBrainzScrobbler extends BaseScrobbler<'ListenBrainz'>
 		// NOTE: Use 'same-origin' credentials to fix login on Firefox ESR 60.
 		const promise = fetch(url, {
 			method: 'GET',
+			// #v-ifndef VITE_FIREFOX
```
yayuyokitano, line 338: ifndef is for every browser except for firefox, i think you mightve meant ifdef?

inverse: Good catch :see_no_evil:

yayuyokitano:
Still works on my end and cant see how its used so LGTM

PR #3827, opened 2023-05-31
Update deploy.yml

inverse:
New addon was published with debug flag to get more insights into what failed https://github.com/inverse/edge-addon/releases/tag/v1.0.5


yayuyokitano (approved):
Since its not using core package i guess this is the cleanest solution

PR #3829, opened 2023-05-31
Reduce and prefix logging

### LINKED ISSUE #3828: Extension polluting console on every website
**Describe the bug**
The extension is writing to console.log for every website - most users probably won't notice but I am a web developer and this hinders me.

**How to reproduce**
1. Open website
2. Open console...

**Expected behavior**
Don't pollute console

-   OS: W10
-   Browser: Vivaldi (latest)
-   Extension version: latest

yayuyokitano:
Small patch that more or less completes the process of reducing our logging verbosity and prefixing logs.
Does modify the loading behavior of the content script, but shouldn't be in a meaningful way (other than for reducing logs).
Completely eliminates all logs on non-supported websites, hence closes #3828

### DIFF
--- a/src/core/content/main.ts
+++ b/src/core/content/main.ts
@@ -14,11 +14,14 @@ async function main() {
 	updateTheme();
 	try {
 		await fetchConnector();
+		start();
 	} catch (err) {
+		if (err instanceof Error && err.message === 'dontlog') {
+			return;
+		}
 		Util.debugLog(err, 'error');
 		return;
 	}
-	start();
 }
 
 /**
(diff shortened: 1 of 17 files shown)


src/core/content/main.ts
```diff
@@ -14,11 +14,14 @@ async function main() {
 	updateTheme();
 	try {
 		await fetchConnector();
+		start();
 	} catch (err) {
+		if (err instanceof Error && err.message === 'dontlog') {
```
inverse, line 19: When would it not be an `Error`?

inverse (approved):

PR #3838, opened 2023-06-03
Fix throttle

### LINKED ISSUE #3811: Deezer scrobbling not recognizing the second track in queue and related issues
**Describe the bug**
If there are say 3 tracks in the queue, first track usually gets scrobbled normally, but web-scrobbler doesn't recognize the second track (sometimes that causes a duplicated scrobble for the first track). When the third track starts, it thinks it's the second track. It's usually scrobbling one track behind, after double scrobbling the first. That sometimes also causes scrobbles to be missed (depending on song lengths). Pausing a song and then clicking play before the 50% mark usually helps the scrobbler recognize that song.

**How to reproduce**
1. Go to 'deezer.com'
2. Click on 'any playlist or album'
3. Wait for 3 songs to play and see which ones got scrobbled

**Expected behavior**
All songs should be scrobbled at their own 50% mark, without delayed, duplicate or missed scrobbles.

OS: MacOS Catalina 10.15.7
Browser: Google Chrome Version 113.0.5672.126 (Official Build)
Extension version: [v3.0.9](https://github.com/web-scrobbler/web-scrobbler/releases/tag/v3.0.9)

yayuyokitano:
Fixes an issue where the throttled function would not actually trigger after the throttle period when throttler was called during throttled time.
Closes #3811

### DIFF
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -287,12 +287,22 @@ export function extractUrlFromCssProperty(
  */
 export function throttle<T>(func: () => T, wait: number): () => T | undefined {
 	let prev = 0;
+	let timeout = setTimeout(() => {
+		// do nothing
+	}, 0);
+
 	return () => {
 		const now = Date.now();
-		if (now - prev > wait) {
+		const remaining = wait - (now - prev);
+		if (remaining <= 0) {
 			prev = now;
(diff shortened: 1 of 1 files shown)


inverse (approved):
Nice one

PR #3856, opened 2023-06-03
Fix image url helper

### LINKED ISSUE #3769: Sirius XM won't scrobble in Google Chrome. Windows 10.
**Describe the bug**

**How to reproduce**

**Expected behavior**

**Screenshots**

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

-   OS: (e.g. Windows, Mac)
-   Browser: (e.g. Chrome, Safari)
-   Extension version: (e.g. v2.10.0)

**Additional context**

yayuyokitano:
Checks off a little edge case that was missed.
Closes #3769 as this was the underlying cause of that connector not working.

### DIFF
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -496,7 +496,7 @@ export function extractImageUrlFromSelectors(
 		return null;
 	}
 	const elements = queryElements(selectors);
-	if (!elements) {
+	if (!elements || !elements.length) {
 		return null;
 	}
 


src/core/content/util.ts
```diff
@@ -496,7 +496,7 @@ export function extractImageUrlFromSelectors(
 		return null;
 	}
 	const elements = queryElements(selectors);
-	if (!elements) {
+	if (!elements || !elements.length) {
```
inverse, line 499: Still learning this TS syntax - just to something more akin to `elements.length === 0

inverse (approved):

src/core/content/util.ts
yayuyokitano: Did it out of habit. It's essentially equivalent to elements.length === 0.
Its a bit of a more javascripty way of doing things, as if _somehow_ (javascript) we end up not actually having an array here it will still short circuit and not error.
I don't think it should be possible for there to not be an array at this point though

PR #3857, opened 2023-06-03
Remove minification

yayuyokitano:
Not convinced this is worth it.
Would take us from ~640kb to ~690kb zipped, and from ~1.3mb to ~1.7mb unpacked.
However, also allows gleaning a lot more information from production logs.

### DIFF
--- a/vite.configs.ts
+++ b/vite.configs.ts
@@ -38,7 +38,7 @@ function distRoot() {
 export const buildBackground: UserConfig = {
 	...common,
 	build: {
-		minify: !isDev(),
+		minify: false,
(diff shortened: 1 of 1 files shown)


inverse (approved):
Seems legit to me - so you're saying it'll give us better stacktraces when things go boom? Do we need sourcemaps then?

yayuyokitano:
Ah actually i just noticed we are not making source maps at all, might be able to solve the same issue by adding those. I'll see how it works out later.

yayuyokitano:
Gonna close this to solve another way in future

PR #3858, opened 2023-06-03
Simplify priority

yayuyokitano:
It seemed good on paper to have errors take precedence over other prioritized modes.
In practice it seems to cause more confusion than it helps.


inverse (approved):
LGTM

inverse (approved):
LGTM

PR #3860, opened 2023-06-04
Pull new translations

inverse:

I noticed we were not pulling down new translations so I ran `tx pull --all`

Looks like some are not translated, but that shouldn't impact current flow as I see.

**Additional context**


yayuyokitano (approved):
Does increase bundle size a fair bit, but doesn't impact the size of any of the JS files, so it shouldn't be able to impact performance in any way

PR #3865, opened 2023-06-05
Fix manifest

yayuyokitano:
Firefox warninged with the previous one


jaccarmac:
Just checking: Chrome doesn't need that block at all?

PR #3870, opened 2023-06-06
Reimplement commands

### LINKED ISSUE #3868: Keyboard Shortcuts not working anymore
With the new version 3 the keyboard shortcuts are not working anymore. With version 2.x you could "toggle-connector", "disable-connector", "love-song" or "unlove-song" using configurable shortcuts, this now doesn't work anymore.

yayuyokitano:
Missed this in the background rewrite.
Didn't test this super thoroughly but all three are working on my end on initial testing and its not like this should have potential to break anything else.
Closes #3868

### DIFF
--- a/manifest.config.ts
+++ b/manifest.config.ts
@@ -45,6 +45,18 @@ export const common: Manifest.WebExtensionManifest = {
 	},
 
 	action: getAction(releaseTarget),
+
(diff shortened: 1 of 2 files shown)


inverse (approved):
Nice catch

PR #3873, opened 2023-06-07
Add fetch album artist from music kit instance

fikimaul:
Add optional fetch album artist to  track info when available. Album artist will availabe on container attributes when song play from Album or queue as Full Album.

**Additional context**

_Album Attributes_
 
container

_Song Play From Album_

album

_Song Play Not from album (search result or playlist)_

song

### DIFF
--- a/src/connectors/musickit-dom-inject.ts
+++ b/src/connectors/musickit-dom-inject.ts
@@ -63,6 +63,7 @@ async function musicKitGetTrackInfo() {
 		album: item.albumName,
 		track: item.title,
 		artist: item.artistName,
+		albumArtist: item.container?.attributes?.artistName,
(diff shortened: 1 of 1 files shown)


jaccarmac:
Looks good to me! @yayuyokitano, does Typescript optional chaining work for stuff like this since the rewrite?

yayuyokitano:
> Looks good to me! @yayuyokitano, does Typescript optional chaining work for stuff like this since the rewrite?

Sorry a bit late on the ball here.
Optional chaining is no issue at all, it's a raw ecmascript feature at this point.
It just results in nullish value potentially being passed into functions which they are designed to handle.
Been testing it out and LGTM, thank you for your contribution!

PR #3876, opened 2023-06-08
Fix RadioParadise connector

jbwharris:
Cleaned up the targeted IDs and added a filter so it wouldn't scrobble the Listener Supported message


src/connectors/radioparadise.ts
```diff
@@ -2,14 +2,18 @@ export {};
...
-Connector.trackArtSelector = '.now_playing_list.ng-star-inserted:not(.dim) img';
+Connector.trackArtSelector = '#now_playing img.album-cover';
+
+Connector.isScrobblingAllowed = () => {
+	return (
+		!Util.getTextFromSelectors('#now_playing .title')?.includes(
+			'Listener-supported'
+		) && !Connector.getArtist()?.startsWith('Commercial-Free')
+	);
+};
```
inverse, line 19:
```suggestion
};

```

inverse (approved):

PR #3888, opened 2023-06-12
Fix Irama Nusantara connector not working properly

Slickerius:
I updated the connector for Irama Nusantara by omitting the play button selector.

**Additional context**
The previous connector file had a bug that did not allow for playback to be scrobbled on several browsers, including Chrome, due to it tracking the play button selector. Now it is removed to allow all browsers to scrobble the site.

irnus


inverse (commented):
We could also do something like this

```
Connector.isPlaying = () => {
	return (
		Util.getAttrFromSelectors('#player-control button svg path', 'd') ===
		'M9 8v6m4-6v6m7-3a9 9 0 11-18 0 9 9 0 0118 0z'
	);
};
```

Where we map to the svg path of the play button - this would enable us to handle pause state better. What do you think?

Slickerius:
Thank you for the review. That's a great idea - I have added it to the connector :)

inverse (approved):

PR #3892, opened 2023-06-13
Fix Idagio connector

LibertyFrog:
Fixed the Idagio connector not recognizing when a track is playing.

**Additional context**
The Idagio connector was not scrobbling music. The `isPlaying` check expects the play button label to equal "Pause" when a track is playing which is also what is visible in the DOM, however they use a `text-transform: uppercase` CSS styling on it.
I call `toUpper` on the found text, so it doesn't break again should they remove the styling.

### DIFF
--- a/src/connectors/idagio.ts
+++ b/src/connectors/idagio.ts
@@ -23,7 +23,7 @@ Connector.currentTimeSelector = '.player-PlayerProgress__progress--2F0qB>span';
 Connector.durationSelector = '.player-PlayerProgress__timeTotal--3aHlj span';
 
 Connector.isPlaying = () =>
-	Util.getTextFromSelectors(pauseButtonSelector) === 'Pause';
+	Util.getTextFromSelectors(pauseButtonSelector)?.toUpperCase() === 'PAUSE';
 
 Connector.isScrobblingAllowed = () =>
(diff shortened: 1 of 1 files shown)


hornbuckle:
If the text inside the element is "Pause," CSS text-transform shouldn't have any effect on the connector script since it would only affect how the browser renders the text.
Can you provide a link to show the issue you're trying to fix?

LibertyFrog:
> If the text inside the element is "Pause," CSS text-transform shouldn't have any effect on the connector script since it would only affect how the browser renders the text. Can you provide a link to show the issue you're trying to fix?

That is not the behaviour I'm observing. As for a link: go to https://app.idagio.com and click the play icon on any album tile. When I do it the extension doesn't start scrobbling and when I look at the `State` object I see the `isPlaying` property is still `false`.

This is what I see in the DOM for the pause button selector, "Pause":
pause-in-dom

This is the CSS applying the `text-transform`:
css-class

I added logging to the isPlaying getter (without the `toUpper` I added later):
log-text

This is the log output, all uppercase "PAUSE":
log-value

This is the log output after toggling `text-transform` in the browser dev tools (no other change):
css-no-text-transform
log-value-no-text-transform

inverse:
Any updates on this one?

yayuyokitano:
Merging this. Improvement suggested by hornbuckle can be applied in a separate PR.

PR #3908, opened 2023-06-20
Added Pleroma scrobbling

TheStranjer:
Added Pleroma scrobbling

**Additional context**
PleromaBE is a microblogging service that is compatible with Mastodon. Its API has scrobbling support. This PR added that.


TheStranjer:
This is very similar to the Maloja scrobbler with a few differences. That being said, it does work.

It working in-browser:
image

Pleroma `GET` API working:
image

I will need to implement something for the frontend to view these results but having a web scrobbler support Pleroma will bring things forward quite a bit.

inverse:
@yayuyokitano  what do you think of this change?  Seems reasonable to me

src/_locales/el/messages.json
```diff
@@ -376,7 +376,7 @@
         "description": "'About' section text"
     },
     "aboutExtensionDesc": {
-        "message": "Web Scrobbler is a browser extension created for people who listen to music online through their browser, and would like to keep an updated playback history using scrobbling services, such as Last.fm, Libre.fm and ListenBrainz.",
+        "message": "Web Scrobbler is a browser extension created for people who listen to music online through their browser, and would like to keep an updated playback history using scrobbling services, such as Last.fm, Libre.fm, ListenBrainz, and Pleroma.",
```
inverse, line 379: Not sure how this will work? We pull translations from remote, only maintaining source language (en).

src/_locales/en/messages.json
```diff
@@ -179,6 +179,14 @@
         "message": "Custom API URL",
         "description": "Input label"
     },
+    "instanceDomain": {
+        "message": "Instance Domain",
+        "description": "Input label"
+    },
+    "instanceDomainPlaceholder": {
+        "message": "ex. iddqd.social",
```
inverse, line 187: Should we reference one of the instances highlighted on the official site https://pleroma.social?

inverse, line 187:
```suggestion
        "message": "ex. stereophonic.space",
```

Should we reference one of the instances highlighted on the official site https://pleroma.social?

TheStranjer: Listing `iddqd.social` is just a vanity thing as it's the instance I run. `stereophonic.space` has nothing to do with music. I recommend using my instance because, as it stands, it's the only one with a frontend that actually makes any use of the scrobble information submitted. Getting people to use the information submitted is a conversation I'm having on the other side of this conversation. @inverse

TheStranjer: If you're still at all interested in this commit, I can send it to `shitposter.club`, which has this implemented as well, and it isn't mine, if you're worried about me ego wanking about it too much. @yayuyokitano @inverse

src/_locales/el/messages.json
TheStranjer: I just removed all these because I believe I made them in error. Currently the only one that's edited is the English one, which I hope is right.

TheStranjer:
@yayuyokitano I just fixed the merge conflicts. Merry Christmas! 🎅

TheStranjer:
I wasn't getting those linter errors before. I'll fix them when I return home this coming week

TheStranjer:
@yayuyokitano @inverse Try now. The errors/linter problems don't appear present (at least when I run them)

yayuyokitano:
Looks like just some automatically fixable stuff through `npm run prettierfix`.
Similar to inverse I'm a bit worried about the example, as writing down an instance is kind of an implicit endorsement, so I think avoiding instances that allow controversial content is a good idea.
Do you know if there is such an instance that has updated pleromaFE sufficiently yet?
Alternatively we could go the pleroma API docs route and use something like `mypleroma.com` which I think should still convey what we need it to convey.

TheStranjer:
@yayuyokitano Two things:

* I can come up with a bunch of different instances that run updated PleromaFE, but most of them allow for controversial content, as you call it. In fact, I can't think of a single Pleroma instance that doesn't allow content that someone would find controversial. There's a lot of political talk on Pleroma. For this reason, instead of trying to find an instance that allows for no controversial content, I made the domain `example.com` just to convey that in order for it to work, they need to enter a domain, and JUST a domain.

* I ran `npm run prettierfix` like you said and committed its output.

yayuyokitano:

PR #3930, opened 2023-06-30
Fix popup links width

dokugo:
Added `width: fit-content` to `PopupAnchor`, `TrackMetadata` and `TrackControls` component styles.

**Additional context**
Until the recent 3.0 update, the links (track, artist, album) in the popup took up just enough space for the text inside them, but since the update these links have started to take up all available space. Because of this, users can't double-click on the nearby empty space right to the links to quickly select a track or artist or album link and copy its text to the clipboard for a fast googling.

Before 3.0:
For example, I was able to double click on the space marked with the yellow circle and quickly select the artist name. Then I could press Ctrl+N to open a new window, Ctrl+V to paste the artist name from the clipboard, and Enter to google.

After 3.0:
If there is a cursor over the space marked with the yellow circle, the cursor appears to be of pointer style and clicking here opens a new tab.

Here we can see that the artist link takes up all available space:

### DIFF
--- a/src/ui/popup/popup.module.scss
+++ b/src/ui/popup/popup.module.scss
@@ -90,6 +90,7 @@ a {
 			overflow: hidden;
 			text-overflow: ellipsis;
 			white-space: nowrap;
+			width: fit-content;
 		}
 
 		@supports (-webkit-touch-callout: none) {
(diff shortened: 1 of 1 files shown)


inverse:
Thanks! Seems reasonable to me

Issue #3821, opened 2023-05-30
Album art is too large for some albums


Mirika1:
**Describe the bug**
The art for some albums is too large when you open Web Scrobbler to edit the track.

Albums I've noticed this with so far:
* LoveLive! Sunshine!! Second Solo Concert Album THE STORY OF FEATHER starring Kurosawa Ruby
* LoveLive! Sunshine!! Third Solo Concert Album THE STORY OF “OVER THE RAINBOW” starring Kurosawa Ruby
* LoveLive! Sunshine!! Second Solo Concert Album THE STORY OF FEATHER starring Watanabe You
* LoveLive! Sunshine!! Second Solo Concert Album THE STORY OF FEATHER starring Takami Chika
* LoveLive! Sunshine!! Third Solo Concert Album THE STORY OF "OVER THE RAINBOW" starring Takami Chika

**How to reproduce**
1. Go to open.spotify.com
2. Play any song from the albums listed above
3. Click on the Web Scrobbler icon in the top right corner
4. Notice the very large album art

**Expected behavior**
The album art size should be the same size as other albums.

**Screenshots**
Observed size:
https://i.imgur.com/29QiR4D.png

Expected size:
https://i.imgur.com/zSAjgA6.png

**Environment**
-   OS: Windows 11
-   Browser: Opera, Google Chrome
-   Extension version: v3.0.10

yayuyokitano:
Thank you for reporting! Since the popup size is not fixed, there seems to be an issue where the image sometimes “races” the overall popup height to an unreasonable size. This bug is currently high priority.

PR #3940, opened 2023-07-01
Fix popup regression

yayuyokitano:
#3930 introduced a regression reverting the fix to #3821 
This should fix that without undoing the PR itself


inverse:
:pray: thanks for the regression fix!

PR #3972, opened 2023-07-18
Add connector for Radio Carpice (radcap.ru)

projecteurlumiere:
Add a connector for Radio Caprice (radcap or http://radcap.ru).

**Additional context**
I had some doubts about the label naming: 

- RADio CAPrice (RADCAP) is the title on their main page,
- but their app's title is Caprice Radio, 
- and throughout their website and on many radio aggregators, it is simply Radio Caprice. 

I opted for the last one because it seemed more widespread.

Other than that, that is my first PR ever - I hope I got everything correctly.

### DIFF
--- a/src/connectors/radcap.ts
+++ b/src/connectors/radcap.ts
@@ -1,0 +1,39 @@
+export {};
+
+function setupConnector() {
+	if (isDesktopPlayer()) {
+		setupDesktopPlayer();
+	} 
+  else {
+		setupMobilePlayer();
+	}
+}
+
+function isDesktopPlayer() {
+	return Boolean(document.querySelector('.radio'));
+}
+
+function setupDesktopPlayer() {
+	Connector.playerSelector = '.radio';
+
+  Connector.artistTrackSelector = '.stream.stream-signal > tbody > tr > td';
+
+  Connector.playButtonSelector = '#oframeplayer-d > pjsdiv:nth-child(7)';
+}
+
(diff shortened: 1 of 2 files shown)


src/connectors/radcap.ts
```diff
@@ -0,0 +1,39 @@
...
+function setupMobilePlayer() {
+  Connector.playerSelector = '.container';
+
+  Connector.artistTrackSelector = 'table.stream > tbody > tr > td';
+
+  const upperPlayerPlayButton = '#oframeaudioplayer2 > pjsdiv:nth-child(8) > pjsdiv:nth-child(2)';
+
+  const lowerPlayerPlayButton = '#oframeaudioplayer0 > pjsdiv:nth-child(8) > pjsdiv:nth-child(2)';
+
+  Connector.isPlaying = () => {
```
inverse, line 33: Is this working for you? For me the play/pause is not working.

inverse (commented):
Looks goot to me except the mobile interaction for play/pause - looks like a nice site for discovering genres :)

src/connectors/radcap.ts
projecteurlumiere: You were right! I have just fixed it. Take a look, please.

Selectors are fine. Apparently, the problem is that they remove the play button with `visibility: hidden` instead of `display: none`, so `Util.isElementVisible` did not work in this case.

inverse (approved):

PR #3975, opened 2023-07-18
Fix (mostly any-related) issues introduced by the ESLint upgrade.

jaccarmac:
Related to #4003.


jaccarmac:
This is just a grab bag of changes for right now. One question since I haven't touched the scripts in a while: How does one run the formatter if not in VSCode?

jaccarmac:
This look reasonable? There are unrelated formatting changes that I can strip out if needed and it might be worth noting some of the changes. But it's almost all narrowing from `any`, which should make us safer if anything.

inverse:
Seems reasonable to me - looks a lot stricter!

@yayuyokitano what do you think?

jaccarmac:
Related: Is there a known solution for the stylelint warnings? We're two majors behind now with mostly-similar messages.

jaccarmac:
@inverse @yayuyokitano I'm going to press the button on this in a couple days if there are no objections. That'll let us clear one more Dependabot PR. However, we'll also need to look at releasing a build around that time, which I may or may not be able to do and certainly have not done before.

Issue #3319, opened 2022-08-08
SiriusXM DJ breaks being scrobbled on certain stations


pixiekat:
I don't know if this is technically a bug report or not since I wasn't sure if anything can be done about it.

When I listen to certain stations - Lithium, Pop Rocks, etc - you'll occasionally get an interlude scrobbled which looks like the following:

`@JuliaCunningham | @SXMPopRocks | 5 minutes ago`

Is there any way to filter these out or have them not scrobble? It's not a big deal but it just irks me while I'm trying to curate my Last.fm. :)

inverse:
It looks like some effort was made to remove these. 

Is what you provided the artist?

If it's the artist we could include `@SXMPopRocks` into the filter to ignore. Right now this filter includes `siriusxmu`

pixiekat:
When it scrobbled it linked to a page like this: https://www.last.fm/music/@radiomadison/_/@sxmlithium (this is another station but it's similar). So, I think the station is the track. 

But after I made this post, it also scrobbled a little advert for the station on Facebook too (That was on the 1st Wave channel)

Issue #2222, opened 2020-01-02
Add option to exclude tracks based on pattern matching


NeilN1:
Add two fields (one for artist, one for track title) a user would use to enter lists of artists and track titles. A track having info matching the artist or track title would not be scrobbled.

This would stop the scrobbling of promos/DJ chatter some radio-like services play.  Example from SiriusXM: Artist: @SiriusXMHits1 Track: @MorningMashUp

alexesprit:
@NeilN1 We already have a simple rule for SiriusXM. Instead of adding new options, I suggest to add excludes directly to SiriusXM connector. The advantage of this way is it will work for ALL users w/o any configuration. I'm pretty sure no one wants to scrobble non-music stuff.

> This would stop the scrobbling of promos/DJ chatter some radio-like services play. Example from SiriusXM: Artist: @SiriusXMHits1 Track: @MorningMashUp

This could be fixed by adding a rule like this:
> If an artist name and (maybe or?) a track title start with "@" symbol, then ignore this track.

If there're additional things should be skipped, please let me know.

NeilN1:
Fair enough. I didn't want to bother the developers with constant exclusion requests so I suggested something user based.  I will keep an eye out for other examples and post here (let me know if there's somewhere else you'd prefer I send the requests).

seafea:
ignoring track titles that start with @ would filter out a good chunk of them, but there's still some like "w/ @kadeshow" or "Wednesday DL"

Could we maybe have a list of known DJ / promo tracks to specifically exclude? I could maybe help out with this one.

alexesprit:
@seafea yes, feel free to post this list here. I will add it to the SiriusXM connector.

JunkiEDM:
+1, I frequently use Soundcloud for listening but had to stop using this extension because it kept scrobbling tracks & artists that I don't want scrobbled.

PR #3977, opened 2023-07-19
Updated SiriusXM Connector with more filters

jbwharris:
Related to #2222 #3319 

Added some improved filtering for station IDs and DJ handles. I did some testing and playing around with finding different criteria for exclusion and found some weird edge cases. 

- My filter now checks the artist and track instead of just artist, as it increases the percentage of properly filtering the station IDs and Twitter handles
- I discovered that filtering by ﹫ tends to be too broad. The character is used sometimes with certain channels to denote the artist, most prominently The Grateful Dead channel using ﹫greatfuldead all the time. Or it can be used in some like "Live ﹫ SiriusXM" which would still be a valid scrobble.
- I added a few Twitter account handles, but found the list tended to be too extensive to filter against them all. We could compile such a list, but I don't readily have that list handy. The good thing is that some of the filters in tandem with the Twitter filters tends to get most of the cases (ie. like where they have a host handle with a URL or hashtag)
- I added a ".c" filter which would catch the vast majority of URLs, both .ca and .com. I'm not sure if there are other regionalizations of SiriusXM that might have URLs that should be filtered. I didn't encounter any others.
- Adding #️⃣ tends to catch a lot of the ones we want to filter. I did note this had a possibility to excluding classical tracks, but looking at the lastfm data I didn't see much indication people were probably scrobbling those channels.
- I tried having "siriusxm" but found that one was way too broad. It would exclude live sessions or some show names. Didn't want to limit how some people might use Web Scrobbler for tracking.
- I also tweaked the trackArtSelector, as it had been pulling null previously

**Scrobbling Data I used to find filter criteria**
- https://www.last.fm/search/tracks?q=%40sxm
- https://www.last.fm/search/tracks?q=siriusxm
- https://www.last.fm/search/tracks?q=fb.com

### DIFF
--- a/src/connectors/siriusxm-player.ts
+++ b/src/connectors/siriusxm-player.ts
@@ -48,8 +48,6 @@ Connector.isScrobblingAllowed = () => {
 	return !filteredTerms.some(
 		(term) => artist?.includes(term) || track?.includes(term)
 	);
-
-	return !filteredTerms.track?.includes("@");
 };
 
 Connector.applyFilter(filter);
--- a/src/connectors/siriusxm-player.ts
+++ b/src/connectors/siriusxm-player.ts
@@ -10,34 +10,40 @@ export {};
 const filter = MetadataFilter.createFilter({ track: removeYear });
 const removeYearRe = /\s\(\d{2}\)\s?$/g;
 
-Connector.playerSelector = '.sxm-player-controls';
-
-Connector.artistSelector = '.sxm-player-controls .artist-name';
-
-Connector.trackSelector = '.sxm-player-controls .track-name';
+Connector.playerSelector = ".sxm-player-controls";
+Connector.artistSelector = ".sxm-player-controls .artist-name";
+Connector.trackSelector = ".sxm-player-controls .track-name";
 
 Connector.isPlaying = () => {
 	return (
 		Util.getAttrFromSelectors(
-			'.sxm-player-controls .play-pause-btn',
-			'title'
-		) === 'Pause'
+			".sxm-player-controls .play-pause-btn",
+			"title"
+		) === "Pause"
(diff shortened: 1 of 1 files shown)


src/connectors/siriusxm-player.ts
```diff
@@ -10,34 +10,47 @@ export {};
...
+	const filteredTerms = [
+		"@siriusxm",
+		"@jennylsq",
+		"@radiomadison",
+		"@morningmashup",
+		"josiah",
+		"1-877-33-sirius",
+		"@sxm", //will broadly catch a bunch of sxm Twitter handles
+		"altnation",
+		".c", // will catch .com and .ca URLs
```
inverse, line 40: Would it be better to be more explicit and have two entries for `.ca` and `.com`?

jbwharris: @inverse I can do that. Probably getting too cute doing that anyways 😂

jbwharris: @inverse I pushed that change over

src/connectors/siriusxm-player.ts
```diff
@@ -10,34 +10,50 @@ export {};
...
+		"facebook",
+		"twitter",
+	];
+
+	return !filteredTerms.some(
+		(term) => artist?.includes(term) || track?.includes(term)
+	);
 
-	return false;
+	return !filteredTerms.track?.includes("@");
```
inverse, line 52: Should this be: 

```suggestion
	return !track.includes("@");
```

jbwharris: @inverse I just removed that line. Was experimenting with another idea for the "@" symbol, but found it was too greedy.

PR #4000, opened 2023-07-29
Fix Globalplayer (RadioX, Smooth) Connector

jbwharris:
From what I can tell Global Player has been broken for awhile. This connector powers a bunch of UK radio stations like https://radiox.co.uk, https://smoothradio.com and https://heart.co.uk/ and I believe there was a big redesign in the last 6 months that was never addressed by the connector. 

I was able to get the connector working again with live radio, as well as being able to scrobble the curated playlists the site has. The radio archives don't have the tracks split up, so those can't be scrobbled. 

## Known Issues
The one caveat about getting this site to work is that scrobbling will stop working if you navigate away from the particular playlist or live show you're listening to. The site doesn't use the persistent bottom bar to show the track meta (only the show or playlist name), so it's possible to have the music still playing while you navigate the site, but it will only scrobble correctly if you're on the show page because that's the only place that track meta exists to pull in. I attempted to pull it from the `<title>` tag, but that stops updating if you navigate away from the page you're listening on. So it's imperfect based on how the site is constructed. 

## Live Radio
image

## Playlist Playback
image

### DIFF
--- a/src/connectors/globalplayer.ts
+++ b/src/connectors/globalplayer.ts
@@ -1,14 +1,16 @@
 export {};
 
-const showNameSelector = '.show-info-with-image__text__title';
+const showNameSelector =
+	'div[class^="style_playbarInfo"] div[data-testid="show-info-title"]';
 
-Connector.playerSelector = '.track-container';
-
-Connector.trackSelector = '.track-container .show-info__text__title';
-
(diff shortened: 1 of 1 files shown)


jbwharris:
@inverse Is there something I should be doing that I'm not with this type of PR? Am I able to add PR labels?

PR #4005, opened 2023-08-02
 Update SiriusXM track meta filtering

jbwharris:
I've been working to really cull down the extra cruft in the SiriusXM connector submissions lately. I have used the connector a fair bit recently and it's significantly improved the quality of submitted scrobbles. #3977 was focused on filtering out info that shouldn't be sent as scrobbles to lastfm. This update is focused on removing extraneous info within the tracks being sent from the various channels the service offers. 

- Refactored the existing filter to handle removing the year (78), (Artist Name Cover) and EXCLUSIVE from submitted tracks
- Updated the connector comments

## SiriusXMU initial track
image

## Successful lastfm submission with Cover brackets and EXCLUSIVE removed
image


src/connectors/siriusxm-player.ts
```diff
@@ -1,14 +1,19 @@
...
+ * - A lot of stations seem to employ adding EXCLUSIVE to the track name, so that is now removed
  *
- * This filter removes such suffixes.
+ * Filter Station Meta
+ * - There is now a filter for much of the station meta on SiriusXM. Removes social media handles, URLs, shownames
  */
 
-const filter = MetadataFilter.createFilter({ track: removeYear });
-const removeYearRe = /\s\(\d{2}\)\s?$/g;
+const filter = MetadataFilter.createFilter({ track: removeStationMeta });
```
jaccarmac, line 14: How feasible is it to keep the year removal as a separate regex and remove the other information on a new pass? It's reasonable now, but I always second-guess complex regex.

src/connectors/siriusxm-player.ts
```diff
@@ -1,14 +1,19 @@
...
+const removeExtraRe =
+	/\s\(\d{2}\)\s?|\([^)]*\b(?:cover|Cover)\b[^)]*\)|EXCLUSIVE/gi;
```
jaccarmac, line 16: Removing "cover" for that case seems like a good improvement; However, I imagine there are a non-zero number of songs which legitimately end in (... cover). Misinterpreting those is not a big deal, but if there's any way to access the channel information, we could cut down that false negative (positive?) rate.

jaccarmac (commented):
This looks nice overall!

src/connectors/siriusxm-player.ts
jbwharris: @jaccarmac That shouldn't be much issue. Honestly I'm terrible with regex and ran it through ChatGPT to optimize the 3 filters. I can definitely split them out into 3 separate filters instead.

src/connectors/siriusxm-player.ts
jbwharris: @jaccarmac I think pulling in the show name wouldn't be too hard. I updated the Global Player connector recently and they had a show name filter, so the idea of filtering based on the channel meta seems reasonable.

jbwharris: @jaccarmac I separated the 3 regex criteria into different filters, though didn't do the channel identification. I did a search on [XMPlaylist](https://xmplaylist.com) and see there are a bunch of channels that do have cover labels, and we'd really have no idea which ones might have "cover)" as part of the official title of the song. I think it's an all or nothing proposition. I searched how some of them are labelled in Apple Music, and it's without the mention of being a cover. Though you can see many are exclusive, so may not exist anywhere aside from on SXM. Which I suppose doesn't matter, since last.fm is just logging the listen, not the validity or findability of the song.  

image

inverse:
@jaccarmac ready to merge this one?

jaccarmac:
Thanks for the contribution!

PR #4033, opened 2023-08-17
Improve onboarding and other UX

yayuyokitano:
- Reimplements the notification telling user to sign in on startup if not signed in.
- Adds functionality to go to a specific part of options using URL.
  - Used in disabled connector popup item to go to connector list.
  - Used in service error popup item to go to accounts.
  - Used in onboarding notification to go to accounts.
- Also, one small unrelated change, reimplementing the code updating the version number in storage in case it's needed in the future, was missed in the V3 migration.

I think 3.1 should probably be released after this + pleroma and sirius PRs (and a quick sanity test of everything together, of course)

### DIFF
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -29,6 +29,7 @@ import {
 import { CloneableSong } from '@/core/object/song';
 import {
 	clearNowPlaying,
+	showAuthNotification,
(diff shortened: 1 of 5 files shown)


inverse (approved):
Nice one

PR #4049, opened 2023-08-24
Add Webhook support

dkordik:
Add ability to wire up to a custom Webhook to consume nowplaying, scrobbled, paused and resumed-playing events

**Additional context**
Reused the "Accounts" feature to keep complexity and risk low, since this isn't primary path functionality. Not perfect- requires you to refresh the extension when you change the URL, since there's no actual session to log in/out of. But this was acceptable for my needs.

---
image

image

---
Additional note: I noticed Prettier adding trailing commas to the files I touched. Upon investigation, this appears to be the [default Prettier behavior after 3.0](https://prettier.io/docs/en/options.html#trailing-commas), with [this project on 3.0.1](https://github.com/web-scrobbler/web-scrobbler/blob/master/package.json#L60) (at the time of this writing). The [prettier-config-web-scrobbler](https://github.com/web-scrobbler/prettier-config-web-scrobbler/blob/master/index.js) project doesn't take a stance on trailing commas, so I suppose this is why it defaults to this behavior now. I figured it'd be best to not try to fight it- as that's not the Prettier way :). But if we want to update the `prettier-config-web-scrobbler` setting, I'd be happy to apply updated formatting, which would certainly make this diff a lot less noisy.


yayuyokitano:
I think we shouldn't modify our config - guess it got bumped by dependabot and changes werent checked so thoroughly.
I guess we just deal with the noise for now and then we can get the migration done more thoroughly later.
What do you guys think @web-scrobbler/web-scrobbler-team 

Forgive me for asking without looking too thoroughly at the diffs yet, is this using some standardized format?
Or is it basically just dumping data with some event names attached?
If it is the latter, we may want to do some documentation (I guess this would be more of a wiki angle)

dkordik:
Hi!
> Forgive me for asking without looking too thoroughly at the diffs yet, is this using some standardized format?
Or is it basically just dumping data with some event names attached?
If it is the latter, we may want to do some documentation (I guess this would be more of a wiki angle)

it's the latter- just passing along the data with event names attached.

yayuyokitano:
Figured.
Thank you for contributing this.
I’ve actually been mulling over the concept of adding discord rpc support, but since they *really* dont like you accessing that through browser, this + a separate executable might be a more sensible approach.
Anyway – that time, that sorrow.

i’ll look into this in a bit more detail asap, though im in the middle of moving, but hopefully not too long.

jaccarmac:
> I think we shouldn't modify our config - guess it got bumped by dependabot and changes werent checked so thoroughly.
> I guess we just deal with the noise for now and then we can get the migration done more thoroughly later.
> What do you guys think @web-scrobbler/web-scrobbler-team

Re: this issue: Does CI fail if you make commits without conforming to Prettier perfectly? I was under the impression it doesn't since I had to figure out how to run Prettier manually the last time I did a significant change.

Re: configs more broadly: This really belongs in a dedicated issue, but IMO we should make two changes:
1. Make formatting automatic/CIed so we can have one giant formatting commit during upgrades and then move on.
2. Take a critical look at the way our configurations are managed; They exist in separate repositories, which makes changes slower and forces a second approval process to handle upgrades. My attempt to fix #4034 is blocked because of these two issues.

jaccarmac:
#4041's also open, and 3.0.2 looks like it reverts some kind of trailing comma change. Not sure if it's the one in question here.

yayuyokitano:
> > I think we shouldn't modify our config - guess it got bumped by dependabot and changes werent checked so thoroughly.
> > I guess we just deal with the noise for now and then we can get the migration done more thoroughly later.
> > What do you guys think @web-scrobbler/web-scrobbler-team
> 
> Re: this issue: Does CI fail if you make commits without conforming to Prettier perfectly? I was under the impression it doesn't since I had to figure out how to run Prettier manually the last time I did a significant change.
> 
> Re: configs more broadly: This really belongs in a dedicated issue, but IMO we should make two changes:
> 
> 1. Make formatting automatic/CIed so we can have one giant formatting commit during upgrades and then move on.
> 2. Take a critical look at the way our configurations are managed; They exist in separate repositories, which makes changes slower and forces a second approval process to handle upgrades. My attempt to fix [Bump stylelint from 13.13.1 to 15.10.3 #4034](https://github.com/web-scrobbler/web-scrobbler/pull/4034) is blocked because of these two issues.

It does not fail afaik.

1. this would be nice.
2. the reason we have a separate repo for configs is that we have multiple repos and we want to have a shared codestyle between them. I think this is a good thing. However, they are indeed in dire need of updates. With v3 having been a bit rushed this repo currently exists in a bit of a grayzone between using that config and also doing some extending of its own – we should be removing the vue part of the config repo (afaik unused) and replacing it with a solid config that uses the rules adapted by this repo probably. I think the main issue with stylelint major update is that sass is no longer supported out of the box and we need to do some extra loading there.

the 3.0.2 change should be unrelated to this.

Anyway, let’s make an issue for this, I guess something along the lines of overhaul linting.

Issue #3822, opened 2023-05-31
Extension frequently doesn't recognize existing edits


AquaDoesStuff:
Title; I'm not sure if this is intentional, but it's annoying enough to where I decided I should probably write an issue about it. I listen to most of my music on Soundcloud, which often results in incorrect detected metadata by the extension, so I set it to only scrobble tracks I've saved edits for for it to automatically detect the edited metadata later and use on replays. However, even after the edit is saved, it often prompts that the track is unrecognized again, and forces me to edit it once more, completely ignoring the existing edit. If I view my saved edits, I can see that there is more than one edit saved per repeated tracks. Is there a way to fix this? Changing the mode doesn't seem to fix it either.

yayuyokitano:
This is probably a soundcloud connector issue with getting a unique ID for the song. This should not be happening. Thank you for reporting!

inverse:
Do you have a sample page which is problematic? I'm unable to reproduce this.

AquaDoesStuff:
it seems to be an issue based on playing a track on its own page vs in a playlist or an album, which I've already noticed already in soundcloud itself that if you're playing the track from an album it doesnt show that you're playing it on its page so I believe that may help with reproducing it if that's related to the issue.

yayuyokitano:
Makes sense. I would guess the URL being played from somehow affects the unique ID

inverse:
Resolved in #4050

PR #4050, opened 2023-08-25
Ignore search parameters when getting a unique ID for SoundCloud

everlyy:
This patch makes it so that the SoundCloud connector doesn't include the search
parameters when it's asked for a unique ID.

This fixes the problem of having to re-edit your scrobble if you're listening from
an album or playlist.

`src/connectors/soundcloud.ts`: Made it so when getUniqueID() is called it doesn't return a URL with search parameters so you don't have to re-edit a scrobble if you're listening to the same song in a different playlist/album.

**Additional context**
Made this change because it was slightly annoying having to edit my scrobble every time I listened to a playlist or an album.


yayuyokitano:
Possibly related to #3822

Issue #3110, opened 2022-01-27
Way to disable logging to console?


Tamarindo94:
The logs might be useful, but they become annoying when trying to debug other scripts by using logs. I haven't found a way to disable them...

inverse:
Could be exposed as a setting within core - happy to have someone contribute this 👍🏼

EvHaus:
console.log spam seems to have gotten a lot worse in v3.x.x. Is it possible to disable this?

![image](https://github.com/web-scrobbler/web-scrobbler/assets/226640/c1ecc039-f65e-4bd9-87e5-1d1fa9d87b84)

Tamarindo94:
The logging seems to be done by function Util.debugLog() in file util.ts; it could be modified to print only if a certain variable is true - the variable could be a user setting, or the installType of the extension, something like: 
```
if(chrome.management.get('installType') == "development" || userSetting === true) {
  // print debug
} else {
  // don't print
}
```

xul7:
I'm sure it was added for debugging and then forgot to remove it

https://github.com/web-scrobbler/web-scrobbler/blob/a4d2d0de1753532c905aaa1a9527d1ea191679f1/src/core/storage/wrapper.ts#L241

mklusak:
This is insane, how am I supposed to develop something and find my output to Dev console, when I have the dev console full of hundreds of useless lines from this extensions?

But, I found a quick solution how to remove Chrome extensions outputs from Dev console -> https://superuser.com/a/1346827

busybox11:
You can also just right click on any of webscrobbler's logs and click on 'Hide messages from main.js' (it will only hide WebScrobbler's index.js logs)
![image](https://github.com/web-scrobbler/web-scrobbler/assets/29630035/3c8c1f8e-3082-4f51-869d-1e278fc37457)

inverse:
Some noisy logs have been removed from production builds - which should improve things. 


But as @busybox11 said you can easily filter out logs from our extension.

on3dd:
It's ridiculous that you're logging something using console.log in production environment. When will we be able to turn it off without messing with "Hide messages from..."? 🤦🏻

LasseGuldhammer:
There should definitely be an option to disable the logging. As a developer, these logs are simply noise that I can't turn off.

inverse:
> There should definitely be an option to disable the logging. As a developer, these logs are simply noise that I can't turn off.


Pull requests welcome :)

[...]

PR #4067, opened 2023-08-29
Silence console logs from content scripts by default

blimpage:
## Changes

Adds a new "Enable debug logging" option to the extension's global options (default: `false`). All logging from content scripts to the browser console is silenced unless the user manually enables the option.

## Context

Web Scrobbler can sometimes print a _lot_ of messages to the user's browser console, even on websites that don't have a connector. For example:

image

This can be very annoying for users who frequently use their browser console, since it can make it hard to find other messages printed there, and there's currently no way to silence console messages from Web Scrobbler other than disabling/uninstalling the extension. There are a few existing issues about this: #3828, #3110

This PR adds a new "Enable debug logging" option to Web Scrobbler's global options, sets it to `false` by default, and only allows the extension to print messages to the browser console if the user has switched it to `true`.

This doesn't affect any of the logging to the extension's service worker console or the console on the extension's options pages, just logs from content scripts to the user's main browser console when they're on a regular web page.

## Screenshots

New "Enable debug logging" option added to the global options UI:

image

## Considerations

[The instructions for collecting logs from content scripts](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension#logs-of-content-scripts) would need to be updated once this change is released, since users will now need to switch on "Enable debug logging" before any logs will appear.

There are a few aspects of this PR that could use some improvement, I'd love any help making these better since they're a bit beyond my skills with TypeScript:

- ~~I've added a new `getGlobalOption()` function that has a return type of `Promise<unknown>`. The Promise should only ever resolve to one of the value types of [the `GlobalOptions` interface](https://github.com/web-scrobbler/web-scrobbler/blob/1f455fd/src/core/storage/options.ts#L23), but I couldn't work out the syntax for that so I've left it as `unknown`.~~ Now resolved as of a1101ef.
- ~~The new `getGlobalOption()` function duplicates a lot of code from the existing `getOption()` function above it, so they could probably be combined or refactored to share some code. I couldn't reuse `getOption()` in its current state because it calls `debugLog()`, which could lead to an infinite loop if it's called from within `debugLog()`. If maintainers would be okay with removing that `debugLog()` call, then the functions would become easier to refactor.~~ Now resolved as of a1101ef.
- I haven't added any tests for this new behaviour because there are no existing tests for `debugLog()` or `getOption()` for me to extend, and I'm not familiar enough with TypeScript or vitest to confidently add new tests.

### DIFF
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -7,7 +7,7 @@ import type {
 	TimeInfo,
 	TrackInfoWithAlbum,
 } from '@/core/types';
-import type { ConnectorOptions } from '@/core/storage/options';
+import type { GlobalOptions, ConnectorOptions } from '@/core/storage/options';
 import type { DebugLogType } from '@/util/util';
 
 const BrowserStorage = (async () => {
@@ -704,6 +704,25 @@ export async function getOption(
 }
 
 /**
+ * Read global option from storage.
+ * @param key - Option key
+ * @returns Option value
+ */
+/* istanbul ignore next */
+export async function getGlobalOption(
+	key: keyof GlobalOptions,
+): Promise<unknown> {
+	const awaitedStorage = await BrowserStorage;
+	const data = await awaitedStorage
+		.getStorage(awaitedStorage.OPTIONS)
+		.get();
+	if (data && key in data) {
+		return data[key];
+	}
+	return false;
+}
+
+/**
  * Normalize given URL. Currently it only normalizes
  * protocol-relative and root-relative links.
  * @param url - URL, which is possibly relative
@@ -747,8 +766,13 @@ export function injectScriptIntoDocument(scriptUrl: string): void {
  * @param logType - Log type
  */
 /* istanbul ignore next */
-export function debugLog(text: unknown, logType: DebugLogType = 'log'): void {
+export async function debugLog(text: unknown, logType: DebugLogType = 'log'): Promise<void> {
 	const logFunc = console[logType];
+	const loggingEnabled = await getGlobalOption('debugLoggingEnabled');
+
+	if (!loggingEnabled) {
+		return;
+	}
 
 	if (typeof logFunc !== 'function') {
 		throw new TypeError(`Unknown log type: ${logType}`);
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -746,27 +746,69 @@ export function injectScriptIntoDocument(scriptUrl: string): void {
 }
 
 /**
+ * Handle async checks for DEBUG_LOGGING_ENABLED option while ensuring
+ * that logs are still printed in a predictable order.
+ */
+class DebugLogQueue {
+  private queue: {text: unknown, logType: DebugLogType}[] = [];
+  private isActive = false;
+
+  /**
+   * Enqueue a log message to be printed.
+   * @param text - Debug message
+   * @param logType - Log type
+   */
+  public push(text: unknown, logType: DebugLogType): void {
+    this.queue.push({text, logType});
+    this.start();
+  }
+
+  /**
+   * Process the queue to print logs in order.
+   */
+  private async start(): Promise<void> {
+    if (this.isActive) return;
+    this.isActive = true;
+
+    try {
+      for (let i = 0; i < 100 && this.queue.length > 0; i++) {
+        const currentMessage = this.queue.shift();
+        if (currentMessage && await this.shouldPrint()) {
+          const logFunc = console[currentMessage.logType];
+
+          if (typeof logFunc !== 'function') {
+            throw new TypeError(`Unknown log type: ${currentMessage.logType}`);
+          }
+
+          const message = `Web Scrobbler: ${currentMessage.text}`;
+          logFunc(message);
+        }
+      }
+      this.isActive = false;
+    } catch(err) {
+      this.isActive = false;
+    }
+  }
+
+  /**
+   * Check if the user has enabled debug logging in the extension options.
+   */
+  private async shouldPrint(): Promise<boolean> {
+    const awaitedOptions = await Options;
+    const debugLoggingEnabled = await awaitedOptions.getOption(awaitedOptions.DEBUG_LOGGING_ENABLED);
+    return !!debugLoggingEnabled;
+  }
+}
+const debugLogQueue = new DebugLogQueue();
+
+/**
  * Print debug message with prefixed "Web Scrobbler" string.
  * @param text - Debug message
  * @param logType - Log type
  */
 /* istanbul ignore next */
-export async function debugLog(text: unknown, logType: DebugLogType = 'log'): Promise<void> {
-	const awaitedOptions = await Options;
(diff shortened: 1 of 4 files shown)


src/core/content/util.ts
```diff
@@ -747,8 +766,13 @@ export function injectScriptIntoDocument(scriptUrl: string): void {
  * @param logType - Log type
  */
 /* istanbul ignore next */
-export function debugLog(text: unknown, logType: DebugLogType = 'log'): void {
+export async function debugLog(text: unknown, logType: DebugLogType = 'log'): Promise<void> {
 	const logFunc = console[logType];
+	const loggingEnabled = await getGlobalOption('debugLoggingEnabled');
```
inverse, line 771:
```suggestion
	const loggingEnabled = await getGlobalOption(Options.DEBUG_LOGGING_ENABLED);
```

But we would need to import `import * as Options from '@/core/storage/options';`

src/core/content/util.ts
```diff
@@ -703,6 +703,25 @@ export async function getOption(
 	return false;
 }
 
+/**
+ * Read global option from storage.
+ * @param key - Option key
+ * @returns Option value
+ */
+/* istanbul ignore next */
+export async function getGlobalOption(
```
inverse, line 712: @yayuyokitano do we have a function to get at global options? if not would it be worth moving this to `options.ts`?

inverse (approved):
Nice one :clap: 

Thanks for contributing! 

Will wait to see what another contributor has to say on the two open questions before merging :+1:

src/core/content/util.ts
blimpage: Done now in a1101ef

src/core/content/util.ts
blimpage: There is an existing `getOption()` function in `options.ts`, but when I tried it out earlier I was running into issues when trying to import and use it - it kept throwing `Uncaught ReferenceError: Cannot access 'OPTIONS' before initialization`.

I've just tried again in a1101ef with an async `import` and it seems to work just fine! 🎉 so I've deleted the new `getGlobalOption()` function I had added. Let me know if you want me to squash those commits together.

src/core/content/util.ts
```diff
@@ -747,7 +751,14 @@ export function injectScriptIntoDocument(scriptUrl: string): void {
  * @param logType - Log type
  */
 /* istanbul ignore next */
-export function debugLog(text: unknown, logType: DebugLogType = 'log'): void {
+export async function debugLog(text: unknown, logType: DebugLogType = 'log'): Promise<void> {
```
yayuyokitano, line 754: This should not be async.
By being async, the order of debug logs will not be the same as the order of the extension calling debugLog, which has the potential to cause confusion when debugging.

Make the function synchronous, and instead of doing the check inside the function, I think I'd suggest creating a queue class, something like (and im typing this on my phone so this probably does not even compile much less be good code, but you get the point):
```
class Queue {
  private queue: [any, DebugLogType][] = [];
  private isActive = true;
  private shouldPrint = Options.getOption(Options.DEBUG_LOGGING_ENABLED);
  
  private async start() {
    if (this.isActive) return;
    this.isActive = false;
    for (const i = 0; i < 100 && this.queue.length > 0; i++) { // i imagine some fixed upper limit is prolly good idea
      const cur = this.queue.shift();
      if (await this.shouldPrint) {
        // print
      }
    }
  }
  
  public push(text: unknown, logType: DebugLogType) {
    queue.push([text, logType]);
    this.start();
  }
}
const queue = new Queue();

export function debugLog(text:unknown, logType....
  queue.push(text, logType);
```
This approach also helps reduce IO access a bit, not that I expect that to be a big issue anyway though, but I don't see a solid reason not to.

src/core/content/util.ts
```diff
@@ -14,6 +14,10 @@ const BrowserStorage = (async () => {
 	return import('@/core/storage/browser-storage');
 })();
 
+const Options = (async () => {
```
yayuyokitano, line 17: This is really smelly. Are you 100% sure you can't just do an `import * as Options from '@/core/storage/options'` here?
I saw you described meeting some issues before, but the message you described felt like if you were to use import type instead of a normal import, or otherwise botched the import.

yayuyokitano (changes requested):
Thank you for contribution, I agree that this is a good change.
However, I do not like some of the details of the approach, I have detailed the issues that I want to see changed in the review.
I am also very skeptical of adding a new option, but I think it is acceptable for now and does not need changing in this PR.

src/core/content/util.ts
yayuyokitano: The queue should of course also get a more descriptive name

blimpage:
@yayuyokitano Thank you for the feedback! I'll have a go at implementing those changes over the next few days.

src/core/content/util.ts
blimpage: Here's a commit where I've tried to do exactly that and it's causing the error I described earlier: https://github.com/blimpage/web-scrobbler/commit/fb3ccd4 I've put screenshots of the errors I'm seeing in a comment on the commit. I don't _think_ I'm doing anything wrong there, what do you reckon?

src/core/content/util.ts
blimpage: @yayuyokitano I've added a queue for log messages now, based on your suggested implementation: 83336ce

src/core/content/util.ts
yayuyokitano: Ah yeah I realize now this is within content. Yeah, this is fine for now. I want a better long term solution but this will do for the time being.

src/core/content/util.ts
```diff
@@ -745,28 +745,70 @@ export function injectScriptIntoDocument(scriptUrl: string): void {
...
+      this.isActive = false;
+    } catch(err) {
+      this.isActive = false;
+    }
+  }
+
+  /**
+   * Check if the user has enabled debug logging in the extension options.
+   */
+  private async shouldPrint(): Promise<boolean> {
```
yayuyokitano, line 796: I'm not entirely sure, but I'm leaning towards preferring this as a variable rather than function. So something like this:

```ts
private shouldPrint = Options.then((awaitedOptions) => {
		return Boolean(
			 awaitedOptions.getOption(awaitedOptions.DEBUG_LOGGING_ENABLED),
		);
});
```
Benefit is it's a lot faster and less demanding than having it as a function as it only actually fetches the option once.
Drawback of course being that if you change the option you will need to refresh the website for it to go through.

yayuyokitano: To add to that, I'm not quite convinced the speed actually matters here, but it might on a slow HDD (and for modern machines there's the whole death by a thousand paper cuts thing)

blimpage: :+1: That seems pretty reasonable, users probably won't be toggling this option very often. I've switched to that here and it feels much nicer: edc6372

yayuyokitano (approved):
LGTM

yayuyokitano:
Thank you for your contribution🎉

blimpage:
Thank you!

PR #4068, opened 2023-08-29
Add Stationhead connector

### LINKED ISSUE #2879: Requesting to add STATIONHEAD on scrobbling connector

fikimaul:
Stationhead now has web player. The web player can connect to apple music or spotify. With this connector user can scrobble when Stationhead connected to Apple Music or when lastfm not connected to spotify account.

The app url :
https://www.stationhead.com/ or https://app.stationhead.com/

Fix #2879

**Additional context**
Testing
image

Thank You


src/connectors/stationhead.ts
```diff
@@ -0,0 +1,4 @@
+export {};
+
+Connector.useMediaSessionApi();
+Connector.playerSelector = '#root>div>div:nth-child(2)>div>div>div>div>div:nth-child(2)>div:nth-child(2)>div';
```
inverse, line 4: Is there not any class or ID on the element to make this less brittle to changes? 

Also please add a new line after :)

fikimaul: The website only have generated class from library they use, when website have an update will regenerate the class. 
image

Done add new line on end of file.

inverse (approved):
Don't have an account so didn't test but seems better than nothing. 

Hopefully they don't change they layout too much as it seems quite brittle

Issue #4018, opened 2023-08-09
Beatbump connector not working


xRealNeon:
**Describe the bug**
The Beatbump connector isn't working because the domain has changed from .ml to .io

**How to reproduce**
1. Go [beatbump.io](https://beatbump.io) and play a song

**Expected behavior**
Should scrobble the music

inverse:
Resolved in #4069

PR #4069, opened 2023-08-31
Fix beatbump connector by updating url

ixhbinphoenix:
Since beatbump changed their domain from `beatbump.ml` to `beatbump.io` the connector was no longer picking up on it's own and needed an additional URL pattern to be added to the settings. The issue was already noticed by another user in #4018. This PR fixes this issue by replacing the default pattern `*://beatbump.ml/*` with `*://beatbump.io/*`.

**Additional context**
I have already run linter, test suite, and tested it myself as a temporary add-on in Firefox

### DIFF
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -1926,7 +1926,7 @@ export default <ConnectorMeta[]>[
 	},
 	{
 		label: 'Beatbump',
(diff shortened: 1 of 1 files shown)


inverse:
Thanks!

PR #4070, opened 2023-08-31
Add Gensokyo Radio connector

jkap:
Added a new connector and injected script for Gensokyo Radio (https://gensokyoradio.net)

**Additional context**
Gensokyo Radio's player is a bit weird; while all metadata can easily be grabbed via HTML selectors, the actual playback state isn't displayed anywhere and can only be grabbed off of the audio element, which is never inserted into the DOM. We're able to grab the element off of a global `window.audio`, but this requires an injected script. Seems more complicated than it needs to be.

Manually tested against a personal last.fm account in Firefox 117 and Chrome 116, both on Windows 11.


inverse (approved):
Nice one! thanks!

PR #4080, opened 2023-09-06
Add soundcloud mobile support

yayuyokitano:
There are substantial differences between soundcloud desktop and web mobile.
This should enable web scrobbler to work on both.


inverse (approved):
Nice improvements!

PR #4082, opened 2023-09-06
Fix placeholder on about changelog text

inverse:

I noticed we were not substituting it correctly.

Wonder what would be the best way to handle this across all translations? I mean we could update the source files in code but I am not sure how that'll be when pulling new strings down or resolving those on the remote side? 

**Additional context**

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -376,7 +376,7 @@
         "description": "'About' section"
     },
     "aboutChangelog": {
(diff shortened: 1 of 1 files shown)


yayuyokitano (approved):
Well spotted!
Personally I think translation changes should exclusively be handled in transifex.

inverse:
Agreed - push the change up so hopefully it'll be handled. Might try and take a look to see what can be done - but I remember previously it was quite confusing UI for these sorts of HTML in text.

yayuyokitano:
Worst case we can download the dumps and modify them on our end IIRC

PR #4083, opened 2023-09-06
add technobase.fm connector

Haiyn:
Added a connector for technobase.fm and other We aRe oNe radio stations. 

**Additional context**
- technobase.fm, housebase.fm etc. all work the same, so different connectors are not needed despite it being different domains. 
- There is a popup player, but it is rather buggy and consistently fails to display the artist and refresh on track changes. Therefore I've ignored it in the implementation.
- Mobile and desktop DOM are identical.
- I've chosen to use technobase.fm as the blanket naming for all adjacent radios since it is by far the most well known one.

### DIFF
--- a/src/connectors/technobase.fm.ts
+++ b/src/connectors/technobase.fm.ts
@@ -1,0 +1,19 @@
+export {};
+
+const filter = MetadataFilter.createFilter({
+	track: replaceWith,
+});
+Connector.applyFilter(filter);
+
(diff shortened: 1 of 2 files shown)


yayuyokitano (approved):
Seems sane to me. I will trust that it works.

calvinthefreak:
Related to #4663 -> WeAreOne.fm is the collective site of all the Stations from the Technobase group.

PR #4085, opened 2023-09-07
Allow multiple webhooks and general improvements

yayuyokitano:
Allows the webhook scrobbler to have multiple webhooks set up for multiple different applications.
Also makes some small changes to make UX consistent.
Also fixes some random type errors and such from elsewhere in the code.
This is pretty scuffed, but it needs to be pushed with next update imo, and we have an urgent hotfix pending.
I don't think this should impact functionality of existing scrobblers.

### DIFF
--- a/src/core/scrobbler/webhook-scrobbler.ts
+++ b/src/core/scrobbler/webhook-scrobbler.ts
@@ -44,13 +44,10 @@ export default class WebhookScrobbler extends BaseScrobbler<'Webhook'> {
 
 	/** @override */
 	getSession(): Promise<SessionData> {
-		console.log('getting session');
 		if (!this.arrayProperties || this.arrayProperties.length === 0) {
-			console.log('fail');
 			return Promise.reject('');
 		}
 		// Webhook connection doesn't have a session.
-		console.log('success');
 		return Promise.resolve({ sessionID: 'webhook' });
 	}
 
@@ -107,6 +104,7 @@ export default class WebhookScrobbler extends BaseScrobbler<'Webhook'> {
 			);
 			for (const response of responses) {
 				if (response.status !== 200) {
+					this.debugLog(`Error in ${response.url}.`, 'error');
 					return ServiceCallResult.ERROR_OTHER;
 				}
 			}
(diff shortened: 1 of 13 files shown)


src/core/scrobbler/webhook-scrobbler.ts
```diff
@@ -100,15 +94,21 @@ export default class WebhookScrobbler extends BaseScrobbler<'Webhook'> {
...
 		try {
-			response = await timeoutPromise(timeout, promise);
-			if (response.status !== 200) {
-				return ServiceCallResult.ERROR_OTHER;
+			const responses = await timeoutPromise(
+				timeout,
+				Promise.all(promises),
+			);
+			for (const response of responses) {
+				if (response.status !== 200) {
```
inverse, line 109: Should we add a log line here to aid in debugging what requests fail or do you think the network tab is clear enough?

src/ui/options/components/accounts.tsx
```diff
@@ -36,7 +37,18 @@ const scrobblerPropertiesMap = {
...
 	},
+};
+
+/**
+ * Properties associated with each scrobbler that are stored as an array with items add-and-removable, and the type of input to use to edit them.
+ */
+const scrobblerArrayPropertiesMap = {
 	Webhook: {
+		applicationName: {
+			type: 'text',
```
inverse, line 48: Should we change the `userApiUrl.type` to `url`?

inverse (approved):
Seems legit when I did some minimal testing with https://putsreq.com/.

Does it take care of migration from the previous model?

src/core/scrobbler/webhook-scrobbler.ts
yayuyokitano: Agreed, added.

src/ui/options/components/accounts.tsx
yayuyokitano: Long term I think its very much worth considering (as mentioned I want to completely overhaul the accounts UX in the near future). In the short term I think we should just keep it consistent with the userApiUrl type of the other scrobblers.

yayuyokitano:
Guess I'm merging this and releasing tomorrow morning.
Its urgent, but I don't want it being pushed and something breaking while I'm asleep.

yayuyokitano:
> Does it take care of migration from the previous model?

I forgor to reply to this💀
It does not. However it is a non-issue, because webhook has not been released yet, and the rest of the data models are unchanged.
It’s actually a major reason why I want no cersion push before this is through, despite the needed hotfix.

PR #4094, opened 2023-09-11
Fix safari listenbrainz login

yayuyokitano:
Safari is stripping crucial information from the listenbrainz request everywhere except background script.
Work around this by pushing login requests to background script.
Chrome login is still a little unstable (no domparser API in background script) but handling it probably means adding an entire dom parsing library or offscreen document jank. Don't think the time figuring that out is worth it when its a fairly minor issue compared to the inverse.

### DIFF
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -45,6 +45,7 @@ import {
 	toggleLove,
 } from './scrobble';
 import scrobbleService from '../object/scrobble-service';
+import { fetchListenBrainzProfile } from '@/util/util';
 
 const disabledTabs = BrowserStorage.getStorage(BrowserStorage.DISABLED_TABS);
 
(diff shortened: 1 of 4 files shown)


inverse (approved):
Tested in chrome and works :+1: 

Interesting findings - feels like this sort of logic messes a bit with your head about where it's being invoked and what's available :dagger:

yayuyokitano:
> Tested in chrome and works 👍
> 
> Interesting findings - feels like this sort of logic messes a bit with your head about where it's being invoked and what's available 🗡️

If its being called in the options (happens whenever you tab into the options page) it is being called from content.
Else, when it's called during startup or during scrobbling, it is being called from background.

Someday we (or rather, I, I guess) should probably document the details of this stuff

PR #4095, opened 2023-09-11
Add pipeline to test for typescript errors

yayuyokitano:
Vite doesn't really concern itself with type checking, just transforming.
Add a separate pipeline that checks for typescript errors within the test action.
Draft for now as I test it out a bit with commits to check it's working as intended.

EDIT: Also fixed some hidden ts errors hopefully not breaking the translate pipeline.

EDIT 2: Undrafted, seems to work as intended.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,5 +20,8 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Check for typescript errors
+        run: npm run checkts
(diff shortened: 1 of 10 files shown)


yayuyokitano:
@inverse Do you think the change to translations-prepare.ts will impact anything? Typescript didnt love top level await because its ES2022. Idk if we're ready to push up to that point or not, would have to review browser support.

Issue #4071, opened 2023-09-01
Turntable.fm not scrobbling


fuzz-lightyear11:
Turntable.fm has changed its url to www.deepcut.fm and the web scrobbler no longer supports the site. Please fix!

onchyophoran:
wanted to second this request, the deepcut.fm version of the site improves some playback, so it'd be very much appreciated if scrobbling worked from it as well as it has been for TTFM

many thanks

jbwharris:
> Turntable.fm has changed its url to [www.deepcut.fm](http://www.deepcut.fm/) and the web scrobbler no longer supports the site. Please fix!

It does still work if you use the old URL, I think it just requires adding the new URL to the list of valid URLs

jaccarmac:
Is this fixed with your pull request @jbwharris ?

jbwharris:
@jaccarmac Yup, fully functioning with the renamed connector. This issue can be closed.

![image](https://github.com/web-scrobbler/web-scrobbler/assets/2962327/875f567b-cb6f-4327-9d0e-eebd647e9c1a)

onchyophoran:
> @jaccarmac Yup, fully functioning with the renamed connector. This issue can be closed.
> 
> ![image](https://user-images.githubusercontent.com/2962327/267746327-875f567b-cb6f-4327-9d0e-eebd647e9c1a.png)

they broke it again.
both turntable.fm and deepcut.fm now redirect to **deep-cut.fm**, and the extension no longer recognizes it

jbwharris:
@onchyophoran I just pushed a fix for this.

jaccarmac:
Thanks @jbwharris

onchyophoran:
big thanks

PR #4099, opened 2023-09-11
Fix turntable.fm connector

jbwharris:
- This service recently rebranded as deepcut.fm, so I have updated the connector name and added the new URL to the connectors.ts
- Related to #4071

**Additional context**


inverse:
Nice one - thanks!

PR #4113, opened 2023-09-18
Add TuneGenie Support

jbwharris:
I created 2 new connectors to support radio stations that use [TuneGenie](https://tunegenie.com/) for streaming. In following the model I'd seen for Bandcamp and YouTube connectors, where they have a separate connector for the main website and sites that use an embedded player. 

## Website Connector
There are a ton of sites that use this player straight up, just linking to this player page from their sites. I found the player a bit finicky from station to station though. Some would refresh properly and others wouldn't. Not quite sure if it's an implementation issue on the stations side or not.  Sites tested against [WERS](http://wers.tunegenie.com/), [MYVRadio](http://mvyradio.tunegenie.com/#listenlive) and [KROQ](http://kroq.tunegenie.com/)
image

## Embed Connector
Some sites have a more advanced player embedded in their website. The tough thing is identifying just what sites are using this embed. The few I tested against are [WERS](https://wers.org), [KXT](https://kxt.org) and [WNXP](https://wnxp.org/). The cool thing I found with this implementation is that it functions based on the embed URL, so there's no need to list all the different URLs that it works on, it just connects through the iframe loading. 
image

### DIFF
--- a/src/connectors/tunegenie.ts
+++ b/src/connectors/tunegenie.ts
@@ -1,0 +1,8 @@
+export {};
+
+Connector.playerSelector = '#back_matte';
+Connector.trackSelector = 'ul.currentonair li.slot:first-child .song';
+Connector.artistSelector =
+	'.currentonair > li.slot:first-child .left div:not(.song)';
+Connector.trackArtSelector = '.currentonair > li.slot:first-child .disc img';
+Connector.isPlaying = () => Util.hasElementClass('.play.btn i', 'icon-stop');
--- a/src/connectors/tunegenie-embed.ts
+++ b/src/connectors/tunegenie-embed.ts
@@ -1,0 +1,13 @@
+export {};
+
+Connector.playerSelector = '#tgmp';
+Connector.artistSelector = '.info .artist';
+Connector.trackSelector = '.info .song';
+Connector.trackArtSelector = '.cover img';
+
+Connector.isPlaying = () => {
+	return (
+		Util.getAttrFromSelectors('.control a svg path', 'd') ===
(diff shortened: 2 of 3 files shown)


src/connectors/tunegenie.ts
```diff
@@ -0,0 +1,8 @@
+export {};
+
+Connector.playerSelector = '#back_matte';
+Connector.trackSelector = 'ul.currentonair li.slot:first-child .song';
+Connector.artistSelector =
+	'.currentonair > li.slot:first-child .left div:not(.song)';
+Connector.trackArtSelector = '.currentonair > li.slot:first-child .disc img';
+Connector.isPlaying = () => Util.hasElementClass('.play.btn i', 'icon-stop');
```
yayuyokitano, line 8:
```suggestion
Connector.playButtonSelector = '.play.btn .icon-play';
```
I don't think the original here will consistently work between live and playing back playlists. I think this should do the trick, but is there some issue I didn't spot? Disclaimer I have not run this.

yayuyokitano (commented):
Thank you for the contribution! The isPlaying detection seems a little sketchy, could this be more reliable, or have I missed something?

src/connectors/tunegenie.ts
jbwharris: @yayuyokitano I changed both is playing implementations. The tunegenie.ts to your suggestion, and made the second one that relies on the SVG to check if it's not the play button code instead of if it is the stop button. This allowed it to handle the podcast playback. Unfortunately the only thing that changes on play/pause is the SVG code and a couple inline styles, so not really any more reliable. It can pickup the podcasts, but I won't say it's great as they're usually 60 minute shows that aren't broken up into individual songs. It can also pickup the 30 second song previews, but it won't scrobble them due to the length.

yayuyokitano: Cool, will merge this then, the player finickyness can always be worked around better at a later time if someone figures something out

PR #4115, opened 2023-09-19
Add save button to local properties

yayuyokitano:
This button doesn't actually do anything as the extension autosaves.
However, I think it's a good addition for the sake of user feedback.
The user probably only interacts with this once or twice, and when they do they are brand new, let's be as clear as possible.


yayuyokitano:
btn

inverse (approved):
Nice!

PR #4117, opened 2023-09-20
Improve bracket matching

yayuyokitano:
New bracket just dropped.
Also makes the pattern matching of brackets less greedy.


inverse (approved):
Wonder why this stuff isn't baked into the metadata filter :thinking:

PR #4118, opened 2023-09-20
Add instant local login

yayuyokitano:
Currently, local signins require reload of extension.
This makes that not happen anymore.
Just pings off a message to the background scrobbler telling it to reinitialize props whenever user modifies props.
Should significantly improve onboarding for those using local scrobbling/webhooks.

### DIFF
--- a/src/ui/options/components/accounts.tsx
+++ b/src/ui/options/components/accounts.tsx
@@ -8,6 +8,7 @@ import styles from './components.module.scss';
 import browser from 'webextension-polyfill';
 import Delete from '@suid/icons-material/DeleteOutlined';
 import { debugLog } from '@/util/util';
+import { sendContentMessage } from '@/util/communication';
 
 /**
  * Properties associated with each scrobbler, and the input type to use for the user to edit them.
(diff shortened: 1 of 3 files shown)


src/ui/options/components/accounts.tsx
```diff
@@ -212,7 +213,14 @@ function Properties(props: { scrobbler: Scrobbler }) {
 											data = {};
 										}
 										data[typedKey] = e.currentTarget.value;
-										scrobbler.applyUserProperties(data);
+										scrobbler
+											.applyUserProperties(data)
+											.then(() => {
+												sendContentMessage({
+													type: 'updateScrobblerProperties',
+													payload: void 0,
```
inverse, line 221: What's the significance of `void 0`? So we can send a message without a payload?

inverse (approved):

src/ui/options/components/accounts.tsx
yayuyokitano: Its the same as `undefined`, old habit. Can change if you want.

inverse: It's fine :+1:

PR #4120, opened 2023-09-20
Add support for webhook website auth

yayuyokitano:
makes the user experience of connecting webhook a dream. It just works(TM)
Depends on #4118 which must be merged first.

In action gif:
login

### DIFF
--- a/src/connectors/webhook-auth-dom-inject.ts
+++ b/src/connectors/webhook-auth-dom-inject.ts
@@ -1,0 +1,27 @@
+enum State {
+	Waiting = 'WAITING',
+	Rejected = 'REJECTED',
+	Accepted = 'ACCEPTED',
+	Confirmed = 'CONFIRMED',
+}
+
+declare interface Window {
+	wsState: () => State;
+	setWSState: (newState: State) => void;
+}
+
+setInterval(() => {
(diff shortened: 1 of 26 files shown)


src/connectors/webhook-auth-dom-inject.ts
```diff
@@ -0,0 +1,27 @@
...
+
+setInterval(() => {
+	if (window.wsState() === State.Accepted) {
+		const search = new URLSearchParams(window.location.search);
+		window.postMessage(
+			{
+				sender: 'web-scrobbler',
+				type: 'confirmLogin',
+				applicationName: search.get('applicationName'),
+				userApiUrl: search.get('userApiUrl'),
```
inverse, line 21: Do you think we should do any validation of input?  e.g. if it's a URL or do you think it's fine?

inverse (approved):
Very nice :+1: Will def improve the UX of onboarding with webhook based scrobbling :)

src/connectors/webhook-auth-dom-inject.ts
yayuyokitano: I think it's fine. get is always a string, and the webhook handles any string. The user is also shown what the URL is when accepting.

PR #4139, opened 2023-09-25
Add connector for RTL+ Musik

opcode81:
Added connector for _RTL+ Musik_

**Additional context**
_RTL+ Musik_ is a new German music streaming service by media conglomerate RTL Group, which is powered by Deezer.

### DIFF
--- a/src/connectors/rtl-plus-musik.ts
+++ b/src/connectors/rtl-plus-musik.ts
@@ -1,0 +1,20 @@
+export {};
+
+Connector.playerSelector = 'plus-player-container';
+
+Connector.trackSelector = 'plus-player-container .album';
+
+Connector.artistSelector = 'plus-player-container .artist';
+
+Connector.currentTimeSelector = 'plus-player-container music-progress .current-time';
+
+Connector.remainingTimeSelector = 'plus-player-container music-progress .duration';
+
+Connector.isPlaying = () => {
+    const titleElements = Util.queryElements('plus-player-container .play-icon > svg > title');
+    if (titleElements && titleElements[0].textContent) {
+        const isPaused = titleElements[0].textContent.indexOf('Abspielen') >= 0;
+        return !isPaused;
+    }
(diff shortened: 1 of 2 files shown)


src/connectors/rtl-plus-musik.ts
```diff
@@ -0,0 +1,20 @@
+export {};
+
+Connector.playerSelector = 'plus-player-container';
+
+Connector.trackSelector = 'plus-player-container .album';
```
yayuyokitano, line 5: This seems very odd, so just to be sure this is not a typo, can you confirm that the track title is in fact contained in an element of class album?

yayuyokitano (commented):
Tested this out on the radio channels since that's what's available without paying, and seems to be completely nonfunctional there. Not a blocker for merging IMO though, as some functionality is better than none, but if you want to add it it would be cool. Other than that, I've one thing I want to confirm that I left in review comment.

src/connectors/rtl-plus-musik.ts
opcode81: Confirm, it's not a typo. The track name is in an element of class `album` because it *links to* the album the track is part of.

opcode81:
> Tested this out on the radio channels since that's what's available without paying, and seems to be completely nonfunctional there. Not a blocker for merging IMO though, as some functionality is better than none, but if you want to add it it would be cool.

I'll see what I can do. The stuff I implemented works fine when playing either any user-selected stuff or the curated playlists, but it does not work for radio (which I had never even used before).

opcode81:
@yayuyokitano I have now added radio stream support; tested that it also works when not logged in.

src/connectors/rtl-plus-musik.ts
yayuyokitano: Understood, just wanted to make sure just in case it was a mistake 😊

yayuyokitano:

PR #4140, opened 2023-09-26
Filter sports broadcasts from SiriusXM

dasevilla:
This prevents scrobling most sports broadcasts

Examples scrobles to avoid:
https://www.last.fm/search/tracks?q=bdcast


yayuyokitano:

PR #4141, opened 2023-09-26
Remove "Sign in" button for accounts not using it

yayuyokitano:
This just adds noise, and makes things more confusing for users.

Before:
image

After:
image

It still shows the signed in label:
image


inverse (approved):
Nice UX improvement :clap:

PR #4143, opened 2023-09-27
Fix warnings

yayuyokitano:
Theres a fair number of warnings right now, pretty much all relating to reactivity.
This silences all that.
Marking as draft, as  this still needs testing to ensure functionality is still equivalent.
Some of this is admittedly a bit cursed.

### DIFF
--- a/src/ui/options/components/connector-override.tsx
+++ b/src/ui/options/components/connector-override.tsx
@@ -118,7 +118,6 @@ function ConnectorOptions() {
  * The connector override options for one connector
  */
 function ConnectorOption(props: { connector: ConnectorMeta }) {
-	const { connector } = props;
 	const [ref, setRef] = createSignal<HTMLDetailsElement>();
 	const [active, setActive] = createSignal(false);
 	createEffect(() => {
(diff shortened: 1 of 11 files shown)


yayuyokitano:
Undrafting, been testing a bit and seem unable to find any regressions in affected bits

src/ui/options/components/connector-override.tsx
```diff
@@ -118,7 +118,6 @@ function ConnectorOptions() {
  * The connector override options for one connector
  */
 function ConnectorOption(props: { connector: ConnectorMeta }) {
-	const { connector } = props;
```
inverse, line 121: why this change?

inverse (approved):
Tested and seems legit - one question about why :+1:

src/ui/options/components/connector-override.tsx
yayuyokitano: By design, SolidJS components only ever run once, unlike react/similar where it constantly reruns every change.
This has advantages, but one disadvantage is that if you assign a static variable, it will keep that value, it's not reactive.
props, on the other hand, is reactive, so it updates.
This doesn't really create any immediate issues for us in the previous code, but the linter throws warnings about it because it could create issues if there was ever a need to update values.

[quick example I whipped up illustrating it, you can see how the destructured one is dead](https://playground.solidjs.com/anonymous/eae9078d-ddbb-4908-abb0-cfa7c8142bee)

PR #4145, opened 2023-09-27
Add ignored state and overhaul control flow

yayuyokitano:
Does some invisible overhauling of control flow to improve type checking a bit and just generally be a bit less of a mess.
Also adds ignored state popup, as this was necessary to satisfy type checker now.
Also, fallback is now unsupported website, not empty element.
This is relevant on safari on mac, as new tab would trigger fallback.

### DIFF
--- a/src/ui/popup/main.tsx
+++ b/src/ui/popup/main.tsx
@@ -1,12 +1,10 @@
-import { render } from 'solid-js/web';
+import { Dynamic, render } from 'solid-js/web';
 import './popup.module.scss';
 import Unsupported from './unsupported';
 import * as ControllerMode from '@/core/object/controller/controller-mode';
 import { initializeThemes } from '@/theme/themes';
 import '@/theme/themes.scss';
 import {
-	Match,
-	Switch,
 	createResource,
 	Show,
 	createMemo,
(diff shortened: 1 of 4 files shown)


src/ui/popup/main.tsx
```diff
@@ -86,44 +85,43 @@ function Popup() {
...
+		[ControllerMode.Disabled]: () => <Disabled />,
+		[ControllerMode.Err]: () => <Err />,
+		[ControllerMode.Playing]: () => <NowPlaying tab={tab} />,
+		[ControllerMode.Skipped]: () => <NowPlaying tab={tab} />,
+		[ControllerMode.Scrobbled]: () => <NowPlaying tab={tab} />,
+		[ControllerMode.Disallowed]: () => <Disallowed tab={tab} />,
+		[ControllerMode.Unknown]: () => <Edit tab={tab} />,
+		[ControllerMode.Unsupported]: () => <Unsupported />,
+		[ControllerMode.Ignored]: () => <Ignored tab={tab} />,
+		[ControllerMode.Loading]: () => <></>,
```
inverse, line 99: What's the significance of this one? e.g. the meaning behind the empty tag?

inverse (approved):
Nice additions

PR #4148, opened 2023-09-27
Add tomorrowland connector

### LINKED ISSUE #1917: OneWorld Tomorrowland Radio Web Player
Tomorrowland started their own radio station yesterday available at [One World Radio - Tomorrowland](https://oneworldradio.tomorrowland.com).
It shows the current song playing, so it should be relatively easy to add this page to the scrobbler..

yayuyokitano:
Not 100% sure about how to handle the metadata on this one, whether there should be some filter added.
Here is what it looks like currently:
image
First part of track is the track name.
then theres (played by \<performing artist\> \<venue\> \<year\>) (no way to separate artist from venue reliably here except a massive library of venues thatll need to be updated).
then finally artist field is whoever *originally* made the song being performed.

It might be simple enough to say "ok so remove the entire parenthesis", but not really, as often these are not being performed by original artists, and are often even mashups of multiple songs by multiple artists.

Sometimes, there is no (played by) part at all, those cases seem simple enough.

Closes #1917


inverse (approved):
I think this should be fine for now :)

Awesome investigation!

PR #4169, opened 2023-10-01
add khinsider connector

duckfromdiscord:
Can detect the track name, and also album art if any is given on the page. Able to detect album name. Attempts to detect artist from the description, first looking for a composer and then the developers of the game. In some cases, the "developer" _is_ the composer instead of the actual game dev, like on the Minecraft OST. Also able to detect duration, position, and play status.


yayuyokitano (changes requested):
Please fix linting on the PR before merge. Other than that, the PR seems good.

yayuyokitano (approved):
LGTM! Thank you for the contribution 🚀

PR #4172, opened 2023-10-01
Update vite-plugin-conditional-compiler to 0.2.0

yayuyokitano:
This needed some change as the update deprecated #v-ifndef, which we were using


inverse (approved):
nice :)

PR #4177, opened 2023-10-03
Fix iramanusantara multiple artists scrobbling

fikimaul:
Iramanusantara can have multiple artist on their song metadata. Before this change selector only get first artist with the sparator. 

Before
before

After
after

**Additional context**
Link for test https://www.iramanusantara.org/release/5515


yayuyokitano:

PR #4179, opened 2023-10-03
Add button to block specific youtube channels

### LINKED ISSUE #2443: Option to not scrobble anything from a youtube channel

yayuyokitano:
Recreated for better diffs/commits.
Adds a new context menu item to disable/enable scrobbling youtube channel.

image

Uses a new connector function Connector.getChannelID and a new connectors.ts item "usesBlocklist".
While this only works for youtube for now, any connector that implements these two functions, the functionality will work for.
I think we should accept such PRs if for the given site they are meaningfully different from just being an artist tag block (which is a separate issue and should be handled separately)

Draft for now:
- [x] Document new functions
- [x] Make hitting the button refresh state (currently, you must refresh for the button press to actually come into effect)
- [x] Add some way to manage blocklisted items in settings
- [x] Test

Closes #2443


yayuyokitano:
Have added way to edit blocklist now in connector settings (only under youtube for now).
Logic is largely based on our edits management section and seems solid.
IMPORTANT: this is untested for cases where users have channels blocked on multiple websites, as only youtube is currently implemented. Once another website is added, more testing must be done into the case where there are edits for multiple sites, but the logic seems sound.

image

yayuyokitano:
Also redid control flow of options modal

src/core/content/connector.ts
```diff
@@ -144,6 +144,14 @@ export default class BaseConnector {
 		alignItems: 'center',
 	};
 
+	/**
+	 * Function that gets a unique ID for channel/user blocklist.
+	 *
+	 * Has to be specified if usesBlocklist is set to true in connectors.ts.
+	 * If connectors.ts does not have usesBlocklist set to true, this should be null.
+	 */
+	public getChannelID: (() => string | null | undefined) | null = null;
```
inverse, line 153: Should we call it `getChannelId` to be consistent? I know we have `getUniqueID` but this feels like a violation :/

src/core/storage/blocklist.ts
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	public async shouldScrobbleChannel(
+		id: string | undefined | null,
+	): Promise<boolean> {
+		if (!id) {
+			return true;
+		}
+		await this.isReady;
+		const data = await this.storage.get();
+		if (!data || !(this.connectorId in data) || !data[this.connectorId]) {
```
inverse, line 89: Is it worth moving this data fetching to a common function as it's used in 3 places or do you think that'll be more overhead

inverse (approved):
Awesome :clap:

src/core/storage/blocklist.ts
yayuyokitano: The data fetching is only actually the same in addToBlocklist and removeFromBlocklist, the other functions handle it in subtly different ways. I think this is fine.

src/core/content/connector.ts
yayuyokitano: Done

src/core/background/util.ts
```diff
@@ -61,6 +65,32 @@ export async function filterInactiveTabs(activeTabs: ManagerTab[]) {
...
+
+/**
+ * Is current channel blocklisted
+ *
+ * @param channelID - ID of the channel to check
+ * @param connector - Details about the connector to check
+ * @returns true if current channel is blocklisted; false otherwise
+ */
+export async function isChannelBlocklisted(
+	channelID: string,
```
inverse, line 87: I guess we should also rename this one to `channelId` too - and then we can merge :tada:

stephanebruckert:
Amazing! Can't wait to use it

PR #4180, opened 2023-10-04
Add connector for WDR

Ceddicedced:
I created a connector for the wdr stations. I used this new connector today while listening to 1Live and couldn't experience any errors.

Added a connector for the WDR radio stations (1Live, WDR2, WDR3, WDR4, WDR5, COSMO, etc.)

I faced some problems with the station WDR5 because the track information was like this: "WDR 5 Quarks - Wissenschaft und mehr mit Sebastian Sonntag". But I were not able to filter strings like this with the following code:
`
Connector.isScrobblingAllowed = () => {
	const artistTrack = Util.getTextFromSelectors(
		Connector.artistTrackSelector,
	);
	const disallowedStrings = [
		'1Live',
		'WDR2',
		'WDR3',
		'WDR4',
		'WDR5',
		'COSMO',
	];

	// Check if the artistTrack includes any of the disallowed strings
	const containsDisallowedString =
		artistTrack !== null &&
		disallowedStrings.some((disallowedString) =>
			artistTrack.includes(disallowedString),
		);

	// Scrobble only if none of the disallowed strings are included
	return !containsDisallowedString;
};
`

Maybe there is a nicer solution...

### DIFF
--- a/src/connectors/wdr.ts
+++ b/src/connectors/wdr.ts
@@ -2,6 +2,10 @@ export {};
 
 // Connector for the WDR radio stations (1Live, WDR2, WDR3, WDR4, WDR5, COSMO, etc.)
 
+const filter = MetadataFilter.createFilter({
+	artist: cleanupArtist,
+});
+
 Connector.playerSelector = '.wdrrCurrentChannels';
 
 Connector.artistTrackSelector = '.wdrrCurrentShowTitleTitle';
@@ -34,3 +38,16 @@ Connector.isScrobblingAllowed = () => {
 	// Scrobble only if none of the disallowed strings are included
 	return !containsDisallowedString;
 };
+
+function cleanupArtist(artist: any) {
+	// Define patterns to find additional artists or features.
+	const patterns = [/ & .*/, / x .*/, / feat\..*/];
+
+	let cleanedArtist = artist;
+
+	patterns.forEach((pattern) => {
+		cleanedArtist = cleanedArtist.replace(pattern, '');
+	});
+
+	return cleanedArtist.trim();
+}
--- a/src/connectors/wdr.ts
+++ b/src/connectors/wdr.ts
@@ -36,7 +36,7 @@ Connector.isScrobblingAllowed = () => {
 	return !containsDisallowedString;
 };
 
-function cleanupArtist(artist: any) {
+function cleanupArtist(artist: string) {
 	// Define patterns to find additional artists or features.
 	const patterns = [/ & .*/, / x .*/, / feat\..*/];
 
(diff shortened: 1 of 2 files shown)


yayuyokitano:

You can run Connector.getArtist() in the isScrobblingAllowed function btw if that makes for a cleaner solution than running the Util.getTextFromSelectors() raw.

As for the WDR5 issue, to me it looks like the issue is you need to add a space to "WDR 5" in the disallowedStrings array.

Also, it looks like you may not have submitted the full script in your PR, please have a look and confirm, it looks a bit interrupted.

Ceddicedced:
Thank you so much for the quick feedback!

Thanks for pointing out the string issues, such a rookie mistake by me...

Should look better now!

src/connectors/wdr.ts
```diff
@@ -0,0 +1,36 @@
+export {};
+
+// Connector for the WDR radio stations (1Live, WDR2, WDR3, WDR4, WDR5, COSMO, etc.)
+
+Connector.playerSelector = '.wdrrCurrentChannels';
+
+Connector.artistTrackSelector = '.wdrrCurrentShowTitleTitle';
+
+Connector.isPlaying = () => {
```
yayuyokitano, line 9:
```suggestion
Connector.pauseButtonSelector = '#playCtrl.playing';
```
Does this work instead?

yayuyokitano (commented):
Overall looks good, just one suggestion to get playing detection more in line with our best practices, if it works.

src/connectors/wdr.ts
Ceddicedced: Yes this is working perfectly fine. Thank you

src/connectors/wdr.ts
```diff
@@ -0,0 +1,53 @@
...
+});
+
+Connector.playerSelector = '.wdrrCurrentChannels';
+
+Connector.artistTrackSelector = '.wdrrCurrentShowTitleTitle';
+
+Connector.pauseButtonSelector = '#playCtrl.playing';
+	const playButton = document.querySelector('#playCtrl');
+	return playButton !== null && playButton.classList.contains('playing');
+};
```
yayuyokitano, line 16:
```suggestion
```

My apologies, bad suggestion from my end, this also needs gone

src/connectors/wdr.ts
```diff
@@ -0,0 +1,50 @@
...
+		artistTrack !== null &&
+		disallowedStrings.some((disallowedString) =>
+			artistTrack.includes(disallowedString),
+		);
+
+	// Scrobble only if none of the disallowed strings are included
+	return !containsDisallowedString;
+};
+
+function cleanupArtist(artist: any) {
```
yayuyokitano, line 39:
```suggestion
function cleanupArtist(artist: string) {
```
I believe this will fix the linter errors

yayuyokitano (approved):
LGTM. Thank you for the contribution!

PR #4186, opened 2023-10-06
Update messages.json

osvaldobarrera:
Changed "Álbum del artista" (tr: artist album) to "Artista del álbum" (tr: album artist).

### DIFF
--- a/src/_locales/es/messages.json
+++ b/src/_locales/es/messages.json
@@ -468,7 +468,7 @@
         "description": "Placeholder of album input"
     },
     "infoAlbumArtistPlaceholder": {
-        "message": "Album del artista (opcional)",
+        "message": "Artista del álbum (opcional)",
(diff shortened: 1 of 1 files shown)


salazarbarrera:
I made a transifex account and sent the suggestion there. Effectively, I can only suggest, not edit. Thanks for your help!

inverse:
Not an expert with Transifex, I don't remember having to approve other folks translations.

yayuyokitano:
> Not an expert with Transifex, I don't remember having to approve other folks translations.
> 
> Def need a way to automate things in this area more.

Its because it’s an edit of an already reviewed translation, rather than a new translation

yayuyokitano:
Late, but this has now been corrected in transifex. Once new transifex merge is done it will come in the version after that. Closing.

PR #4195, opened 2023-10-08
Improve popup cover for non-square covers

yayuyokitano:
Once again took a clue from eggs enhancement suite in web scrobbler design.
Not sure if this is optimal, but it's at least way better than what we have now.

For wide cover:

Before:
image

After:
image

For tall cover:

Before: 
image

After:
image

Square covers look same as before


inverse (approved):
Looks great!

PR #4196, opened 2023-10-08
Add advanced settings

yayuyokitano:
Replaces the old hidden settings from v2 (which was long broken by the time we rolled out v3, but I digress)
Also move enabling/disabling debug logs here, I think it fits.

image

Draft:
- [x] Style the range input better
- [x] Test cross browser (including iOS)

Also, is this current placement good, or should it be moved? Should it be hidden behind a key combination again?


yayuyokitano:
New styling:

Chrome (with hover/focus):

image


Firefox:

image


Safari (macos):

image


Safari (ios):

image

inverse (approved):
Yeah this makes a lot more sense!

PR #4200, opened 2023-10-11
Improve iOS onboarding

yayuyokitano:
Onboarding on iOS was really poor, this is an attempt to improve it a bit.
safari-onboarding


inverse (approved):
Seems legit - unable to test though as I don't have have Apple.

PR #4201, opened 2023-10-11
Fix disallow being too aggressive

yayuyokitano:
Notably, on youtube, if a song took too long to load, it would cause the extension to display disallowed until scrobbling, which no longer happens
Don't think there should be any negative effects from this

### DIFF
--- a/src/core/object/controller/controller.ts
+++ b/src/core/object/controller/controller.ts
@@ -677,6 +677,14 @@ export default class Controller {
 
 		if (isPlayingStateChanged && isPlaying !== void 0) {
 			this.onPlayingStateChanged(isPlaying);
+		} else if (
+			this.mode === ControllerMode.Disallowed &&
+			this.shouldScrobble() &&
+			isPlaying
+		) {
+			// we need to unset disallowed whenever needed.
+			// this is not necessarily tied to pausing/unpausing
+			this.setSongNowPlaying();
 		}
 	}
 


src/core/object/controller/controller.ts
```diff
@@ -677,6 +677,14 @@ export default class Controller {
 
 		if (isPlayingStateChanged && isPlaying !== void 0) {
 			this.onPlayingStateChanged(isPlaying);
+		} else if (
+			this.mode === ControllerMode.Disallowed &&
```
inverse, line 681: How does this one work?   seems a bit counter intuitive reading it

yayuyokitano: Its aimed mainly at youtube – where scrobbling is disallowed when currenttime is 0.

basically if we set mode to disallowed, then it will stay disallowed until a different mode is set. Even if the song is no longer disallowed mid-playthrough this doesnt actually happen until halfway through the song when it scrobbles, leading to a mismatch between what user expects and what is displayed.

this line of code basically does a check for if the mode is disallowed and the state no longer actually disallows scrobbling, then it will unset disallowed mode and show playing state.

inverse: Ahhh gotcha! wonder how we can improve the readability on this as it just seems odd to set a function called setSongNowPlaying when checking mode disallowed. 

I can't think of anything so won't block this merging :+1:

inverse (approved):

PR #4203, opened 2023-10-12
Add "Use local instance" button

yayuyokitano:
I have seen people struggle with the listenbrainz signin process before because of the lack of separation between listenbrainz.org and custom instance signin.
This makes that separation clearer.
This only affects listenbrainz currently, but it's not tied to just listenbrainz, just any scrobbler that has both a main instance and a custom instance option.

localsignin


inverse (approved):
Much clearer!

PR #4205, opened 2023-10-13
Fix cover art

yayuyokitano:
Some various issues that were not spotted in initial testing of #4195 
Fixes DOM in editing section and improves iOS styling.


inverse (approved):
pngwing com

Issue #3889, opened 2023-06-12
BBC Sounds connector no longer scrobbling


markwheeler:
**Describe the bug**
The connector for BBC Sounds seems to longer be recognizing songs played.

**How to reproduce**
1. Go to https://www.bbc.co.uk/sounds/play/m001mf0t
2. Click on play
3. Note web scrobbler recognizes the site but does not display any song info.

**Expected behavior**
This site was previously scrobbling like a champ.

**Environment (please complete the following information)**
-   OS: MacOS 13.4
-   Browser: Brave
-   Extension version: 3.0.12

neanias:
I'm also seeing problems with this:

* OS: MacOS 13.4
* Browser: Firefox 114.0.2
* Extension version: 3.0.13

inverse:
Will be in the next release

neanias:
Thanks for the fix! When are you expecting to release a new version?

PR #4226, opened 2023-10-20
BBC Sounds connector fixes

markwheeler:
Fixes to the BBC Sounds connector. This was previously non-functional – I believe the page element it was checking to find out if radio was live or recorded wasn't always loaded when the check took place. Changes:
- Updated isLiveRadio() function to check a more reliable element.
- Adjusted player selectors.
- Fixed function name typo.

**Additional context**
Related issue: #3889


inverse:
Thanks!

PR #4247, opened 2023-10-30
Apply album artist when processing metadata

everlyy:
This makes sure you're scrobbling the right album artist instead of the album artist always being set to the same artist as the current song, which is problematic when you're scrobbling a track that's in an album released as a 'Various Artists' album.

I made sure the album artist is getting put as the album artist when an album is getting autofilled.

**Additional context**
Here's a before and after:
https://imgur.com/a/NA9NPVB

### DIFF
--- a/src/core/object/pipeline/metadata.ts
+++ b/src/core/object/pipeline/metadata.ts
@@ -60,7 +60,9 @@ export async function process(
 			}
 
 			if (!song.getAlbum() || song.flags.isAlbumFetched) {
+				song.processed.albumArtist = songInfo.albumArtist;
 				song.processed.album = songInfo.album;
+				song.noRegex.albumArtist = songInfo.albumArtist;
 				song.noRegex.album = songInfo.album;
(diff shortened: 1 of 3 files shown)


inverse (approved):
Changes seem reasonable to me.

@yayuyokitano what do you think?

yayuyokitano:

There are a couple of issues that will occur from this change.
1. Some sites do not have a unique ID. In this case, a hash is generated from metadata to store edits. In these cases, edits users have set up will no longer work after this change.
2. Unpredicted things could occur with regex edits as well.

For cases where this makes a meaningful improvement, I think it is a fine tradeoff.
However, please ensure that the album artist is not added when it is the same as the artist, as in this case there is no benefit.

jaccarmac:
It's not clear to me if this does what it says or if it's still necessary. Leaning towards a close unless I hear differently.

rob9315:
i think this exists now, added with some other pr, closing

PR #4251, opened 2023-10-31
Fix to LINE Music Web player connector

mmenti:

The LINE Web player contains a span element containing the labels for song name (再生中の曲名) and artist name (アーティスト名) which caused the song title and artist name to be scrobbled incorrectly, as every song and artist name would include these descriptive labels.

**Additional context**


inverse (changes requested):
Could we revert the dependancy changes - Other than that it should be good to go.

inverse:
Thanks!

Issue #4266, opened 2023-11-07
Radio France / FIP scrobbling doesn't seem to work anymore


zas:
**Describe the bug**

**How to reproduce**
1. Go to https://www.radiofrance.fr/fip
2. Click on 'ÉCOUTER FIP'

**Expected behavior**
WebScrobbler add listens

**Screenshots**

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

```Web Scrobbler: Element '#player' is missing```

**Environment (please complete the following information)**
-   OS: Linux Mint 21.2
-   Browser: Firefox
-   Extension version: (e.g. v2.10.0)

**Additional context**

zas:
Quick fix, thanks!

PR #4269, opened 2023-11-08
Fix connector for RadioFrance

Mehavoid:
Closes: #4266


inverse (approved):
Thanks for the contribution!

Issue #4263, opened 2023-11-06
nowyswiat.online connector not working


zaisty:
**Describe the bug**
Radio nowyswiat.online connector not working at all.

**Expected behavior**
Should scrobble titles.

**Environment**
-   OS: Linux Ubuntu MATE
-   Browser: Firefox
-   Extension version: 3.2.0

Issue #3715, opened 2023-05-09
Nowyswiat radio connector request


introkreatik:
**Provide information about website**
Online radio "NowySwiat", https://nowyswiat.online/

**Additional context**
It would be awesome to get this site able to scribble.
Thanks.

PR #4271, opened 2023-11-08
Fix connector for Nowy Swiat Radio

Mehavoid:
Closes: #4263 and #3715


inverse (approved):
Thanks for the contribution!

Issue #4181, opened 2023-10-04
Connector for radio.de


Ceddicedced:
**Provide information about website**
https://www.radio.de/
This is a webplayer for various german radio stations. 

**Additional context**
Sadly I was not able to build a connector by myself because I normally don't work with websites. 

`
Connector.playerSelector = 'div.sc-flv6xd-1.jsJQCr'
Connector.artistTrackSelector = 'div.sc-1253cl8-0.gnhoAj'
`

My first guess was something like this, but it seems kinda cryptic. 

It should be noted that, the playback information contains additionally the station. "Miley Cyrus - Used To Be Young auf 1LIVE" so that  " auf 1LIVE" needs to be extracted.

PR #4295, opened 2023-11-14
new connector: radio.de

carpet-berlin:
Closes https://github.com/web-scrobbler/web-scrobbler/issues/4181

New connector for radio.de, a German site which allows to play from several radio stations.

**Additional context**


src/connectors/radio.de.ts
```diff
@@ -0,0 +1,17 @@
+export {};
+
+Connector.playerSelector = 'div.player-nav-holder';
+
+Connector.artistTrackSelector = 'span';
```
inverse, line 5: Do we need to assign this to the Connector prop? would it make sense to just pass `span` into `Util.getTextFromSelectors`?

inverse (commented):
LGTM aside one comment!

src/connectors/radio.de.ts
carpet-berlin: Right. Simplified it.

Issue #4144, opened 2023-09-27
Radiojavan connector is broken


MasoudRahmani:
### RadioJavan
![image](https://github.com/web-scrobbler/web-scrobbler/assets/5321268/374aa01f-9892-42c4-93c2-fee8cfe264af)

Website **address** is **changed** as it says it is not a supported website.

![image](https://github.com/web-scrobbler/web-scrobbler/assets/5321268/bf250d71-0227-4b3a-b751-cd9d897ec06e)

After i manually add the new address `https://play.radiojavan.com/` It still does not recognize playing song.

![image](https://github.com/web-scrobbler/web-scrobbler/assets/5321268/2dbcfab5-e8bd-49ae-b070-eed4b5e3cf18)

**Expected behavior**
Add new address website, recognize playing song.

**Environment (please complete the following information)**
-   OS: Windows 11
-   Browser: FireFox and chrome

yayuyokitano:
Struggling to work on this, the website is putting hurdles in the way of being able to access any DOM information unless you have premium, and they're declining my card. Might look into getting around it at a later time

MasoudRahmani:
Thank you, It's really disappointing when a streaming service don't understand what is needed for avid listener.

Appreciated for you efforts. ❤️🌹

PR #4298, opened 2023-11-16
Fix RadioJavan connector

Mehavoid:
Closes: #4144 

**Additional context**
They using `disable-devtool`, which adds problem to connector developing.
Script below removes these restrictions.
```js
// ==UserScript==
// @name        RJ: Bypass disable-devtool
// @namespace   rj-bypass-disable-devtool
// @match       https://play.radiojavan.com/*
// @grant       none
// @version     1.0
// @run-at      document-start
// ==/UserScript==

(function () {
  "use strict";

  Object.assign = new Proxy(Object.assign, {
    apply(target, thisArg, argumentsList) {
      const [, obj] = argumentsList;
      if (obj && "isDevToolOpened" in obj) {
        return Reflect.apply(target, thisArg, [
          {},
          { isDevToolOpened: () => false, ignore: () => true },
        ]);
      }
      return Reflect.apply(target, thisArg, argumentsList);
    },
  });
})();


```

### DIFF
--- a/src/connectors/radiojavan.ts
+++ b/src/connectors/radiojavan.ts
@@ -1,22 +1,22 @@
 export {};
 
-Connector.artistSelector = '.mainPanel .artist';
+const PLAY_ICON =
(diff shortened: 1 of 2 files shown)


inverse (approved):
Thanks!

Issue #2350, opened 2020-04-11
Please add support for audius.co


quentincaffeino:
**Provide information about website**
Audius is a decentralized, community-owned audio sharing protocol.
[audius.co](https://audius.co)

**Additional context**
A free trial is available at [audius.co/trending](https://audius.co/trending)

corrupteddrive:
When is the feature going to be added?

quentincaffeino:
Devs of my favorite distro have a saying: it will be out when it's ready.
If it would be in development it would be told so here, as you can see, no plans on it currently.

inverse:
We happy accept contributions to the project so feel free to implement it if you want it done quicker.

fckawe:
Just had a look on the free trial. Shouldn't be very difficult to find the right selectors... but!...: The info that is given within the artist and the title element doesn't seem to be consistent.

For Example:

![image](https://user-images.githubusercontent.com/1059389/157911427-195e2bea-8c73-4a12-8804-b122dda6ca1f.png)

The first entry has "Quality Good Records" as artist and "sebjin - Ring feat. MARS 88". So the title also contains the actual artist. The third entry follows the same rules, I think.
In the second entry, though, the title element contains the title only and the artist can be found in the artist element.

I have no idea how both types of entries could be distinguished by the connector...

PR #4308, opened 2023-11-18
Add Audius connector

Mehavoid:
Closes: #2350 

**Additional context**


src/connectors/audius.ts
```diff
@@ -0,0 +1,16 @@
+export {};
+
+Connector.playerSelector = '._playBarContentWrapper_1sp1c_10';
```
inverse, line 3: These look like they will mutate on each rebuild of the site. Anything else we can use to bind too?

inverse (approved):
Thanks for the contribution!

PR #4310, opened 2023-11-18
Added connector for Radio Student

dinogrgic1:
I added connector for http://www.radiostudent.hr/. This is a student radio in Croatia which embeds songs that were played last with timestamps. Only available information to us is artist, track and timestamp.

**Additional context**
Here is the picture of this element live on website. Website does feature a web player but sadly web player does not contain metadata for scrobbler to function properly so we can only used this latest played element.
image


.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
   "editor.formatOnSave": true,
 
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
inverse, line 6: What's the reason for this change?

src/connectors/radiostudent.ts
```diff
@@ -0,0 +1,5 @@
+export {};
+
+Connector.playerSelector = '#icecast_list';
+Connector.artistTrackSelector = '#icecast_list > ul > li > p';
+Connector.currentTimeSelector = '#icecast_list > ul > li > h3';
```
inverse, line 5: This one is incorrect - `currentTimeSelector` is more for like when you have a player like `00:02 / 04:23`, where `00:02` would be the current play time of a track.

inverse (commented):
Apologies in the delay in getting back to this one.

I think it should probably work without the `currentTimeSelector`. while it would be super basic in the support it'll provide it'll indeed be better than nothing.

src/connectors/radiostudent.ts
dinogrgic1: Thanks for the clarification! I will remove the `currentTimeSelector`.

.vscode/settings.json
dinogrgic1: I had issues with Prettier locally, and changed this by mistake. Reverting it.

inverse (approved):
Thanks for the contribution!

PR #4340, opened 2023-12-02
Added connector for vk-save.com

gasievt:
I added a connector for https://vk-save.com/. This is a website that enables you to listen VK music without logging in to your account.


inverse:
Thanks for the contribution! It will be in the next release :)

PR #4343, opened 2023-12-08
Fix scrobble percent option

yayuyokitano:
An issue caused scrobble percent value to be corrupted if set using the range slider.
Fixes this by parsing integer.
Also ensures graceful fallback for any issues related to corrupted scrobble percent values by falling back to default if it is somehow not a number.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -27,6 +27,12 @@ export const HIDDEN_PLACEHOLDER = '[hidden]';
 export const DEFAULT_SCROBBLE_TIME = 30;
 
 /**
+ * Percentage of track to playback before the track is scrobbled.
+ * This value is used only if scrobble percent storage is somehow corrupted.
+ */
+export const DEFAULT_SCROBBLE_PERCENT = 50;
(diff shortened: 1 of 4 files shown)


src/util/util.ts
```diff
@@ -26,6 +26,12 @@ export const HIDDEN_PLACEHOLDER = '[hidden]';
  */
 export const DEFAULT_SCROBBLE_TIME = 30;
 
+/**
+ * Percentage of track to playback before the track is scrobbled.
+ * This value is used only if scrobble percent storage is somehow corrupted.
+ */
+export const DEFAULT_SCROBBLE_PERCENT = 50;
```
inverse, line 33: Can we use this const in `GlobalOptions` where we set to `50` to start with?

inverse (approved):
nice catch!

src/util/util.ts
yayuyokitano: Fixed, nice one

inverse (approved):

PR #4353, opened 2023-12-10
Remove uploading build artifacts

inverse:

Not sure when we ever reference these artifacts and seems to be a big chunk of CI time doing it.

**Additional context**


yayuyokitano (approved):
Agreed, I think this is just one of those "always been that way" things that was never questioned. Don't see a good reason to keep it. Nice one

Issue #4342, opened 2023-12-08
YouTube Music connector unable to find unique IDs


LostLuma:
**Describe the bug**

The YouTube Music connector can no longer find the unique ID (video ID) of each track.  
This prevents using previous saved edits, and also saves them in a less useful format.

**How to reproduce**
1. Save an edit for a track on YouTube music
2. Export the saved track storage
3. A hash of the pre-edit track data is used as the key instead of the video ID

**Expected behavior**

Saved edits should use the video ID, same as they've done previously and do on regular YouTube.

**Screenshots**

```json5
{
  // This should be 1tkdmY7sdUc instead
  "6cb3267d1f7d02d71b367996999f5350": {
    "artist": "IZ*ONE",
    "track": "Panorama",
    "album": "One-reeler / Act IV",
    "albumArtist": null
  }
}
```

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   Browser: Firefox 122.0a1
-   Extension version: 3.2.0

SeijiSuenaga:
This problem has been intermittent in the past few days. Sometimes the YouTube Music DOM contains the `.yt-uix-sessionlink` element (which the connector depends on), but other times it doesn't. 🤔

inverse:
Will beh handled in the next release - thanks @LostLuma!

PR #4359, opened 2023-12-13
Get custom ID from URL if possible on YouTube Music

LostLuma:

Changes the YouTube Music connector to get the custom ID from the current page URL if possible, and only then falls back to searching the page for it. This helps with #4342, but is not an entire fix as it doesn't work when the user is not on the video view.

**Additional context**

I've also experimented with a way of fixing this while the user is in another view, however since sometimes the current video ID is no longer found in the DOM at all I resorted to opening the current track's context menu and extracting it from the "Start radio" button that is then rendered. Reason I have not included it in this PR is that I don't think any other connector automatically interacts with the page beyond reading it, and I'd like to get some feedback on whether the maintainers oppose it doing so first.

### DIFF
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -104,12 +104,14 @@ Connector.isPlaying = () => {
 };
 
 Connector.getUniqueID = () => {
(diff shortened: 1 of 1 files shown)


inverse (approved):
Changes seem reasonable to me - thanks for the fix :+1:

PR #4394, opened 2024-01-02
Update SiriusXM to support player redesign

### LINKED ISSUE #4362: SiriusXM Website Update, No Longer Scrobbles
**Describe the bug**
SiriusXM's website was updated and it appears the connectors are broken.

**How to reproduce**
1. Go to https://www.siriusxm.com/player/my-library
2. Click on a station
3. Play station
4. "This website is not supported"

**Expected behavior**
Extension should be scrobbling the current song playing.

**Screenshots**
image

-   OS: Windows 11
-   Browser: Firefox
-   Extension version: 3.3.0

dasevilla:
Update the domain and selectors to work with the SiriusXM website redesign. Fixes #4362.

### DIFF
--- a/src/connectors/siriusxm-player.ts
+++ b/src/connectors/siriusxm-player.ts
@@ -15,20 +15,18 @@ const filter = MetadataFilter.createFilter({
 	track: [removeYear, removeCover, removeExclusive],
 });
 
-Connector.playerSelector = '.sxm-player-controls';
-Connector.artistSelector = '.sxm-player-controls .artist-name';
-Connector.trackSelector = '.sxm-player-controls .track-name';
+Connector.playerSelector = '._playbackControls_1cag9_17';
 
-Connector.isPlaying = () => {
-	return (
-		Util.getAttrFromSelectors(
-			'.sxm-player-controls .play-pause-btn',
-			'title',
-		) === 'Pause'
-	);
-};
+Connector.artistTrackSelector = '._playbackControls_1cag9_17 ._title_1cag9_286';
 
-Connector.trackArtSelector = '.album-image-cell img.album-image';
+Connector.trackArtSelector =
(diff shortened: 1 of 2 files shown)


src/connectors/siriusxm-player.ts
```diff
@@ -15,20 +15,18 @@ const filter = MetadataFilter.createFilter({
 	track: [removeYear, removeCover, removeExclusive],
 });
 
-Connector.playerSelector = '.sxm-player-controls';
-Connector.artistSelector = '.sxm-player-controls .artist-name';
-Connector.trackSelector = '.sxm-player-controls .track-name';
+Connector.playerSelector = '._playbackControls_1cag9_17';
```
inverse, line 18: While this will work - it feels like the connector is going to be super brittle to any new publish of the site as I suspect these IDs will mutate. Is there any other DOM selectors?

dasevilla: I agree. The new site smells of auto generated classes from a framework. There aren't good classes or IDs to use.

Another approach would be to try selectors that match the base of the class without the IDs. I haven't tried that before and I'm open to suggestions.

inverse: I'm unable to use the site to see what the DOM is like due to georestrictions so I suspect this is better than nothing. happy to accept it in its current form since the site is currently broken.

brodie-g: Thanks for updating to support the re-design!

To chime in here, I dealt with a similar selector creating some custom css for the new re-design - https://userstyles.world/style/13696/siriusxm-com-dec-2023

Are you able to use wildcard attribute selector, like
```
*[class^='_playbackControls']
```

Would web-scrobbler support the attribute selector syntax?

dasevilla: I appreciate the suggestion and have implemented it in #4407

PR #4407, opened 2024-01-06
Update SiriusXM player to use attribute selectors

dasevilla:
Improve upon #4394 by using attribute selectors as recommended by @brodie-g


yayuyokitano (approved):

Issue #3717, opened 2023-05-10
bilibili.com connector request


raemoend:
Bilibili is a video streaming platform similar to Youtube and Niconico: https://www.bilibili.com/video/BV11o4y1L79d)

PR #4409, opened 2024-01-09
Add connector for BiliBili

turbodog03:
I added a connector for BiliBili.com, a video platform similar to YouTube and NicoNico.
This connector fulfills the request in #3717 

Since it's a video site, the page doesn't display song information directly. I use the video title and tags to extract the artist and track names as accurately as possible. It works well when the uploader tags the video correctly (which most of them do), and exclude non-music videos as well.

If there's anything I can do to make it better please let me know :)

### DIFF
--- a/src/connectors/bilibili.ts
+++ b/src/connectors/bilibili.ts
@@ -1,0 +1,504 @@
+export {};
+
+/*
+	Since BiliBili is a video site, accurately scrobbling song information from the title can be challenging. 
+	However, people often include the artist's name and track title in the video tags. 
+
+	Therefore, this connector uses tags to identify artist and track information and determines whether we should scrobble this video.
+	The video title is then used to determine which tag is most likely to contain the artist or track name.
+*/
+
+/**
+ * Quick link to debug and test the connector: https://www.bilibili.com/video/BV1LW4y1g7kN
+ *
+ * the scrobble info will shown beside the video info, under the video title.
+ *
+ * special behaviors:
+ * if the video has bgm tag(a tag with music symbol in front of it), it'll be recognized as track name
+ * if the video has no track or artist related tag, this connector will try grepping it from video title
+ */
+
+/**
+ * selectors used by this connector
+ */
+const videoTitleSelector = '.video-title';
+const videoSelector = 'video';
+const tagPanelSelector = '.tag-panel';
+const tagLinkSelector = '.tag-link';
+const uploaderSelector = '.up-name';
+const bgmTagSelector = '.bgm-tag .tag-link .tag-txt';
+
+/**
+ * selectors used by Connector core
+ */
+const PlayerSelector = '.bpx-player-container';
+const currentTimeSelector = '.bpx-player-ctrl-time-current';
+const durationSelector = 'bpx-player-ctrl-time-duration';
+/**
+ * patterns that used to match possible song info
+ */
+const trackPattern = /[《『「](.*?)[》』」]/g;
+const artistPattern = /[【](.*?)[】]/g;
+const decorationPattern = /[(（](.*?)[)）]/g;
+// do not use lookbehind pattern like this, firefox does not support it
+// const coveredByPattern = /(?<=covered by).*/i;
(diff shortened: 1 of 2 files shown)


src/connectors/bilibili.ts
```diff
@@ -0,0 +1,513 @@
...
+			!filterKeyWords.some((keyword) => {
+				return item.toLowerCase().includes(keyword.toLowerCase());
+			}),
+	);
+}
+/**
+ * main function used to get song info.
+ * @returns \{artist, track\}
+ */
+function grepSongInfo() {
```
yayuyokitano, line 283: Without having looked *too* thoroughly at this yet, could it be reasonable to reuse the Util.processYtVideoTitle() function here, add whatever separators and filters that won't break YT behavior that are missing to that, and then do some additional processing on top of that after?

I think this should simplify connector logic a fair bit, and will have a bonus of improving chinese track detection for all sites that use youtube's processing function.

Or is there some specific reason this is not realistic?

yayuyokitano (commented):
Seems to be a very well-crafted connector! Thank you for the contribution!
I have one inquiry about the title recognition that I have detailed in my review though.

src/connectors/bilibili.ts
turbodog03: Yes. It's a great function to get track and artist name from title, and I agree that it's awesome if I can adjust and reuse it, the code will be more neat and other future connector will benifit from it. Howerver, it's not suitable for this connector's logic.

I tried this func, and found it didn't work well on BiliBili as the preferences of uploaders vary in how they title their videos. thus I decide to prioritize the use of tags to enhance  accuracy. 

In this tag-centric method, the video title serves as assistance. The connector use patterns to get as much title fragment that may contain track or artist infomation as possible, push them into `possibleArtist/Track` array in pattern priority order, traverse it and find if there's any tag included in it. 

For example, consider the test link in the code (which was well-titled and tagged, though many other videos have more complicated titles) :

> videoTitle: 【官方】RADWIMPS - me me she [Music Video]
> matched possibleTrack: me me she [Music Video] 
> matched possibleArtist: 官方,RADWIMPS拉德温普斯,【官方】RADWIMPS 
> FilteredTags: me me she,RADWIMPS,摇滚
> 1. traverse possibleTrack array, found there's an tag contained in it, so take it as track name
> 2. traverse possibleArtist array, when get to the second element('RADWIMPS拉德温普斯'), found there's an tag contained in it, so the tag is going to be the artist name.

So it's hard to modify the `Util.processYtVideoTitle()` function and reused it without break it's origin behavior, since the connector needs an array containing many title fragments , not an exact string. 

I use `forEach()` method for separator array to separate the text instand of using the `Util.splitString() with the same reason. Just to ensures that, if there are multiple separators in the title, it can obtain each split result for increased accuracy.

Sorry for being this long, thanks for taking your time reading it!
Through may not being able to reuse the function, I went through the code again and tested it on more videos, then did some optimize on the connector.  If there're other inquires please let me know!

yayuyokitano: Understood. I will have a closer look at this, and then I think we can probably merge it tomorrow 😊

src/connectors/bilibili.ts
```diff
@@ -0,0 +1,504 @@
...
+const tagLinkSelector = '.tag-link';
+const uploaderSelector = '.up-name';
+const bgmTagSelector = '.bgm-tag .tag-link .tag-txt';
+
+/**
+ * selectors used by Connector core
+ */
+const PlayerSelector = '.bpx-player-container';
+const currentTimeSelector = '.bpx-player-ctrl-time-current';
+const durationSelector = 'bpx-player-ctrl-time-duration';
```
yayuyokitano, line 36:
```suggestion
const durationSelector = '.bpx-player-ctrl-time-duration';
```

turbodog03: Fixed. Sorry for the mistake 😥 thank you for pointing it out!

yayuyokitano:

PR #4416, opened 2024-01-11
Add notification when first scrobbling site with built-in scrobbler

### LINKED ISSUE #1375: Notify of native scrobbling
Some websites have own implementation of Last.fm scrobbling. The extension should notify of such feature. Optionally, it should provide a simple way to disable connectors to prevent double scrobbling.

I suggest to show a notification with informational message when an user visits the website first time. Users can disable connectors for the website by clicking on the notification.

Extracted from #630.

yayuyokitano:
When first playing on a site with built-in scrobbler, a notification is shown to the user to make them aware.
Notification shows when a song actually starts playing, not on visit.
Notification will show only once per website and then be disabled through new local storage property.

Currently only enabled for spotify, but I'm sure there are more sites.
A site having a built-in scrobbler is denoted using new hasNativeScrobbler property in connectors.ts

To accomodate this change, the song types have been changed to have a new property connector: ConnectorMeta.
song.connectorLabel and song.connectorId have been removed, since they are made obsolete by song.connector.

image

Closes #1375

### DIFF
--- a/src/core/storage/native-scrobbler-notification.ts
+++ b/src/core/storage/native-scrobbler-notification.ts
@@ -28,11 +28,8 @@ export default class NativeScrobblerNotification {
 	 * @returns true if notification has not been given for connector; false otherwise
 	 */
 	public async shouldNotifyAboutNativeScrobbler(
-		id: string | undefined | null,
+		id: string,
 	): Promise<boolean> {
-		if (!id) {
-			return false;
-		}
 		const data = await this.storage.get();
 		if (!data) {
 			return true;
--- a/src/core/storage/native-scrobbler-notification.ts
+++ b/src/core/storage/native-scrobbler-notification.ts
@@ -1,0 +1,42 @@
+import * as BrowserStorage from '@/core/storage/browser-storage';
+
+export default class NativeScrobblerNotification {
+	private storage = BrowserStorage.getStorage(
+		BrowserStorage.NATIVE_SCROBBLER_NOTIFICATION,
+	);
+
+	/**
+	 * Add connector ID to list of connectors that has been notified about.
+	 *
+	 * @param id - ID of connector to add
+	 */
+	public async saveHasNotified(id: string): Promise<void> {
+		if (!id) {
+			return;
+		}
+		let data = await this.storage.getLocking();
+		if (!data) {
+			data = {};
+		}
+		data[id] = true;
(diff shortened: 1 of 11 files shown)


src/core/storage/native-scrobbler-notification.ts
```diff
@@ -0,0 +1,42 @@
...
+		await this.storage.setLocking(data);
+	}
+
+	/**
+	 * @param id - ID of connector to check
+	 *
+	 * @returns true if notification has not been given for connector; false otherwise
+	 */
+	public async shouldNotifyAboutNativeScrobbler(
+		id: string | undefined | null,
```
inverse, line 31: When can this be either `undefined` or `null`? isn't `id` manditory for connectors? Should we create a type to enforce this?

src/util/notifications.ts
```diff
@@ -157,6 +158,46 @@ async function showNotification(
...
+
+	try {
+		await showNotification(options, () => {
+			browser.tabs.create({
+				url: browser.runtime.getURL(
+					'src/ui/options/index.html?p=connectors',
+				),
+			});
+		});
+		await nativeScrobblerNotification.saveHasNotified(connector.id);
```
inverse, line 194: :bulb: what about making the notification such that it has to be actively dismissed to not come up again.

I can imagine folks would miss this? I guess for now this is a good first step and we can always iterate though.

inverse (approved):

src/core/storage/native-scrobbler-notification.ts
yayuyokitano: Removed this, probably triggered this some other way before and moved the function call somewhere id isnt null later.

src/util/notifications.ts
yayuyokitano: Yeah it is a little unlucky, but unfortunately I'm not aware of any nice UX way to do this with notifications (no custom buttons on notifs).

If so we'd have to in that case have that be on the user manually closing the notif before it disappears on its own using the normal "X" button, but that seems very unintuitive to me, and I think a fair number of people will never click to close it and be annoyed that it keeps popping up all the time.

Alternatively we'd have to go into dom injection which seems annoying to maintain.

inverse: Lets go with this and we can always iterate if folks still have problems :)

PR #4418, opened 2024-01-11
Add high contrast themes

### LINKED ISSUE #784: High-Contrast-Mode/Theming ...

yayuyokitano:
Draft for now
Adds high contrast themes
Rough around the edges but I think this is sufficient? Not sure, not very well read on this.

- [x] Add high contrast dark theme
- [x] Add high contrast light theme
- [x] Set up system theme to use these to the best extent possible.

Closes #784 

image

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -155,6 +155,10 @@
         "message": "Dark",
         "description": "Option label"
     },
+    "optionThemeHighContrastDark": {
+        "message": "High Contrast Dark",
+        "description": "Option label"
+    },
     "optionThemeSystem": {
         "message": "System",
         "description": "Option label"
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -159,6 +159,10 @@
         "message": "High Contrast Dark",
         "description": "Option label"
     },
+    "optionThemeHighContrastLight": {
+        "message": "High Contrast Light",
+        "description": "Option label"
+    },
     "optionThemeSystem": {
         "message": "System",
         "description": "Option label"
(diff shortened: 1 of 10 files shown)


yayuyokitano:
Added an override that makes it so windows high contrast rules arent applied at all. This is a bit of a blunt instrument, but it was doing a lot more damage than helping (checkboxes just werent working at all). System theme will default to using high contrast themes, so shouldnt be huge issue hopefully.

inverse (approved):
This is awesome!

yayuyokitano:
Let's await merging this until we've gotten 3.5.0 shipped since its a bit of a rushed release

PR #4419, opened 2024-01-11
Add blocklist channel label

yayuyokitano:
Extremely core functionality, but is a relatively small change that should be easy to work through.
Lightly tested as of now.
Adds three new properties to connector.

Connector.getChannelLabel is a function that returns a channel label (or **falsy** (not just nullish) if there is none)
Connector.channelLabelSelector is the selector of element with channel label used in default Connector.getChannelLabel implementation.
Connector.getChannelInfo returns an object with both id and label, by default uses Connector.getChannelID and Connector.getChannelLabel, falling back on Connector.getChannelID if Connector.getChannelLabel is falsy, and returning null if Connector.getChannelID is **falsy**.

This also means that you only need to specify Connector.getChannelID for the functionality to work, with the label becoming the ID.

Also renames Connector.getChannelId to Connector.getChannelID because this was needed for consistency with Connector.getUniqueID.

Underneath the hood, storage changes from blocklist being object {[channelID]: true}
to being object {[channelID]: channelLabel} (this allows us to show channel label in connector setings when deleting)

Also implemented the new system for both youtube and bilibili which are currently only connectors using this.

Before/after
image
image

### DIFF
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -150,7 +150,51 @@ export default class BaseConnector {
 	 * Has to be specified if usesBlocklist is set to true in connectors.ts.
 	 * If connectors.ts does not have usesBlocklist set to true, this should be null.
 	 */
-	public getChannelId: (() => string | null | undefined) | null = null;
+	public getChannelID: (() => string | null | undefined) | null = null;
+
+	/**
+	 * Selector of an element containing channel label.
+	 *
+	 * Only applies when default implementation of
+	 * {@link getChannelLabel} is used.
+	 */
+	public channelLabelSelector: string | string[] | null = null;
+
+	/**
+	 * Function that gets a label for the channel of ID fetched by getChannelID.
+	 *
+	 * This is the name that will be displayed to the user, and has no bearing on internal logic.
+	 * If not specified, {@link getChannelID} will be used.
+	 */
+	public getChannelLabel: (() => string | null | undefined) | null = () =>
+		Util.getTextFromSelectors(this.channelLabelSelector);
+
+	/**
+	 * Function that gets ID and label for a channel.
+	 *
+	 * A connector can specify this in lieu of {@link getChannelID}, as this is basically combo of
+	 * both {@link getChannelID} and {@link getChannelLabel}
+	 *
+	 * You may return null in this function, but if you return a result for one property,
+	 * you must return a result for both properties, even if the label is just
+	 * a duplicate of the id.
+	 */
+	public getChannelInfo: () =>
+		| {
+				id: string;
+				label: string;
+		  }
+		| null
+		| undefined = () => {
+		const id = this.getChannelID?.();
+		if (!id) {
+			return null;
+		}
+		return {
+			id,
+			label: this.getChannelLabel?.() || id,
+		};
+	};
 
 	/**
 	 * Selector of element contains a track art of now playing song.
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -150,7 +150,7 @@ export default class BaseConnector {
 	 * Has to be specified if usesBlocklist is set to true in connectors.ts.
 	 * If connectors.ts does not have usesBlocklist set to true, this should be null.
 	 */
-	public getChannelID: (() => string | null | undefined) | null = null;
+	public getChannelId: (() => string | null | undefined) | null = null;
 
 	/**
 	 * Selector of an element containing channel label.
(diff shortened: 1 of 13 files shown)


yayuyokitano:
@inverse this PR is highest priority. This should be considered blocking of a new release.

yayuyokitano:
Feel pretty confident this doesnt have any regressions at this point

src/core/content/connector.ts
```diff
@@ -150,7 +150,51 @@ export default class BaseConnector {
 	 * Has to be specified if usesBlocklist is set to true in connectors.ts.
 	 * If connectors.ts does not have usesBlocklist set to true, this should be null.
 	 */
-	public getChannelId: (() => string | null | undefined) | null = null;
+	public getChannelID: (() => string | null | undefined) | null = null;
```
inverse, line 153: I know its a minor detail but I still think we should keep the casing of `Id` as it aligns with other things in the repo and the stdlib. e.g. `trackId`  `videoId` variable naming, `getElementById` etc.

src/core/content/connector.ts
```diff
@@ -150,7 +150,51 @@ export default class BaseConnector {
...
+		Util.getTextFromSelectors(this.channelLabelSelector);
+
+	/**
+	 * Function that gets ID and label for a channel.
+	 *
+	 * A connector can specify this in lieu of {@link getChannelID}, as this is basically combo of
+	 * both {@link getChannelID} and {@link getChannelLabel}
+	 *
+	 * You may return null in this function, but if you return a result for one property,
+	 * you must return a result for both properties, even if the label is just
```
inverse, line 179: Is this correct? Don't we use `id` already if the label is `null`?

src/core/storage/blocklist.ts
```diff
@@ -37,8 +37,13 @@ export default class Blocklist {
 	 *
 	 * @param id - ID of channel to add
 	 */
-	public async addToBlocklist(id: string): Promise<void> {
-		if (!id) {
+	public async addToBlocklist(
+		channelInfo: {
```
inverse, line 41: Should this type be extracted?

src/core/object/controller/controller.ts
```diff
@@ -317,7 +317,7 @@ export default class Controller {
 				type: 'removeFromBlocklist',
 				fn: async () => {
 					await this.blocklist.removeFromBlocklist(
-						this.connector.getChannelId?.() ?? '',
+						this.connector.getChannelInfo?.()?.id ?? '',
```
inverse, line 320:
```suggestion
						this.connector.getChannelInfo?.()?.id ?? null,
```

Wouldn't this be more correct? Why would we pass an empty string instead of null when the `id` does not exist?

src/util/communication.ts
```diff
@@ -168,7 +168,13 @@ interface BackgroundCommunications {
 		payload: undefined;
 		response: {
 			connector: ConnectorMeta;
-			channelId: string | null | undefined;
+			channelInfo:
```
inverse, line 171: Another example for type extraction :)

src/core/object/controller/controller.ts
```diff
@@ -97,8 +97,8 @@ export default class Controller {
 		}
 		return (
 			this.currentSong?.parsed.isScrobblingAllowed &&
-			(await this.blocklist.shouldScrobbleChannel(
-				this.connector.getChannelId?.(),
+			!(await this.blocklist.getChannelLabel(
```
inverse, line 100: I feel the sematics of this is a lot harder to read now. Like the channel label dictacts if scrobbling is allowed? Can we add a util function that is called `shouldScrobbleChannel` back again?

inverse (commented):
Some minor comments :+1:

src/core/content/connector.ts
yayuyokitano: The issue is its inconsistent with `Connector.getUniqueID`, which AFAIK is the only public facing function we have using id.
Maybe we should rename that to `Connector.getUniqueId` in a future PR?

src/core/content/connector.ts
yayuyokitano: This behavior is ensured by the default implementation of `getChannelInfo` yes.
However, if the connector developer decides to override `getChannelInfo`, they must do their own implementation.
I think it is nice to force the developer to consciously think about what they are doing with that.
Most connectors will be using `getChannelId` and `channelLabelSelector` probably, and default implementation of `getChannelInfo`.

src/core/object/controller/controller.ts
yayuyokitano: This was done because the removeFromBlocklist signature demands a string, but we can probably change that to accept null. It will not impact implementation at all.

src/core/object/controller/controller.ts
yayuyokitano: Done

src/core/storage/blocklist.ts
yayuyokitano: Done

src/core/object/controller/controller.ts
yayuyokitano: Done

inverse (approved):
Makes sense :+1: 

Looks a lot clearer to me now too :)

yayuyokitano:
Lets push new version tomorrow after I wake up.

PR #4430, opened 2024-01-15
Improve keyboard and screen viewer accessibility

yayuyokitano:
Draft as it is still unfinished, but starting to get close.
Does many things to improve ability of user to navigate without mouse and without seeing, which were both things we were severely lacking before.

Some missing things:
- [x] The connector list in options is not behaving correctly
- [x] Edits are being applied when clicking enter on other buttons in edit menu .stopImmediatePropagation is not helping, but shouldnt be terribly difficult to fix regardless
- [x] Regex/bulk edit menu is a mess
- [x] Check and fix iOS if needed


yayuyokitano:
Edit and regex edit menus look a bit different. I have decided that I do not care.

Before (macos/edge):
image
image

After (macos/edge):
image
image

Before (iOS):
image
image

After (iOS):
image
image

inverse (approved):
LGTM

PR #4434, opened 2024-01-18
Add detection through DOM elements

yayuyokitano:
Adds ability for web scrobbler to detect sites through DOM elements.
Seems to be working, but is behaving very weird.
Also funkwhale connector seems to be extremely buggy in general, but I don't think it's connected to this PR?
Still need to do more checking on this and try to figure out why funkwhale isn't working properly.

### DIFF
--- a/src/connectors/funkwhale.ts
+++ b/src/connectors/funkwhale.ts
@@ -39,9 +39,9 @@ function setupNewPlayer() {
 
 	Connector.trackSelector = '.content .track';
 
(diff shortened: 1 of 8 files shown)


yayuyokitano:
For the record, while there is an issue where funkwhale does not load the .controls-row element before web scrobbler tries to set up an observer there, delaying setting up the observer until the element is there does not even fix the issue.

jaccarmac:
This is enabled in certain connectors now, right?

PR #4437, opened 2024-01-19
Add ability to love songs through listening services

### LINKED ISSUE #1424: Auto like on lastfm if a person likes a track on scrobbleable service
### Expected behavior
If you like a track, for example, on youtube - it should also like that track on lastfm

### Current behavior
Nothing happens.

yayuyokitano:
Closes #1424 
Really dont know about these labels in options, I think they can be improved a lot.
Adds properties to connector:
`Connector.isLoved`: function that returns true/false/nullish for whether a song is loved on website or not.
`Connector.loveButtonSelector`: selector or selectors for love/like button. When this is visible the song is interpreted as not loved.
`Connector.unloveButtonSelector`: selector or selectors for unlove/unlike button. When this is visible the song is interpreted as loved.

Only one of these three should be specified in any given connector similar to play/pause button selectors and isplaying function.

Property has been implemented to youtube and spotify connectors so far, but I think we could do with a few more. You can see how it works in practice though in those connector implementations.

### DIFF
--- a/src/core/content/connector.ts
+++ b/src/core/content/connector.ts
@@ -646,16 +646,22 @@ export default class BaseConnector {
 	/**
 	 * Callback set by the controller to listen on state changes of this connector.
 	 */
-	private _isLovedCallback: ((isLoved: boolean | null) => void) | null = null;
+	private _isLovedCallback:
+		| ((isLoved: boolean | null) => Promise<void>)
+		| null = null;
 
 	/**
 	 * Callback set by the controller to listen on state changes of this connector.
 	 */
-	public get isLovedCallback(): ((isLoved: boolean | null) => void) | null {
+	public get isLovedCallback():
+		| ((isLoved: boolean | null) => Promise<void>)
+		| null {
 		return this._isLovedCallback;
 	}
 
-	public set isLovedCallback(callback: (isLoved: boolean | null) => void) {
+	public set isLovedCallback(
+		callback: (isLoved: boolean | null) => Promise<void>,
+	) {
 		callback(this.isLoved() ?? null);
 		this._isLovedCallback = callback;
 	}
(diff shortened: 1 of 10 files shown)


src/core/content/connector.ts
```diff
@@ -614,6 +643,29 @@ export default class BaseConnector {
...
+	public get isLovedCallback():
+		| ((isLoved: boolean | null) => Promise<void>)
+		| null {
+		return this._isLovedCallback;
+	}
+
+	public set isLovedCallback(
+		callback: (isLoved: boolean | null) => Promise<void>,
+	) {
+		callback(this.isLoved() ?? null);
```
inverse, line 665: is the `?? null` to ensure either `bool` or `null`?

src/core/object/controller/controller.ts
```diff
@@ -567,6 +567,60 @@ export default class Controller {
...
+
+		/**
+		 * State did not change, don't do anything.
+		 */
+		if (this.currentSong.flags.isLovedInService === isLoved) {
+			return;
+		}
+
+		/**
+		 * song already had fetched a definitive loved state,
```
inverse, line 608:
```suggestion
		 * Song already had fetched a definitive loved state,
```

inverse (approved):
Change seems nice - awesome work :+1: the callback stuff in src/core/content/connector.ts is a bit hard to follow  but it works :)

src/core/content/connector.ts
yayuyokitano: Correct, I allow connectors to return undefined on no result to be more lenient on connector creators, in line with other connector functions.

yayuyokitano:
> Change seems nice - awesome work 👍 the callback stuff in src/core/content/connector.ts is a bit hard to follow but it works :)

Yeah, I think maybe we want to rethink this structure now that controller and connector are tightly coupled anyway, but for now I have just maintained the same structure as our existing callback to be consistent.

PR #4440, opened 2024-01-20
Ignore yarn.lock

TheStranjer:
One of the edits @yayuyokitano made in PR 3908 was to remove `yarn.lock`, so I figured it might be ideal to just add it to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@ coverage/
 
 # Cache
 .eslintcache
+
+# Yarn
+yarn.lock
+


inverse:
We're not using yarn on this project though? won't that potentially cause dependency issues since we're using the `package-lock.json` to locke dependencies via npm?

yayuyokitano:
> We're not using yarn on this project though? won't that potentially cause dependency issues since we're using the `package-lock.json` to locke dependencies via npm?

Adding to gitignore prevents the file from being committed, even if some contributor is using yarn on their end (which does work, even though it is not great for reproducible builds since we use package-lock)

yayuyokitano (approved):
Good call!

PR #4452, opened 2024-01-24
Fix isScrobblingAllowed logic for BiliBili

turbodog03:
## changes I made
I've corrected the logic for determining whether to scrobble a video on BiliBili. 
The `tagFilterKeyWords` should always being used when `useScrobbleTagFilter` is true. However, in the previous code, it wasn't when there was a '音乐' (which means music) tag.

## example
When a video is tagged ['音乐','音乐教学'] (which means ['music', 'music tutorial']), it would be scrobbled, but actually it shouldn't be. This pull request fixes that issue."


inverse (approved):
Thanks for the contribution!

PR #4455, opened 2024-01-26
Add regex flags and allow replace with empty

yayuyokitano:
Draft for now: testing of UI is still a bit light (though unit tests have been added for application of the edits themselves and I think I'm reasonably satisfied with those). UI design I'm also not completely sure of, would appreciate feedback!

Adds flags to enable/disable regex, enable/disable case sensitivity, enable/disable having to match entire tag to apply. Also adds a button to allow user to actually replace with nothing, instead of just empty text field which makes it just apply the search without doing any replace.

Is built on top of the old system and is completely backwards-compatible with old regex/bulk edits, without any migration needed. For this reason, internal flag names are a bit odd.

**UI:**

Desktop:
image

iOS:
image

iPadOS:
image


yayuyokitano:
Changed bulk edit icon, as old was not obvious.

image

yayuyokitano:
This PR also drastically speeds up tests through mocking fetch API.

yayuyokitano:
For some reason new test seems to be flaky in CI.
Cannot reproduce it in local at all

yayuyokitano:
now i cant repro in ci either

yayuyokitano:
Keep retrying this but it just won’t fail again now that I added logic to force traceable throwing🙃🙃
Gonna keep trying to rerun it for a bit but…

yayuyokitano:
Just had it happen in a different branch, its unrelated to changes made here, I guess something happened to something external

yayuyokitano:
Well got good logs for the test flakiness, it looks vite related. Might talk to them a bit about it and see what they think. Merging this anyhow, test flakiness is unrelated to this PR

Issue #4374, opened 2023-12-18
 Notify if user loved/unloved song via hotkey


CennoxX:
In terms of usability, it would be good to get feedback whether using the hotkey has had any effect. The only way to find out is to open the popup, which makes the hotkey kinda useless. 

In version 2.x this was made by displaying a  temporary browser action icon on hotkey press - with version 3 this doesn't work anymore. 
Here is the implementation of version 2: https://github.com/web-scrobbler/web-scrobbler/commit/a945a4f2715d59aba16df1736399156dd26e399d

ahnesther:
Hello @yayuyokitano: I was wondering if I could take over this issue, since I have been working on this and the major kinks have been worked out. There are some things I have to look into, like localization and testing on Safari. Please let me know!

yayuyokitano:
Go for it! Feel free to open a draft PR if you want help with localization related things and Safari testing.
Thank you for letting me know!

ahnesther:
Awesome, thanks!!

ahnesther:
@yayuyokitano https://github.com/web-scrobbler/web-scrobbler/pull/4472 Draft PR ready to be looked at :)

PR #4472, opened 2024-01-29
 Notify if user loved/unloved song via hotkey

ahnesther:
This PR addresses this issue: https://github.com/web-scrobbler/web-scrobbler/issues/4374

-Added functionality to show a popup notification when a song is loved or unloved through user defined hotkeys. 
-Noticing that hotkeys went through the `commandHandler` function in main.ts, I added an `isCommand` parameter to `setLoveStatus`, backgroundListeners, contentListeners, contentCommunications, backgroundCommunications.
-In `nowplaying.ts`, `isCommand` is set as false as that's the extension popup UI.
- Change icon to loved/unloved when song is loved/unloved toggled through the scrobble service (https://github.com/web-scrobbler/web-scrobbler/pull/4437) or through hotkeys, then is reset to default state.

**Additional context**
While I can run the extension on Windows Chrome and Firefox, I can't seem to do so on Safari. The extension does build, but the instructions to add the extension to Safari doesn't work: after I right click and open or double click, it appears for a microsecond then disappears with no extension added. My mac is an older model so it's limited to Catalina 10.15 and to XCode12.4, so that could be a reason why it's not working. 
Edit: (February 6) still unable to test on Mac/Safari.

### DIFF
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -77,6 +77,7 @@ browser.commands?.onCommand.addListener(
  */
 async function commandHandler(command: string) {
 	const tab = await getCurrentTab();
+	const alreadyLoved = tab.song?.metadata.userloved;
 
 	switch (command) {
 		case 'toggle-connector':
@@ -87,10 +88,18 @@ async function commandHandler(command: string) {
 			}
 			break;
 		case 'love-song':
-			setLoveStatus(tab.tabId, true, true);
+			// only set love status if song is not yet loved, ignore if song is already loved
+			// only send notification when song is unloved but will be loved
+			if (!alreadyLoved) {
+				setLoveStatus(tab.tabId, true, true);
+			}
 			break;
 		case 'unlove-song':
-			setLoveStatus(tab.tabId, false, true);
+			// only set unlove status if song is loved, ignore if song is unloved
+			// only send notification when song is loved but will be unloved
+			if (alreadyLoved) {
+				setLoveStatus(tab.tabId, false, true);
+			}
 			break;
 	}
 }
@@ -100,7 +109,7 @@ async function commandHandler(command: string) {
  *
  * @param tabId	- Tab ID of the tab to update
  * @param isLoved - Whether the song is loved
- * @param shouldShowNotification - Whether the song is (un)loved by hotkey command
+ * @param shouldShowNotification - Whether the song should show notification when (un)loved
  *
  */
 function setLoveStatus(
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -87,10 +87,10 @@ async function commandHandler(command: string) {
 			}
 			break;
 		case 'love-song':
-			setLoveStatus(tab.tabId, true);
+			setLoveStatus(tab.tabId, true, true);
 			break;
 		case 'unlove-song':
-			setLoveStatus(tab.tabId, false);
+			setLoveStatus(tab.tabId, false, true);
 			break;
 	}
 }
@@ -100,13 +100,15 @@ async function commandHandler(command: string) {
  *
  * @param tabId	- Tab ID of the tab to update
  * @param isLoved - Whether the song is loved
+ * @param isCommand - Whether the song is (un)loved by hotkey command
  *
  */
-function setLoveStatus(tabId: number, isLoved: boolean) {
+function setLoveStatus(tabId: number, isLoved: boolean, isCommand: boolean) {
 	sendBackgroundMessage(tabId ?? -1, {
 		type: 'toggleLove',
 		payload: {
 			isLoved,
+			isCommand: true,
 		},
 	});
 }
@@ -413,7 +415,9 @@ setupBackgroundListeners(
 		type: 'toggleLove',
 		fn: (payload, sender) => {
 			const song = new ClonedSong(payload.song, sender.tab?.id ?? -1);
-			showLovedNotification(song, payload.isLoved);
+			if (payload.isCommand) {
+				showLovedNotification(song, payload.isLoved);
+			}
 			return toggleLove(song, payload.isLoved);
 		},
 	}),
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -32,6 +32,7 @@ import { CloneableSong } from '@/core/object/song';
 import {
 	clearNowPlaying,
 	showAuthNotification,
+	showLovedNotification,
 	showNowPlaying,
 	showSongNotRecognized,
 } from '@/util/notifications';
@@ -411,10 +412,9 @@ setupBackgroundListeners(
 	backgroundListener({
 		type: 'toggleLove',
 		fn: (payload, sender) => {
-			return toggleLove(
-				new ClonedSong(payload.song, sender.tab?.id ?? -1),
-				payload.isLoved,
-			);
+			const song = new ClonedSong(payload.song, sender.tab?.id ?? -1);
+			showLovedNotification(song, payload.isLoved);
+			return toggleLove(song, payload.isLoved);
 		},
 	}),
 
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -100,15 +100,19 @@ async function commandHandler(command: string) {
  *
  * @param tabId	- Tab ID of the tab to update
  * @param isLoved - Whether the song is loved
- * @param isCommand - Whether the song is (un)loved by hotkey command
+ * @param shouldShowNotification - Whether the song is (un)loved by hotkey command
  *
  */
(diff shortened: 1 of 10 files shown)


yayuyokitano:
This implementation will not work on safari, as safari has no support for notifications API.
I think we need to do a similar strategy to V2, with the icon temporarily changing (maybe we can do both actually with notification being disableable, I quite like the notification.)
Implementation otherwise looks sound.
One request: Could you rename the isCommand property to something like "shouldShowNotification"? This would play a lot nicer with #4437 which is not a command, but which I think should also be using this.

src/core/background/main.ts
```diff
@@ -99,13 +100,15 @@ async function commandHandler(command: string) {
...
+ * @param isCommand - Whether the song is (un)loved by hotkey command
  *
  */
-function setLoveStatus(tabId: number, isLoved: boolean) {
+function setLoveStatus(tabId: number, isLoved: boolean, isCommand: boolean) {
 	sendBackgroundMessage(tabId ?? -1, {
 		type: 'toggleLove',
 		payload: {
 			isLoved,
+			isCommand: true,
```
yayuyokitano, line 111: Also, this seems to ignore the argument passed to the function

yayuyokitano:
Important note: #4437 has been merged to master, please update this branch and add logic to support it

(set the isCommand/shouldShowNotification/whatever property to true for this new way of liking)

ahnesther:
> Thank you for the contribution! This implementation will not work on safari, as safari has no support for notifications API. I think we need to do a similar strategy to V2, with the icon temporarily changing (maybe we can do both actually with notification being disableable, I quite like the notification.) Implementation otherwise looks sound. One request: Could you rename the isCommand property to something like "shouldShowNotification"? This would play a lot nicer with #4437 which is not a command, but which I think should also be using this.

Could you tell me a bit more about the V2 strategy? I quickly read over the V3 strategy, but didn't see one for V2.

yayuyokitano:
> > Thank you for the contribution! This implementation will not work on safari, as safari has no support for notifications API. I think we need to do a similar strategy to V2, with the icon temporarily changing (maybe we can do both actually with notification being disableable, I quite like the notification.) Implementation otherwise looks sound. One request: Could you rename the isCommand property to something like "shouldShowNotification"? This would play a lot nicer with #4437 which is not a command, but which I think should also be using this.
> 
> Could you tell me a bit more about the V2 strategy? I quickly read over the V3 strategy, but didn't see one for V2.

The V2 implementation can be seen here as mentioned by creator of the issue this PR is linked to https://github.com/web-scrobbler/web-scrobbler/commit/a945a4f2715d59aba16df1736399156dd26e399d
This was lost with the migration to V3.

For icons, [I think these should work](https://drive.google.com/drive/folders/1n2z4MzDZVnF3nFPmHKnV0aWg7IuCXL7Y)
Place these in "src/icons/monochrome" with appropriate name and I think you can set color to same as action_base.svg (monochromeColors constant in "scripts/generate-icons.ts")

yayuyokitano:
Hit the wrong button – don't mind me😅

src/util/notifications.ts
```diff
@@ -8,6 +8,7 @@ import { Scrobbler } from '@/core/object/scrobble-service';
 import { debugLog } from '@/core/content/util';
 import * as BrowserStorage from '@/core/storage/browser-storage';
 import NativeScrobblerNotification from '@/core/storage/native-scrobbler-notification';
+import { Console } from 'console';
```
yayuyokitano, line 11:
```suggestion
```
looks unused

ahnesther:
> > > Thank you for the contribution! This implementation will not work on safari, as safari has no support for notifications API. I think we need to do a similar strategy to V2, with the icon temporarily changing (maybe we can do both actually with notification being disableable, I quite like the notification.) Implementation otherwise looks sound. One request: Could you rename the isCommand property to something like "shouldShowNotification"? This would play a lot nicer with #4437 which is not a command, but which I think should also be using this.
> > 
> > 
> > Could you tell me a bit more about the V2 strategy? I quickly read over the V3 strategy, but didn't see one for V2.
> 
> The V2 implementation can be seen here as mentioned by creator of the issue this PR is linked to [a945a4f](https://github.com/web-scrobbler/web-scrobbler/commit/a945a4f2715d59aba16df1736399156dd26e399d) This was lost with the migration to V3.
> 
> For icons, [I think these should work](https://drive.google.com/drive/folders/1n2z4MzDZVnF3nFPmHKnV0aWg7IuCXL7Y) Place these in "src/icons/monochrome" with appropriate name and I think you can set color to same as action_base.svg (monochromeColors constant in "scripts/generate-icons.ts")

Thanks for the link, looks really useful 👍 Sorry for the silence on my end, but good news is that I've got some commits (https://github.com/web-scrobbler/web-scrobbler/pull/4472/commits/2377908c06f84a5956c8d8b270a7a8633af50dda, https://github.com/web-scrobbler/web-scrobbler/pull/4472/commits/5d041ad8870df7efb78a0672de5d189ddbb38757) that only shows the loved/unloved notification once (ie. liking a track with the hotkeys more than once doesn't trigger the notification, same with unliking a track, hitting the dislike/unlike button on youtube (for example), etc). 

Going to work on the icon changes now. Thanks for your help and patience!

yayuyokitano:
Cool! Heads up, #4484 has high risk of conflict, if it is merged before this.

yayuyokitano:
Oh also, if you update your branch against master now you should see halving (or more) of build times when using npm run dev
shouldnt be any conflicts I think.

yayuyokitano:
Could you:

- Change controller.mode into a private variable (should already not be used outside connector anyway
- Ensure that controller.mode is not used anywhere within the connector directly except for setting values to it, use getMode() instead
- Then, we can make it so that getMode() returns the temporarymode if there is one, and if not it returns regular mode. We can then avoid mutating mode variable at all when setting temporary mode. This seems less error-prone to me in the case that we have mode changes occuring while there is a temporary mode active.

src/core/object/controller/controller.ts
```diff
@@ -683,6 +711,41 @@ export default class Controller {
...
+			this.mode = this.tempMode;
+		}
+		const TEMP_ICON_DISPLAY_DURATION = 5000;
+		this.setMode(newMode);
+		this.timeoutId = setTimeout(() => {
+			this.timeoutId = undefined;
+			this.setMode(pastMode);
+			this.tempMode = null;
+		}, TEMP_ICON_DISPLAY_DURATION);
+	}
```
yayuyokitano, line 747:
```suggestion
	private setTempMode(
		pastMode: ControllerModeStr,
		newMode: ControllerModeStr,
	) {
		if (this.isTempIconVisible()) {
			clearTimeout(this.timeoutId);
			this.timeoutId = undefined;
		}
		const TEMP_ICON_DISPLAY_DURATION = 5000;
		this.tempMode = newMode;
		this.onModeChanged();
		this.timeoutId = setTimeout(() => {
			this.timeoutId = undefined;
			this.tempMode = null;
			this.onModeChanged();
		}, TEMP_ICON_DISPLAY_DURATION);
	}
```
Untested, but we should be able to do something along these lines to completely avoid mutating this.mode and still get the changes shown.

ahnesther: https://github.com/web-scrobbler/web-scrobbler/pull/4472/commits/80e606639a2728ec0d3f23dd729de37470c2868e

src/core/object/controller/controller.ts
```diff
@@ -902,7 +958,7 @@ export default class Controller {
 				void this.setSongNowPlaying();
 			} else {
 				// Resend current mode
-				this.setMode(this.mode);
+				this.setMode(this.getMode());
```
yayuyokitano, line 961:
```suggestion
				this.onModeChanged();
```
To avoid mutating this.mode

ahnesther: Good catch, thanks for catching it 🥲 https://github.com/web-scrobbler/web-scrobbler/pull/4472/commits/6fdfbcf6206d1fe8a6d0afdc21a4d2611d9158bc

yayuyokitano:
#4484 was merged, it conflicts.

When fixing the conflict, in getMode() please provide following priority:

If there is tempmode, apply tempmode, whether song is paused or not.

If no tempmode, apply paused if paused and apply mode if not paused. (this logic is implemented by #4484)

ahnesther:
> This seems mergeable to me now, and it is working in safari macos and ios as well. Is there anything you feel is missing or iffy?

Perfect! I had no other concerns -- I just wasn't able to test on safari macos/ios. Thanks for your all help and review 😸

yayuyokitano (approved):
Thank you for the contribution🚀

PR #4477, opened 2024-01-31
Add NTS Radio connector

HarrisonWilde:
Adding a connector for NTS radio (requires paid membership). This is a simple approach that works but requires the user to manually select the channel they are listening to in the nts.live/live-tracklist window they have open (the connector is only associated with this URL). There is no play / pause functionality as a result, meaning a user must manually close the live tracklist when they are done.

NOTE: I have included some commented code which I thought should stop tracks being repeatedly scrobbled if a user changes from and then back to a channel with the same song still playing. In practice this behaved a bit strangely and sometimes the initial channel then wouldn't scrobble at all. Is there any need for this anyway (I can't tell if the extension already has some way of stopping this from happening, but refreshing a page etc seems to cause repeats no matter what).

**Additional context**
I have had a long conversation with @yayuyokitano about this on the discord. I believe there will be a more complex way that potentially works better than this, but for now this works reasonably well and with the caveats above I think this is of interest to users of web scrobbler.


src/connectors/nts.ts
```diff
@@ -0,0 +1,36 @@
...
+		if (artistElement) {
+			artist = artistElement.textContent || '';
+		}
+		if (trackElement) {
+			track = trackElement.textContent || '';
+		}
+	}
+
+	return { artist, track };
+};
```
yayuyokitano, line 29:
```suggestion
Connector.artistSelector = '.page-live-tracks__list li:first-child .page-live-tracks__artist-title';

Connector.trackSelector = '.page-live-tracks__list li:first-child .page-live-tracks__song-title';
```
We already have internally that should be equivalent to this, I suspect this might work out of the box (though test it first).

HarrisonWilde:
Yes @yayuyokitano it does work, just checked.

yayuyokitano:
Ah that one is probably my bad, could you run `npm run prettierfix` and it should fix up the linter error

HarrisonWilde:
Oops will fix lint in a min

yayuyokitano:

PR #4478, opened 2024-02-01
Update CI and update code accordingly

yayuyokitano:
Updates stylelint to version 16
Updates prettier
Also, we now require type imports always when importing types, through eslint config change.
Draft, currently relying on local configs, awaiting publish of updated eslint and stylelint configs


jaccarmac:
Are the non-stylelint-upgrade changes still useful?

PR #4479, opened 2024-02-01
Don't typecheck connectors at build time

yayuyokitano:
Builds connectors using esbuild instead of tsc, ignoring typechecking.
I am seeing build speed increases of ~40% on my end from ~5s to ~3s (~2s when hot reloading)
Typechecking is still handled by checkts, and presumably by the developers IDE.
At this point the bottleneck of builds seems to be compressing images, which I don't know if we can do anything about (maybe we can entirely drop it in dev build reload for further speed increase when hot reloading?)

### DIFF
--- a/scripts/compile-connectors.ts
+++ b/scripts/compile-connectors.ts
@@ -13,7 +13,7 @@ import chokidar from 'chokidar';
 function generateConnectors() {
 	return new Promise<void>((resolve, reject) => {
 		exec(
-			'esbuild src/connectors/*.ts --bundle --outdir=build/connectorraw',
+			'esbuild src/connectors/*.ts --bundle --outdir=build/connectorraw --tsconfig=tsconfig.connectors.json',
 			(err, stdout, stderr) => {
 				if (err) {
 					colorLog(err, 'error');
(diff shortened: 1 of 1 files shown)


yayuyokitano:
One caveat is that we lose source maps for the connectors specifically with this approach, but I don't think this is an issue, as we also drop minification with this approach.

scripts/compile-connectors.ts
```diff
@@ -13,7 +13,7 @@ import chokidar from 'chokidar';
 function generateConnectors() {
 	return new Promise<void>((resolve, reject) => {
 		exec(
-			'tsc --project tsconfig.connectors.json',
+			'esbuild src/connectors/*.ts --bundle --outdir=build/connectorraw',
```
inverse, line 16: Can't we still use the `tsconfig` with esbuild?

See: https://esbuild.github.io/api/#tsconfig

yayuyokitano: Yeah this is a good call. We still have to specify entrypoint though, as esbuild does not care about what typescript says should be included, it takes whatever entrypoints you give it and transforms from there bringing whatever imports it encounters with it.

inverse (approved):

PR #4480, opened 2024-02-01
Add NTS Shows connector

HarrisonWilde:

This adds support for scrobbling (for premium users who have access to tracklistings) for NTS Radio's collection of archived shows.

**Additional context**

I manually specify all NTS URLs except for `*nts.live/live-tracklist/*` as this is used for live radio scrobbling. Is there a more concise way to express this?

Currently, player selector is a wrapper containing the player bar and the live tracklist. I am not sure if this is the correct approach as the progress bar / time / moving waverform next to the current playing track seems to trigger a state change very frequently. Is this a bad thing or not? I am not sure if the player should somehow just be the play / pause button and the current playing track name, or if all these updates are inconsequential? I.e. in the below, when playing, state changes are triggered all the time, but not when paused as nothing is moving then.

image

image


yayuyokitano (approved):
LGTM, thank you for the contribution! The id of an existing connector is being changed which is normally reason not to merge, but since the connector hasn't actually been pushed to production yet I'll allow it as-is.

PR #4483, opened 2024-02-01
Add album guessing options

### LINKED ISSUE #1575: Disable album guessing
I would like to request an option to disable album auto-completion. On YouTube it often guesses some compilation for single tracks and I don't want that. Scrobbling without the album field IS possible for example by Open Web Scrobbler and then album which Last.fm suggests doesn't appear in my library. However, when using this addon it does.

### LINKED ISSUE #2830: Automatically add album title to manual scrobbles
All the tracks I have manually edited (on Youtube) have empty album titles, and I think it would be cool if the web scrobbler automatically added the album title like it does for tracks that aren't edited. 

After manually editing a track, if the album title is left empty, the extension would automatically grab the album title (if one exists) from Last.fm and add it to the scrobble.

I could manually add all the album titles, but that would be a pain as I have around 500 edits. Also, the extension already automatically adds the album for tracks that aren't manually scrobbled (on Youtube), so I feel like this feature wouldn't be too hard to add.

yayuyokitano:
Adds three options under **advanced settings**

Disable album guessing: Lastfm will never supply album title from pipeline, only albums explicitly declared by the connector or supplied by user through edits will be used.

Only guess album for unedited tracks (default): Mirrors current behavior. Album is supplied from lastfm if there is none supplied by connector, but album is not supplied when user has edited the track.

Guess album title for all tracks: WIll supply album from lastfm even when user has edited track, as long as there is nothing in album field.

Reasoning for adding to advanced is that this is not really something most people will want to fiddle with, and option fatigue is becoming a real issue.

image

Closes #1575 
Closes #2830

### DIFF
--- a/src/core/object/pipeline/metadata.ts
+++ b/src/core/object/pipeline/metadata.ts
@@ -58,12 +58,12 @@ export async function process(
 				// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Set regardless of previous state
 				(song.processed[field] as any) = songInfo[field];
 			}
+		}
 
-			if (!song.getAlbum() || song.flags.isAlbumFetched) {
-				song.processed.album = songInfo.album;
-				song.noRegex.album = songInfo.album;
-				song.flags.isAlbumFetched = true;
-			}
+		if (await shouldCopyAlbum(song)) {
+			song.processed.album = songInfo.album;
+			song.noRegex.album = songInfo.album;
+			song.flags.isAlbumFetched = true;
 		}
 
 		for (const field of METADATA_TO_COPY) {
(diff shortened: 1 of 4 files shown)


src/core/object/pipeline/metadata.ts
```diff
@@ -58,12 +58,12 @@ export async function process(
...
 				(song.processed[field] as any) = songInfo[field];
 			}
+		}
 
-			if (!song.getAlbum() || song.flags.isAlbumFetched) {
-				song.processed.album = songInfo.album;
-				song.noRegex.album = songInfo.album;
-				song.flags.isAlbumFetched = true;
-			}
+		if (await shouldCopyAlbum(song)) {
```
inverse, line 63: Nice refactor :+1:

inverse (approved):
nice one!

PR #4484, opened 2024-02-01
Add pause state

### LINKED ISSUE #1333: Indicate paused state in page action
I think current behaviour (the same icon whether is media playing or not) is a bit confusing.

That could also help to distingush site issues ("it's scrobbling but it shouldn't" from "it indicated playing but it's paused").

yayuyokitano:
Adds pause state for when song is paused.
Operates outside normal mode structure, allowing it to override pausable modes and returning to the correct mode after.
Closes #1333 

image
image


inverse (approved):

inverse:
Nice one!

PR #4496, opened 2024-02-08
Update release-drafter.yml

inverse:
Will make `chore` ones be skipped.

e.g. https://github.com/web-scrobbler/web-scrobbler/pull/4485

otherwise they show up here

image

Unless you think chore things should be tagged differently?


yayuyokitano (approved):
Seems reasonable to me

PR #4497, opened 2024-02-08
Improve npm ci only on cache miss

inverse:
We can improve the performance by caching npm dependencies and restoring them between builds


inverse:
Closing as it doesnt really give much gains :cry:

PR #4499, opened 2024-02-09
Use audible api for determinining isPlaying status

### LINKED ISSUE #1516: Use tab audible API to detect the playing state?
For sites using a flash player, the current connector is usually unable to detect the playing state (e.g.#1515). For this purpose, is it possible to introduce the [tab audible API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab) (support by Chrome and Firefox since v45)?

### Expected behavior

Correctly detect the playing state of sites using a flash player.

### Current behavior

Assume the `isPlaying` state is always true.

### Possible solution

Introduce the [tab audible API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab) (support by Chrome and Firefox since v45).

yayuyokitano:
Closes #1516 
Solves this by adding a function Connector.useTabAudibleApi() which when called replaces Connector.isPlaying with a async fetcher.

I'm a bit unsure about the stability of this concept, but it seems to work. If a connector actually starts using it we can worry more about it I guess.

Also reworks communication.ts to stricten typing as it was needed to work properly with .then() syntax.

### DIFF
--- a/src/core/background/main.ts
+++ b/src/core/background/main.ts
@@ -487,6 +487,21 @@ setupBackgroundListeners(
 			};
 		},
 	}),
+
+	/**
+	 * Listener called by a content script to figure out whether it is currently audible
+	 */
+	backgroundListener({
+		type: 'isTabAudible',
+		fn: async (_, sender) => {
+			const tabId = sender.tab?.id;
+			if (typeof tabId !== 'number') {
+				return Promise.resolve(true);
+			}
+
+			return (await browser.tabs.get(tabId)).audible ?? true;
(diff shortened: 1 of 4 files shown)


inverse (approved):
Seems reasonable - wonder when we will get our first connector that needs it :)

yayuyokitano:
I’d bet that an extremely large percentage of our existing connectors without isPlaying or button selectors specified would be improved by this, but I am not planning on going through each one and trying.

PR #4500, opened 2024-02-09
Improve listenbrainz and maloja local login

yayuyokitano:
Previously login to these services with local instances was clunky, as you had to know you were supposed to input exact scrobble endpoint. Now almost anything works as long as you point it to the right general location.

Also, loving tracks should now work on local instances of listenbrainz, it did not previously.

Should be fully backwards compatible, so existing users will not need to change their login info.


inverse (approved):

ThinkChaos:
Would be nice to merge this, I just had to use this PR as a reference!
Thanks for all the work!

PR #4515, opened 2024-02-13
Fix funkwhale connector

inverse:
The demo one looked broken from https://open.audio/library/albums/

@yayuyokitano should make testing https://github.com/web-scrobbler/web-scrobbler/pull/4434 work nicer :)

### DIFF
--- a/src/connectors/funkwhale.ts
+++ b/src/connectors/funkwhale.ts
@@ -35,13 +35,13 @@ function setupOldPlayer() {
 function setupNewPlayer() {
 	Connector.playerSelector = '.controls-row';
 
-	Connector.playButtonSelector = '.player-controls .play';
+	Connector.pauseButtonSelector = 'button[aria-label=Pause]';
 
-	Connector.trackSelector = '.content .track';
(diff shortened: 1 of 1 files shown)


yayuyokitano:
I'm a little bit afraid that this might break support for older instances, since funkwhale is self hosted its more relevant than most sites.

But most instances seem to already be up to date so this is definitely the more important case to cover, and I'm not sure what we can use to detect each player type.

inverse:
Perhaps @agateblue has an opinion on this?

agateblue:
Hi there! I'm not a Funkwhale maintainer anymore, but maybe you can try asking @georgkrause about this ☺️

georgkrause:
We updated Vue to version 3 in 1.3.0 and this likely led to different html (as far as I understand, you are handling the html). We are probably going to break your scrobbler again since we are working towards a redesign, however this is likely going to take a while. Using the aria-labels as selector might be the most stable thing to do, however no guarantees!

yayuyokitano:
I guess we can just fire this off as is, and if we get someone complaining about their instance on some older version breaking we can worry about it then.

PR #4516, opened 2024-02-14
Update securenetsystems.ts — DOM Changed

spjpgrd:
Wasn't selecting the correct elements anymore. Duration was also changed from being displayed to being a `transition-duration` on a progress bar element.  

Was using [WWCD](https://streamdb8web.securenetsystems.net/cirruspremier/WWCD)'s stream to run against.

### DIFF
--- a/src/connectors/securenetsystems.ts
+++ b/src/connectors/securenetsystems.ts
@@ -1,32 +1,35 @@
 export {};
 
-const songOnNow = '.songLineOnNow';
+const songOnNow = '.songDetails';
 
 Connector.playerSelector = ['.playlistSongArea', '#playerDiv'];
 
 Connector.artistSelector = [
-	`${songOnNow} .songDetail .songArtist`,
-	'#now-playing-artist',
+	`${songOnNow} .songInfo .songArtist`,
 ];
 
 Connector.trackSelector = [
-	`${songOnNow} .songDetail .songTitle`,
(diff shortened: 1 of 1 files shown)


src/connectors/securenetsystems.ts
```diff
@@ -1,32 +1,35 @@
...
-	`${songOnNow} img.songCover`,
-	'#now-playing-album-art',
+	`${songOnNow} img.songCover`
 ];
 
-Connector.durationSelector = `${songOnNow} .songDuration`;
+Connector.getDuration = () => {
+	// Get the computed style of the div
+	let computedStyle = window.getComputedStyle('.progressBar');
+	let durationInSeconds = computedStyle.getPropertyValue('transition-duration');
```
yayuyokitano, line 24: This doesn't work. Try using `const durationInSeconds = Util.getCSSPropertyFromSelectors('.progressBar', 'transition-duration')`

yayuyokitano:
Thank you for the contribution! See comment above.
Will need to do some testing on this to see if it works across other sites, as based on the old code I suspect they might not all have had the exact same dom in the past. See #3470

spjpgrd:
@yayuyokitano thanks for the review. I'll take a look later at [#3470](https://github.com/web-scrobbler/web-scrobbler/pull/3470) to do a proper vetting, and add back in whatever is necessary 👍

Issue #4024, opened 2023-08-12
Don't scrobble tracks set to private 


kresskress:
Artists share unreleased music, demos etc. via private or unlisted Soundcloud, Youtube, etc. links. Currently the Web-Scrobbler does scrobble music from these links. 

Ideal new behaviour would be that the web-scrobbler does not, by default, scrobble music from links set to private/unlisted.  Individual users could turn on the option to turn on scrobbles from private links under General options or individually under each connector. 

I don't know if this was already implemented at some point and stopped working.
The idea was raised by @cenobitedk in 2014.
https://github.com/web-scrobbler/web-scrobbler/issues/333
https://github.com/web-scrobbler/web-scrobbler/pull/367

ahnesther:
Hello! There is a tiny little PR that addresses non-scrobbling of private tracks on Soundcloud here: https://github.com/web-scrobbler/web-scrobbler/pull/4519

I'll also mention that there was a comment about whether adding non-scrobbling of private/unlisted youtube videos is necessary here: https://github.com/web-scrobbler/web-scrobbler/pull/4519#issuecomment-1949570371 . While I'm also not sure about how and if tracks are shared through private/unlinked youtube links, I will play around with different options. In the meantime, it'd be best to just ignore/skip the song you are listening to.

PR #4519, opened 2024-02-16
Stop the scrobbler from scrobbling privately shared Souncloud tracks

ahnesther:
Addresses issue https://github.com/web-scrobbler/web-scrobbler/issues/4024.
Should stop the scrobbler from scrobbling from private Soundcloud, and private/unlisted youtube videos.

**Additional context**
It was easier to stop the scrobbling from Soundcloud as the private css selector for private tag is much better defined than Youtube's. I didn't tackle embedded Soundcloud as there isn't a connector for them... yet. I'm sure I'll run into the same problem below.
For Youtube, there are two cases: directly watching the private/unlisted video in youtube.com or watching an embedded youtube video. The first case seems a little simpler, where a file is injected into youtube.ts to get the `window.ytInitialPlayerResponse` variables (`isPrivate`, `isUnlisted`). The second cases seems much more difficult since the window variables aren't available in iframe/embedded videos. 
Any suggestions, @yayuyokitano ?

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -671,6 +671,10 @@
         "message": "This song is not being scrobbled",
         "description": "The header of 'disallowed' popup"
     },
+    "disallowedDescIsPrivate": {
+        "message": "The song is not being scrobbled, as it is marked private by the owner. ",
+        "desciption": "The description of 'disallowed' popup when a song that is marked as private is being played."
+    },
     "disallowedDesc1": {
         "message": "Web Scrobbler has detected the song \"$2\" by \"$1\", but will not scrobble it.",
         "description": "The description of 'disallowed' popup. $1 is the artist detected. $2 is the track detected"
--- a/src/connectors/soundcloud.ts
+++ b/src/connectors/soundcloud.ts
@@ -89,6 +89,12 @@ Connector.getOriginUrl = () => {
 	return Connector.getUniqueID();
 };
 
+Connector.scrobblingDisallowedReason = () => {
+	if (document.querySelector('.sc-label-private')) {
+		return 'IsPrivate';
+	}
+};
+
 const filterArtistPremiereRules = [
 	{ source: /^\s*Premiere.*:\s*/i, target: '' },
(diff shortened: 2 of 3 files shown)


yayuyokitano:
First of all, you must add an explanatory message for locale entry "disallowedDescIsPrivate" explaining reason song was disallowed. You can see the other ones like "disallowedDescIsAd" etc for examples.

I can kinda see two paths for this. The first path is to only do this for soundcloud, and not do any option about it. The other is to also do it for youtube, and add an option for it. IMO while unreleased tracks may frequently be shared as privated tracks in soundcloud, but I just dont see unlisted youtube videos being a common way of sharing unreleased tracks.

Personally I see the former option, being opinionated and only adding this to soundcloud and having no option, probably being the better option, as we generally aim to be "plug and play". I'm not completely firm in this though if you have reasons to the contrary.

ahnesther:
> First of all, you must add an explanatory message for locale entry "disallowedDescIsPrivate" explaining reason song was disallowed. You can see the other ones like "disallowedDescIsAd" etc for examples.
> 
> I can kinda see two paths for this. The first path is to only do this for soundcloud, and not do any option about it. The other is to also do it for youtube, and add an option for it. IMO while unreleased tracks may frequently be shared as privated tracks in soundcloud, but I just dont see unlisted youtube videos being a common way of sharing unreleased tracks.
> 
> Personally I see the former option, being opinionated and only adding this to soundcloud and having no option, probably being the better option, as we generally aim to be "plug and play". I'm not completely firm in this though if you have reasons to the contrary.

I think for the context of this PR, I'll just do it for Soundcloud and later experiment with Youtube. I'm still not super confident with the codebase so I think it'd be better if I step back and immerse myself more in the codebase. 

Thanks so much for help!

ahnesther:
(had to force push as the tests were failing before, but they're now working, oddly enough)

yayuyokitano:
Merging this in current state, thank you for the contribution!

letruxux:
I'm still having this issue in a private playlist playing public songs. are you working on a fix or still investigating?

inverse:
The next release should fix it :crossed_fingers:

PR #4534, opened 2024-02-21
Draft new release template

inverse:

For the next version I thought it would be nice to have a refined release template to make it clearer. Mostly to seperate new functionality vs new connectors.

Only thing not fully captures is non-connector fixes? Do you think we should handle that?

**Additional context**

I'll need to create the new label's if you agree before merging :+1:

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,7 +1,7 @@
 name-template: 'v$RESOLVED_VERSION'
 tag-template: 'v$RESOLVED_VERSION'
 categories:
-  - title: '🎁 New feature'
+  - title: '🎁 New features'
     label: 'new-feature'
   - title: '🚀 New connectors'
     label: 'new-connector'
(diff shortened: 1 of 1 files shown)


inverse:
@yayuyokitano what do you think? We could also leverage https://github.com/amannn/action-semantic-pull-request

So that we enforce PR titles follow the conventions to automatically label too

yayuyokitano:
I don't really like semantic pull request linting, as it enforces a pretty strict title format on PRs, rather than label based system where we the maintainers just set up the right labels according to our own discretion. Especially for newer contributors that may be inexperienced I think the current approach is better.

I also think a core fix label is needed, given we have those happen fairly often, i think we have 2-3 updates recently including core fixes.

inverse:
Makes sense for this project indeed.

Added Core fixes - not sure if we need a core updated as updated would cover both updates for core and connectors?

inverse:
will make the label changes

PR #4538, opened 2024-02-23
restrict love/unlove to playing and scrobbled modes

yayuyokitano:
In some cases, love/unlove event may be triggered when it shouldn't. This should eliminate that
Check goes straight on the private mode variable, so it ignores temp modes pause/love/unlove


inverse (approved):
Nice one - I actually noticed that today on YouTube when scrobbling wasnt allowed but it showed.

Issue #4536, opened 2024-02-23
Love/unlove track in player not working in YTM and Niconico


azenhawke:
**Describe the bug**
The feature to love/unlove a track in music player from the new update seems to only work on the normal Youtube site, and not the Youtube Music webplayer (or Niconico).

**How to reproduce**
1. Go to [Youtube Music](https://music.youtube.com/) (or [Niconico](https://www.nicovideo.jp/))
2. Play any track
3. Like it (in my case, it causes nothing to change on Last.FM)
4. Play and like the same track on [Youtube](https://www.youtube.com/) (which, if reproduced, should actually mark the song as loved)

**Expected behavior**
From what I can tell, liking a track on the YTM webplayer (and probably Niconico?) is intended to have the same effect as liking music on Youtube.

**Screenshots**
Example of a song liked on YTM not translating to Last.FM:
![ytm scrobble bug 1](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/a2d3b5de-99bc-4fba-89c8-71e3962632b2)
![ytm scrobble bug 3](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/8850e59f-066f-4273-8511-588704ef76ad)
![ytm scrobble bug 2](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/4f79ae9b-4b73-4226-91e9-a028627a5fe0)

Example of a song hearted on Niconico not translating to Last.FM:
![niconico scrobbler bug 1](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/2c53910c-d090-483a-b237-b853834d98d9)
![niconico scrobbler bug 2](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/ad6e2305-d976-4988-be2d-0ec6cab3ac61)
![niconico scrobbler bug 3](https://github.com/web-scrobbler/web-scrobbler/assets/114738870/7ec00cfc-3681-4cee-acdf-95463445c4f6)

**Environment**
-   OS: Windows 10
-   Browser: Chrome
-   Extension version: 3.6.0

**Additional context**
Spotify also works perfectly fine for in-player loving tracks. YTM and Niconico are the main music sites I use, so I can't confirm if any other sites work. I did make sure the setting was on for both platforms.

ahnesther:
[...]

PR #4548, opened 2024-02-25
Add 'Like' selector for youtube music

ahnesther:
Partly addresses the issue, but for youtube music: https://github.com/web-scrobbler/web-scrobbler/issues/4536
Added a 'Like' selector for Youtube Music.

**Additional context**
I tried doing the same with Niconico but the selectors aren't helpful and using different selectors to get a normal Love/Unlove response did not work. For example:
image

image
For now, I need to think about Niconico but it doesn't seem like there aren't any good, simple UI methods.

### DIFF
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -56,6 +56,9 @@ const playingPaths = [
 	'M6 19h4V5H6v14zm8-14v14h4V5h-4z',
 ];
 
+Connector.loveButtonSelector =
+	'#like-button-renderer #button-shape-like button[aria-pressed="false"]';
+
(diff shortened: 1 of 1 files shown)


yayuyokitano:
Looked at niconico a bit. It is changing the icon, so i looked into how this is done and it seems it is updating xlink:href of use element in svg to point to new ID.
Didn't test this much, but I think `'use[*|href="#LikeIcon_svg__frame-0"]'` might be a reliable value for Connector.loveButtonSelector maybe (namespacing is a bit cursed in querySelector)

jaccarmac:
Given #4635, can we close this and see a new PR for the other half of #4536?

PR #4550, opened 2024-02-27
Fix SiriusXM new url structure for Canada (and other regions) player

brodie-g:
SiriusXM url structure changed with recent update. Canada player URL is now 'https://can.siriusxm.com/player/'. This should be able to support any sub-region using this new structure (not that there are any other regions, SXM is only US and Canada, ATM).


inverse (approved):
Thanks!

PR #4586, opened 2024-03-16
Make love/unlove checks stricter

### LINKED ISSUE #4564: Track gets loved when I minimize it
**Describe the bug**
Every time I minimize a music video(like when i search another video while this one is running), I get a notification that my track is added to my "Loved tracks" list. When I return the video to take up the whole page again, it unloves the track (and I get a notification for it). 

**How to reproduce**
1. Go to youtube.com
2. Play any song.
3. Click on the youtube logo so that you open the home page while the song is running as a small window.
4. You should get a notification that you loved the track that is running.
5. You can return to the video in full view again to unlove it

**Environment**
Browser: Chrome
OS: Windows 10
Extension version: 3.6.2

yayuyokitano:
By making a strict requirement for the button to actually transition directly from love to unlove, we effectively stop false positives.

We might be able to get away with a slightly less strict version that allows a null value to exist between false/true and still love/unlove.

Closes #4564 #4589

### DIFF
--- a/src/connectors/spotify.ts
+++ b/src/connectors/spotify.ts
@@ -48,6 +48,8 @@ Connector.getOriginUrl = () => getTrackUrl();
 
 Connector.loveButtonSelector = `${playerBar} button[data-testid="add-button"][aria-checked="false"]`;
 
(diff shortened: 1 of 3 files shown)


inverse (approved):
LGTM - interesting find!

PR #4587, opened 2024-03-16
Rework youtube music connector

### LINKED ISSUE #4561: Double artsit name | YT Music
**Describe the bug**
when scrobbling in youtube music, probably when playlist repeat is enabled, artist names are duplicated
**How to reproduce**
I don't understand how to reproduce it exactly, because I was just listening to music in the background and noticed it for the first time about two days ago
I thought I had to reproduce it that way, but it only worked once out of five tries:
1. enable track list repeat
2. reload page
3. play any playlist, mix, album
4. see double artist name

**Expected behavior**
scrobble without duplicates

**Screenshots**
image
I was just listening to one album here and the double names started scrobbling.

image
here I was listening to one artist, then through search I turned on an another artist and the names started duplicating

image
then I turned on another track from the main page and the duplication disappeared, then I turned on the album through the search. the first two tracks were fine, and then the duplication started again.
image

**Debug logs**
debug mod is enabled now, so far the error has not recurred. i will provide logs when it does.

-   OS: Arch Linux, Hyprland
-   Browser: Firefox
-   Extension version: 3.6.1

### LINKED ISSUE #4571: [YouTube Music] Scrobbling broken on new UI for small windows.
**Describe the bug**
YouTube Music has a new minimal UI when running on a browser window that's small, completely changing the title and artist layout and removing the album information, preventing anything from being scrobbled.

**How to reproduce**
1. Go to music.youtube.com
2. Play any song
3. Resize window until the UI changes to a 'mobile music player' style (in my case, it appeared when window width <616px).

**Screenshots**
example of the new UI

-   OS: Windows 10
-   Browser: Chrome
-   Extension version: 3.6.2

yayuyokitano:
A lot of unnecessary complexity here as pointed out by https://github.com/web-scrobbler/web-scrobbler/issues/4571#issuecomment-1997836031

Scraps entire selector driven approach to use mediaSession API instead.

Still applies similar logic. When song being played is a youtube video, no album is provided. In this case, we process the mediaSession.metadata.title as a youtube title, trying to get both artist and track from it, and using the artist field (channel) as the artist if it fails. If there is an album (means actual well structured upload), we just use mediaSession metadata with only slight modifications.

This is a little bit scary given the size of the changes to such a big platform, but it really does seem way more reliable than it was.

Closes #4571 
Closes #4561

### DIFF
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -24,13 +24,16 @@ const trackArtSelector = '.ytmusic-player-bar.image';
 
 const artistSelectors = [
 	// Base selector, combining both new and old
-	'.ytmusic-player-bar.byline [href*="channel/"]:not([href*="channel/MPREb_"]):not([href*="browse/MPREb_"])',
+	'ytmusic-app-layout>ytmusic-player-bar .ytmusic-player-bar.byline [href*="channel/"]:not([href*="channel/MPREb_"]):not([href*="browse/MPREb_"])',
 
 	// Old selector for self-uploaded music
-	'.ytmusic-player-bar.byline [href*="feed/music_library_privately_owned_artist_detaila_"]',
+	'ytmusic-app-layout>ytmusic-player-bar .ytmusic-player-bar.byline [href*="feed/music_library_privately_owned_artist_detaila_"]',
 
 	// New selector for self-uploaded music
-	'.ytmusic-player-bar.byline [href*="browse/FEmusic_library_privately_owned_artist_detaila_"]',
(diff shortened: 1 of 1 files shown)


yayuyokitano:
Btw known issue: this leaves us with the way youtube handles multiple artists, which tends to be artist1, artist2, & artist3. Previously, we used artist1, artist2, artist3. This could probably pretty reliably filtered using metadatafilter. Should we?

inverse (approved):
Sounds like it could be a nice addition the filter - but I dont think it's blocking this. Nice one!

PR #4601, opened 2024-03-20
scrobblerad.io connector

jbwharris:
New connector for scrobblerad.io. This is a project I've built recently with the goal of making a bunch of different radio stations scrobbleable and easily playable on mobile.


inverse:
Thanks!

PR #4629, opened 2024-04-06
Fix for fritz connector

### LINKED ISSUE #4187: Fritz connector is not working anymore
The problem is the connector seems to be outdated.

Also the livestream website doesn't display the current playback information anymore.
So we can't scrape the track and artist...

Ceddicedced:
Fixed and tested the fritz connector

All selectors where overhauled so they are modern and working again.

Resolves  #4187

### DIFF
--- a/src/connectors/fritz.ts
+++ b/src/connectors/fritz.ts
@@ -14,14 +14,14 @@ const filter = MetadataFilter.createFilter({
 		// and return the first element (the part of the string before the semicolon).
 
 		// Replace .feat with a semicolon
-		artist = artist.replace('.feat', ';');
+		const updatedArtist = artist.replace('.feat', ';');
 
 		// Example: "Artist1; Artist2" -> "Artist1"
-		if (artist.includes(';')) {
-			return artist.split(';')[0].trim();
+		if (updatedArtist.includes(';')) {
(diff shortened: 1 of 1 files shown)


src/connectors/fritz.ts
```diff
@@ -1,9 +1,28 @@
...
-Connector.isPlaying = () => Util.hasElementClass('.player', 'player_playing');
+Connector.pauseButtonSelector = '.audioplayer-playing';
+
+const filter = MetadataFilter.createFilter({
+	artist: (artist) => {
+		// If the artist string contains a semicolon, split the string by the semicolon
+		// and return the first element (the part of the string before the semicolon).
+
+		// Replace .feat with a semicolon
+		artist = artist.replace('.feat', ';');
```
inverse, line 17: Looks like you need to fix this by perhaps using another variable within the function. 

e.g. `const updatedArtist = artist.replace('.feat', ';');` and using this within the function?

inverse (changes requested):

src/connectors/fritz.ts
Ceddicedced: Yeah tbj i just ignored the eslint error. But i fixed it accordingly.

inverse (approved):
Thanks for the contribution!

PR #4651, opened 2024-04-17
fixed vk-save connector

gasievt:
**Fixed vk-save connector**
vk-save changed their name and url to kissvk.me so the connector stopped working


src/core/connectors.ts
```diff
@@ -2341,10 +2341,10 @@ export default <ConnectorMeta[]>[
...
 	},
 	{
-		label: 'vk-save',
-		matches: ['*://*.vk-save.com/*'],
-		js: 'vk-save.js',
-		id: 'vk-save',
+		label: 'kissvk',
+		matches: ['*://*.kissvk.me/*'],
+		js: 'kissvk.js',
+		id: 'kissvk',
```
inverse, line 2347: @yayuyokitano should this be changed if the site is technically the same? Won't it mean any config will be lost?

inverse (commented):
Change seem legit aside the one comment around the ID change as we use this for configuration changes for the connector so that will be lost even though the connector is technically the same. 

We should probably generate GUID's for these

src/core/connectors.ts
gasievt: @inverse Hello. They changed their url and name back to vk-save.com. Should I close the pull request?

inverse: Changing the ID would mean any manual changes etc under the old site would be lost. let wait to see what others suggest...

inverse:
Missed your previous comment @gasievt let close this if they changed it back. Feel free to open again if you notice any other changes :pray:

PR #4670, opened 2024-04-25
Update RTL+ Musik connector

opcode81:

  * Relax site URL to avoid site detection problems when the site is initially opened without a `/musik/*` suffix
  * Fix invalid attribute access in `isPlaying` for case where player is not yet fully loaded (caused exception which was not recovered from)

**Additional context**


jaccarmac:
Thanks!

PR #4686, opened 2024-05-06
Fix connector for Radio Caprice

projecteurlumiere:
Fix connector for http://radcap.ru

I will contact them soon on this nth child css selector mess and, possibly, make another PR if they make a better selector for play/pause button. Last time I emailed them they were very responsive.


jaccarmac:
Thanks!

PR #4691, opened 2024-05-08
Fix piped.ts (connector)

Tamarindo94:
Fixed piped connector ~~by ditching querySelector, which often returned null. Util.getAttrFromSelectors is now used instead~~

UPDATE: Util.getAttrFromSelectors wasn't the correct solution - querySelector is still used, but now the connector avoids trying to access properties on null objects that are querySelected before the video is added to the DOM (which was the reason for the connector crashing)

### DIFF
--- a/src/connectors/piped.ts
+++ b/src/connectors/piped.ts
@@ -11,13 +11,13 @@ Connector.getArtistTrack = () => {
 };
 
 Connector.getCurrentTime = () => {
-	const videoEl = document.querySelector('.shaka-video')
-	return videoEl ? videoEl.currentTime : null
+	const videoEl = document.querySelector('.shaka-video');
+	return videoEl ? videoEl.currentTime : null;
 }
 
 Connector.getDuration = () => {
-	const videoEl = document.querySelector('.shaka-video')
-	return videoEl ? videoEl.duration : null
+	const videoEl = document.querySelector('.shaka-video');
+	return videoEl ? videoEl.duration : null;
 }
 
 Connector.getUniqueID = () => {
--- a/src/connectors/piped.ts
+++ b/src/connectors/piped.ts
@@ -13,12 +13,12 @@ Connector.getArtistTrack = () => {
 Connector.getCurrentTime = () => {
 	const videoEl = (document.querySelector('.shaka-video') as HTMLVideoElement);
 	return videoEl ? videoEl.currentTime : null;
-}
+};
 
 Connector.getDuration = () => {
 	const videoEl = (document.querySelector('.shaka-video') as HTMLVideoElement);
 	return videoEl ? videoEl.duration : null;
-}
+};
 
 Connector.getUniqueID = () => {
 	const videoUrl = Util.getAttrFromSelectors('[aria-current=page]', 'href');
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks for the contribution!

Tamarindo94:
Hope it's ok now

src/connectors/piped.ts
```diff
@@ -10,11 +10,15 @@ Connector.getArtistTrack = () => {
...
 };
 
-Connector.getCurrentTime = () =>
-	(document.querySelector('.shaka-video') as HTMLVideoElement).currentTime;
-
-Connector.getDuration = () =>
-	(document.querySelector('.shaka-video') as HTMLVideoElement).duration;
+Connector.getCurrentTime = () => {
+	const videoEl = (document.querySelector('.shaka-video') as HTMLVideoElement);
+	return videoEl ? videoEl.currentTime : null;
```
inverse, line 15: Would using the `?.` operator be a more concice way of expressing this work?

```
Connector.getCurrentTime = () => (document.querySelector('.shaka-video') as HTMLVideoElement)?.currentTime;
```

Tamarindo94: It should be, yes. I wasn't sure whether the `?.` operator was discouraged by the project guidelines so I opted for the more verbose version. The only difference is that, with the `?.` operator, `undefined` will be returned if `querySelector` doesn't find the element, while the verbose version explicitly returns `null` in such case, and I don't know if that makes a difference to the core. If it doesn't, they should be functionally equivalent

inverse: Looks like there is one linting issue to fix.

Tamarindo94: I'd really like to fix it but I have no clue on how to do that...

inverse: If you run `npx prettier . --write` it should fix the issues automatically

src/connectors/piped.ts
```diff
@@ -10,11 +10,15 @@ Connector.getArtistTrack = () => {
 	return Util.processYtVideoTitle(text);
 };
 
-Connector.getCurrentTime = () =>
-	(document.querySelector('.shaka-video') as HTMLVideoElement).currentTime;
+Connector.getCurrentTime = () => {
+	const videoEl = (document.querySelector('.shaka-video') as HTMLVideoElement);
```
inverse, line 14:
```suggestion
	const videoEl = document.querySelector('.shaka-video') as HTMLVideoElement;
```

src/connectors/piped.ts
```diff
@@ -10,11 +10,15 @@ Connector.getArtistTrack = () => {
...
+	return videoEl ? videoEl.currentTime : null;
+};
 
-Connector.getDuration = () =>
-	(document.querySelector('.shaka-video') as HTMLVideoElement).duration;
+Connector.getDuration = () => {
+	const videoEl = (document.querySelector('.shaka-video') as HTMLVideoElement);
```
inverse, line 19:
```suggestion
	const videoEl = document.querySelector('.shaka-video') as HTMLVideoElement;
```

inverse (approved):
Thanks for the contribution!

PR #4760, opened 2024-06-13
Filter additional SiriusXM host names

dasevilla:
Adds additional SiriusXM host Twitter handles to filter out. These aren't actual music tracks.


inverse (approved):
Thanks for the contribution

PR #4761, opened 2024-06-13
Ensure filtered terms are lowercase for SiriusXM

dasevilla:
The filter terms need to be lowercase for them to match.


inverse (approved):
Thanks for a quick fix!

PR #4809, opened 2024-07-06
Add instant.audio connector

stephanebruckert:
`Connector.trackSelector` would have been fine by itself but the one radio I listen to adds a weird suffix, hence the extra code

And yes it's a lot of different domains, one for each country, all accessible from the main domain https://instant.audio/

Example pages:
- https://ecouterradioenligne.com/skyrock-la-nocturne/
- https://ecouterradioenligne.com/europe-2/
- https://radiosonline.be/top-belgium/
- https://radiosonline.be/instrumentals-forever/
- https://radios.co.cr/los-40-costa-rica/


src/connectors/instantaudio.ts
```diff
@@ -0,0 +1,22 @@
...
+
+Connector.trackSelector = '.playlist__song-name';
+
+Connector.trackArtSelector = '.playlist__img';
+
+Connector.pauseButtonSelector = '.btn-play #radio-button.stop';
+
+Connector.playButtonSelector = '.btn-play #radio-button.play';
+
+Connector.getTrack = () => {
```
inverse, line 15: Could be implemented as a filter

```js
const filter = MetadataFilter.createFilter({ track: removeTrackSuffix });

function removeTrackSuffix(text: string) {
	return text.replace(/ §\d+$/, '');
}
```

inverse (approved):

PR #4810, opened 2024-07-06
Fix warp-aphextwin connector

stephanebruckert:
Fix scrobbling artist with ` - Aphex Twin` instead of `Aphex Twin`

Example pages:
- https://aphextwin.warp.net/tracks
- https://aphextwin.warp.net/release/86954-aphex-twin-windowlicker

### DIFF
--- a/src/connectors/warp-aphextwin.ts
+++ b/src/connectors/warp-aphextwin.ts
@@ -2,6 +2,8 @@ export {};
 
 Connector.playerSelector = '.player';
 
+Connector.artistSelector = '.current-track .js-current-track-artist';
+
 Connector.trackSelector = '.current-track .js-current-track-title';
 
 Connector.durationSelector = '.current-track .js-current-track-duration';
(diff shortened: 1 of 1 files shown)


src/connectors/warp-aphextwin.ts
```diff
@@ -2,6 +2,8 @@ export {};
 
 Connector.playerSelector = '.player';
 
+Connector.artistSelector = '.current-track .js-current-track-artist';
```
inverse, line 5: Not sure we need this defined here as we're overriding the `getArtist()` function which would leverage this.

inverse (approved):
Aside the comment LGTM :+1:

src/connectors/warp-aphextwin.ts
stephanebruckert: I've added `Connector.artistSelector` for consistency with `Connector.trackSelector`, so we can quickly see where we are getting the data. `getArtist` then allows cleaning up what's in `artistSelector`

Issue #4786, opened 2024-06-24
Scrobble from Rayo - UK Radio Stations


cbrowne91:
Multiple radio stations are streamed via the Rayo service: https://hellorayo.co.uk 

This includes some of the biggest radio stations in the UK.

Please could this be added as a service. The song and artist is displayed in the play bar at the bottom.

ReduxGB:
It seems this is just a rebrand of planet radio, with a similar layout to the beta preview they displayed.

PR #4816, opened 2024-07-10
Add Rayo Support

ReduxGB:
Adds hellorayo.co.uk, after planet radio was rebranded.

**Additional context**
Original URLs are kept as some things are still operating on the URLs at this point in time.
Links to #4786

### DIFF
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -824,6 +824,8 @@ export default <ConnectorMeta[]>[
 		matches: [
 			'*://planetradio.co.uk/*/play/*',
 			'*://planetradio.co.uk/*/player/*',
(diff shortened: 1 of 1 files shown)


src/core/connectors.ts
```diff
@@ -824,6 +824,8 @@ export default <ConnectorMeta[]>[
 		matches: [
 			'*://planetradio.co.uk/*/play/*',
 			'*://planetradio.co.uk/*/player/*',
+			'*://hellorayo.co.uk/*/play/*',
```
inverse, line 827: Looks like the old URLs just redirect - shall we drop them?

inverse (approved):
Thanks for the contribution!

Issue #4808, opened 2024-07-06
Radio Free Fedi


candideu:
**Provide information about website**
radio free fedi is consent, agency and artist celebrating community radio from the Fediverse. It features music, spoken word, and programming from Fediverse artists.

https://radiofreefedi.net/

limebar:
duplicate of #4046 ?

candideu:
@limebar Weird, I could have sworn I didn't find any existing posts in my search. Closing.

Issue #4046, opened 2023-08-23
Connector for RadioFreeFedi?


gribbles67:
This is small web, consent driven, artist populated, non-commercial mechanism, attribution promoting, community radio for the fediverse.  Be great to have a connector for it...

https://radiofreefedi.net

It's all public facing, so no accounts needed to check it out...

Cheers,

.g

jbwharris:
@gribbles67 I have been considering making a connector for this lately. I think the main problem is the 3 different players on the page. I wish they had a more sophisticated player layout that allowed you to switch between players. I'm going to see if I can figure it out though.

gribbles67:
Yeah - they used to just have the main feed on that page.

It may be worth getting in touch with them for a chat. They're constantly growing & developing - and suspect they'd be happy to talk. WebScrobbler integration would be good for them and the artists, so it's in their interests to see if they can make life easier for you! ;-)


And who knows what they may be planning - a different UI with individual feeds may be scribbled on a whiteboard somewhere...

But pleased you're considering this - cheers!

jbwharris:
@gribbles67 I did actually DM them on Mastodon, like a crazy person, asking whether they were doing any type of redesign efforts currently. Didn't hear back, so no idea at this point. I thought of having a go at redesigning it myself, but also have about a billion other things on the go. That's why I like building connectors, they're usually micro projects that you can do quickly and move on.

candideu:
Hi! I'd love to see a connector for this!

They seem to be pretty active on Matrix: https://matrix.to/#/#radioFreeFedi:matrix.org

jbwharris:
Good news, I have a working connector for this now, I just need to do some more testing with regards to filtering the station IDs they slot into the feed. All 3 feeds work as expected. I should be able to get it pushed over this week once I've let it run for a bit.

PR #4824, opened 2024-07-15
New RadioFreeFedi connector

jbwharris:
Relating to #4046 #4808 

New connector for radiofreefedi.net. This page had 3 different players on it, so it took a bit of figuring to sort out, but it now plays whichever player is playing and pulls the relevant song, artist, album and art. They slot in a fair amount of station IDs into their feeds, so I've included some keyword filtering (same one as the SiriusXM connector) to have a running list of filtered terms as needed. 

image


inverse (approved):
Thanks for the contribution!

PR #4826, opened 2024-07-19
New Basspistol Radio connector

jbwharris:
https://radio.basspistol.com connector. A lot of the time the artist will have the URL to their website or Bandcamp, so that is filtered out. 

image

### DIFF
--- a/src/connectors/basspistol.ts
+++ b/src/connectors/basspistol.ts
@@ -1,0 +1,20 @@
+export {};
+
+const filter = MetadataFilter.createFilter({
+	artist: cleanupArtist,
+});
+
+Connector.playerSelector = '.radio-player-widget';
+Connector.trackSelector = '.now-playing-title';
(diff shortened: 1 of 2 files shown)


src/connectors/basspistol.ts
```diff
@@ -0,0 +1,20 @@
...
+Connector.trackArtSelector = 'img.album_art';
+Connector.isPlaying = () =>
+	Util.getAttrFromSelectors('.radio-control-play-button', 'aria-label') !==
+	'Play';
+
+Connector.applyFilter(filter);
+
+function cleanupArtist(artist: string) {
+	// remove brackets
+	return artist.replace(/^([^\(]+)(\(.*)$/, '$1');
```
inverse, line 19:
```suggestion
	return artist.replace(/^([^(]+)(\(.*)$/, '$1');
```

Will fix the lint warning

inverse (approved):
Thanks for the contribution!

PR #4868, opened 2024-08-08
gds.fm: Skip scrobbling announcements

rrrnld:
Stops scrobbling announcements made by gds.fm

**Additional context**

### DIFF
--- a/src/connectors/gds.fm.ts
+++ b/src/connectors/gds.fm.ts
@@ -7,3 +7,12 @@ Connector.artistSelector = '.player-meta__artist';
 Connector.trackSelector = '.player-meta__title';
 
 Connector.playButtonSelector = '.player-meta__icon--play';
+
+Connector.scrobblingDisallowedReason = () => {
+	const artist = Connector.getArtist()
+	if (artist != null && artist.match(/gds\.fm/)) {
+		return 'Announcement'
+	}
+
+	return null
+}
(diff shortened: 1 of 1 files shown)


src/connectors/gds.fm.ts
```diff
@@ -7,3 +7,12 @@ Connector.artistSelector = '.player-meta__artist';
 Connector.trackSelector = '.player-meta__title';
 
 Connector.playButtonSelector = '.player-meta__icon--play';
+
+Connector.scrobblingDisallowedReason = () => {
+	const artist = Connector.getArtist();
+	if (artist && artist.match(/gds\.fm/)) {
+		return 'Announcement';
```
yayuyokitano, line 14: Please use a valid DisallowedReason https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/object/disallowed-reason.ts

In this case 'FilteredTag' seems appropriate at a glance.

yayuyokitano (changes requested):
Thank you for contributing, there is one small change I would like for you to make.

src/connectors/gds.fm.ts
heyarne: You're right! Had issues with getting local dev to work, so I didn't find the expected types. Matches what other connectors use in this case.

rrrnld:
Hi there, is there anything I can do to get this merged?

PR #4885, opened 2024-08-18
Prepare for @types/webextension-polyfill @ 0.12

jaccarmac:
Unblocks #4882.


jaccarmac:
@yayuyokitano I'm not sure what the best way to address the type issues here is. The callback change seems to relate to https://github.com/DefinitelyTyped/DefinitelyTyped/commit/5394564b06fccc3bbe30fd98704bdf567af57a47#diff-04b57e3141aa3633559dfb78ae09cb0f551181104181c49a5773cddb1397d27eL60, but I can't figure out how that change applies to our specific case or why the associated changes in that PR are more trivial than `as`

PR #4974, opened 2024-10-14
added support for my website

EduardSkibidiGooner:
added /src/connectors/prigoana.ts
modified /src/core/connectors.ts


inverse:
Thanks for taking the tile to contribute! I tried to test the changes but couldn't get anything to scrobble when building from your branch.

image

PR #4996, opened 2024-10-25
Add support for Echo

sleepyfran:
Hey folks, I've recently released a new music player that can connect with multiple providers. It's available [here](https://echo.sleepyfran.me). This PR adds support to scrobble from there.

I've added the connector to the core connector list and added the connector file following the indications from the wiki. Since the page is developed with Lit, which heavily uses the shadow root, I couldn't use the built-in methods of the connector to query elements since shadow root elements are not queriable from the global window element query. In order for WebScrobbler to know when to start listening for changes I've added an interval that checks every two seconds for the player element to be there (which happens after initialization on the app). Let me know if there's any better way of doing this, I know it's not ideal. If we deem it too hacky I can trigger a custom event on the top level of the app when the player is available so that we could observe it better.

Since I control the whole app I've decided to go the easy route and emit a couple of custom events that I then listen from the connector: `track-playing` (including there the track, artist, album and provider info) and `track-paused`. Upon receiving one of those I keep the last-known state on a local variable and trigger an update.

I've also explicitly ignored tracks coming from Spotify to avoid double scrobbling, since those are covered via the Last.fm-Spotify integration. My first idea was to add this as a setting, but I didn't see any "connector-specific" setting other than the ones for YouTube and I think my project is way too tiny for it to deserve such a spotlight 😄 

**Additional context**
When a track is playing:
Playing

When a track is paused/there's nothing playing:
Paused

Thanks for this wonderful project! I initially thought of adding manual scrobbling support to the website until it hit me that I could just extend the project :^)

### DIFF
--- a/src/connectors/echo.ts
+++ b/src/connectors/echo.ts
@@ -1,0 +1,48 @@
+import type { TrackInfoWithAlbum } from '@/core/types';
+
+export {};
+
+let trackInfo: TrackInfoWithAlbum | undefined;
+
+const queryPlayerElement = () =>
+	document
+		.querySelector('app-root')
+		?.shadowRoot?.querySelector('app-header')
+		?.shadowRoot?.querySelector('echo-player');
+
+Connector.getTrackInfo = () => trackInfo;
+Connector.isPlaying = () => !!trackInfo;
+
+type EchoTrackInfo = {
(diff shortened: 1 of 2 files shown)


inverse (approved):
Thanks

PR #5023, opened 2024-11-11
Support new SomaFM player

sardaukar:
Change made:

- changed CSS selectors for track and artist name

Still to do:

- detecting whether player is playing or not 

Right now, the only different seems to be that the SVG on the play button changes, from

image

to

image

Fixes https://github.com/web-scrobbler/web-scrobbler/issues/4978

PR #5049, opened 2024-11-27
Fix for SomaFM player update

sardaukar:
SomaFM's new player has been updated, and broke [my previous PR to support it](https://github.com/web-scrobbler/web-scrobbler/pull/5023).

**Additional context**

Updated the CSS selectors, and am now matching on the SVG path for playing or not, since they changed the player to always show artist and track even when stopped (previous PR detected presence or not of this info to know if it was playing).


inverse (approved):
Changes seem reasonable  :+1:

PR #5058, opened 2024-12-05
Add support for Gizz Tapes

netclectic:
 - added connector for Gizz Tapes: `tapes-kglw.ts` - https://tapes.kglw.net/
 - added connector details to core/connectors.ts

### DIFF
--- a/src/connectors/tapes-kglw.ts
+++ b/src/connectors/tapes-kglw.ts
@@ -7,13 +7,14 @@ Connector.getArtist = () => {
 };
 
 Connector.trackSelector =
-	'#site-header div#now-playing > span#now-playing-track-title';
+	'#now-playing > #now-playing-track-title';
 
 Connector.albumSelector =
-	'#site-header div#now-playing > span#now-playing-show-date';
+	'#now-playing > #now-playing-show-date';
 
-Connector.currentTimeSelector =
-	'#site-header div#now-playing > span.now-playing-elapsed';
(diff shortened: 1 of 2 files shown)


src/connectors/tapes-kglw.ts
```diff
@@ -0,0 +1,27 @@
...
+Connector.albumSelector =
+	'#site-header div#now-playing > span#now-playing-show-date';
+
+Connector.currentTimeSelector =
+	'#site-header div#now-playing > span.now-playing-elapsed';
+
+Connector.durationSelector =
+	'#site-header div#now-playing > span.now-playing-duration hidden-sm';
+
+Connector.playButtonSelector = '#play-btn';
```
inverse, line 21: I don't think this is needed as you're overriding the isPlaying function below.

inverse (commented):
Seems reasonable aside from the one comment!

src/connectors/tapes-kglw.ts
netclectic: Good to know, updated.

inverse (approved):
Thanks for the contribution!

src/connectors/tapes-kglw.ts
inverse: The duration one could have been left in I think but perhaps could be updated to 


```
Connector.durationSelector =
	'#now-playing > span.now-playing-duration';
```

PR #5070, opened 2024-12-08
Pin stylelint-config-standard-scss to 11.0.0

FichteFoll:
**Describe the changes you made**

This commit/PR pins stylelint-config-standard-scss to 11.0.0 so that I can `npm install` locally.

Otherwise locking/resolving fails because 14.0.0 (latest) specifies `stylelint@"^16.11.0"`, which conflicts with `stylelint@">=15.10.0" from @web-scrobbler/stylelint-config@1.0.1`.

11.0.0 is the version currently in the lock file.

**Additional context**

```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: web-scrobbler@3.11.0
npm error Found: stylelint@15.11.0
npm error node_modules/stylelint
npm error   stylelint@"15.11.0" from the root project
npm error   peer stylelint@">=15.10.0" from @web-scrobbler/stylelint-config@1.0.1
npm error   node_modules/@web-scrobbler/stylelint-config
npm error     dev @web-scrobbler/stylelint-config@"1.0.1" from the root project
npm error
npm error Could not resolve dependency:
npm error peer stylelint@"^16.11.0" from stylelint-config-standard-scss@14.0.0
npm error node_modules/stylelint-config-standard-scss
npm error   peer stylelint-config-standard-scss@">=11.0.0" from @web-scrobbler/stylelint-config@1.0.1
npm error   node_modules/@web-scrobbler/stylelint-config
npm error     dev @web-scrobbler/stylelint-config@"1.0.1" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
npm error
```

PR #5089, opened 2024-12-23
Run CI on the latest Mac image.

jaccarmac:
Reverts 26cf7792309f515ea89ca9b0ad3eadbfe69f8841.

12 has been deprecated, breaking our builds. Pinning would be a preferable
strategy but this brings Mac builds back into line.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
 
   build-safari:
     name: Build the extension for Safari
-    runs-on: macos-12
+    runs-on: macos-latest
     steps:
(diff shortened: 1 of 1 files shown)


inverse:
it's not working :/ Wonder if its similar to this comment :https://github.com/web-scrobbler/web-scrobbler/pull/5070#issuecomment-2543051146

inverse:
Looks like something like this could work https://github.com/web-scrobbler/web-scrobbler/pull/5092

jaccarmac:
Oh, awesome! That's a less onerous fix than the ones I was able to find.

PR #5115, opened 2025-01-05
Make watcher type-nullable.

jaccarmac:
Unblocks #5044.


jaccarmac:
@yayuyokitano Does the `let` do anything with the inner `const`? Maybe my understanding of TypeScript shadowing rules is wrong, but this looks a little odd to me on review.

PR #5120, opened 2025-01-09
Add .com domain support for Yandex.Music connector

skroychan:
Added https://music.yandex.com/ to the list of domains, it's the one I'm using this service through but it was missing from the match list


inverse (approved):
Thanks for the contribution!

Issue #5094, opened 2024-12-26
cosine.club


RVH3000:
Hello,  
could you please make scrobbling possible on **h--ps://cosine.club/**, ideally with the option that a track is only saved after it has been played for at least **x seconds**?  

Thank you!  
Best regards

hyphmongo:
cosine.club connector has been added 🥰

PR #5121, opened 2025-01-09
Add cosine.club connector (#5094)

hyphmongo:
Adds support for [cosine.club](https://cosine.club)

The site uses HTMX and the connector stops working on page navigation. Have added an interval to call `onStateChanged` to re-initialize in this case.


inverse (approved):
Thanks for the contribution!

PR #5084, opened 2024-12-19
Fix NTS Radio live scrobbling

HarrisonWilde:
**Describe the changes you made**
Fixes #5038 as NTS changed some class names on the tracklist window.

Issue #5038, opened 2024-11-20
nts.live no longer scrobbling


jonpsp:
**Describe the bug**
When I have the nts.live tracklist open, songs are no longer scrobbled. Instead, I see the red "WS" extension icon in the extension bar.  This happens in Brave and Vivaldi.  I tried Spotify in the middle of the session to confirm that scrobbling was possible - this is why there's a Rheingold song in the logs.

**How to reproduce**
1. Go to https://www.nts.live/live-tracklist/the-tube (probably requires an account)

**Expected behavior**
The extension icon should change to the magnifying glass or tick icon as applicable, and the song should be scrobbled

**Debug logs**
```
main.js:1 Web Scrobbler: storage.LastFM = {
  "sessionID": "********************************",
  "sessionName": "**********"
}
main.js:1 Web Scrobbler: storage.LibreFM = null
main.js:1 Web Scrobbler: storage.ListenBrainz = {
  "sessionID": "************************************",
  "sessionName": "**********"
}
main.js:1 Web Scrobbler: storage.Maloja = null
main.js:1 Web Scrobbler: storage.Webhook = null
main.js:1 Web Scrobbler: storage.Pleroma = null
main.js:1 Web Scrobbler: storage.Options = {
  "albumGuessingAllTracks": false,
  "albumGuessingDisabled": false,
  "albumGuessingUneditedOnly": true,
  "autoToggleLove": true,
  "debugLoggingEnabled": true,
  "disabledConnectors": "[hidden]",
  "forceRecognize": false,
  "scrobbleEditedTracksOnly": false,
  "scrobblePercent": 10,
  "scrobblePodcasts": true,
  "scrobbleRecognizedTracks": true,
  "showInfobox": true,
  "useNotifications": true,
  "useUnrecognizedSongNotifications": true
}
main.js:1 Web Scrobbler: storage.Connectors = {
  "YouTube": {
    "enableGetTrackInfoFromYtMusic": false,
    "scrobbleEntertainmentOnly": false,
    "scrobbleMusicOnly": false,
    "scrobbleMusicRecognisedOnly": false
  }
}
main.js:1 Web Scrobbler: storage.ConnectorsOverrideOptions = null
main.js:1 Error: Could not establish connection. Receiving end does not exist.
    at G (main.js:1:9173)
(anonymous) @ main.js:1
main.js:1 Error: Could not establish connection. Receiving end does not exist.
    at G (main.js:1:9173)
[...]

Issue #4518, opened 2024-02-14
Support for NTS Radio


a-laurence-uk:
NTS Radio is arguably the most prominent independent radio station in the UK, and over the past 10+ years has become a blueprint for community radio around the world.

Their output comprises two main live channels ([1](https://www.nts.live/1) and [2](https://www.nts.live/2)). Additionally, they have a whole host of live-streamed ["infinite mixtapes"](https://www.nts.live/infinite-mixtapes), playing a random selection of music based on a specific vibe or genre.

I am requesting NTS' broadcasting to be added to Web Scrobbler.

At the least it would be good to see channels 1 and 2 added.

Interestingly, there are two ways users are currently able to view live tracklists for NTS. Both require users to pay a membership fee to NTS, starting at £2.50 per month.

When logged in to the site, users can open a pop-out window with live tracklists for each channel:

![Screenshot 2024-02-14 at 23 45 58](https://github.com/web-scrobbler/web-scrobbler/assets/112822126/328034d8-0146-4810-b7a0-514daf0975cf)
![Screenshot 2024-02-14 at 23 46 09](https://github.com/web-scrobbler/web-scrobbler/assets/112822126/6f45206f-499b-448e-a3ad-19c3837d65ac)

Alternatively, this can be viewed from within NTS' Discord server:

![Screenshot 2024-02-14 at 23 46 33](https://github.com/web-scrobbler/web-scrobbler/assets/112822126/bad34f65-b18f-4d29-9fdf-4bd230dfdfc0)

I have zero coding or web development experience, so this is way out of my depth. However this would be a great addition to the extension and I hope someone is able to work on it.

Thanks

homosaur:
this is similar to #3506

Issue #3506, opened 2022-12-07
Support for NTS


markwheeler:
Would be great to see support for popular radio station NTS.
https://www.nts.live

Their archived shows use Mixcloud embeds, eg: https://www.nts.live/shows/guests/episodes/bill-nace-10th-november-2022

HarrisonWilde:
I would be really keen to see this implemented, happy to help

yayuyokitano:
#4477 adds partial support for NTS

PR #5145, opened 2025-01-22
Combined NTS connector

jbwharris:
Related to #5084, #5038, #3506 and #4518.

I discussed a bit with @HarrisonWilde about the functionality of the NTS connector. The previous version had 2 separate connectors, one for handling the Live feeds and one for handling the shows and mixtapes. I was able to get all the functionality combined into a single connector that handles multiple play pause states from multiple types of players. 

This connector is still imperfect, in that it doesn't just automatically work and does require user input to successfully scrobble using it. This connector will work for Live streams, recorded shows and mixtape playback. 

### Desktop View
image
This view works essentially the same way it did previously. If you click the playlist button, a popup will show up and scrobbling will commence whether music is playing or not, **there is no play/pause functionality, it will scrobble if the popup is present**. You need to ensure that you have the correct feed selected from the popup to scrobble to correct tracks. It is entirely possible to scrobble the wrong music if this is misconfigured. 

### Mobile/Tablet VIew (Sub 1024px wide)
image
On views smaller than 1024px wide the popup no longer appears as a new window and will appear within the standard website view. This view is more accurate, as it does behave more like a conventional connector, with play/pause monitoring, so it will stop scrobbling when paused. It still depends on the user to open the playlist panel, but that panel will show the relevant playlist (instead of needing to be selected like with the Desktop view). The playlist modal must stay open to scrobble music, so you can't browse around the site and expect the connector to keep scrobbling like you might expect with 

### Mobile View (Sub 768px wide)
image
Same as the Mobile/Tablet View, it just pops up from the bottom instead of in the middle. 

### Technical Detail
In this pull request I have deleted the nts-live and nts-shows connectors and just named them nts.

I haven't been able to test this yet, but I anticipate this will make scrobbling on iOS feasible using the Safari extension.


inverse (approved):
Thanks for the contribution!

jbwharris:
I just wrote out an explanation of how the NTS connector functions on Reddit. Hopefully it can head off future support threads on it not working.

[Scrobbling NTS with Web Scrobbler](https://www.reddit.com/r/NTSradio/comments/1jo7zzv/scrobbling_nts_with_web_scrobbler/)

PR #5181, opened 2025-02-12
Fix SiriusXM CSS selectors and add support for Spanish UI

matchilling:

Hey folks 👋, the SiriusXM connector stopped working in the latest version of the extension (`3.12.0`). I tested it on macOS (`14.5 (23F79)`) with Chrome (`133.0.6943.54 (Build oficial) (arm64)`) and Safari (`17.5 (19618.2.12.11.6)`). I did take a quick look at the CSS selectors, and it looks like the SXM UI (`7.52.0`) has changed and therfore broke the old ones.

So, this change:
- Fixes the broken CSS selectors
- Adds like/unlike CSS selectors
- And adds support for the Spanish SiriusXM player UI

### DIFF
--- a/src/connectors/siriusxm-player.ts
+++ b/src/connectors/siriusxm-player.ts
@@ -19,23 +19,17 @@ const playerBar = "[class*='playbackControls']";
 
 Connector.playerSelector = playerBar;
 
-Connector.artistTrackSelector =
-	`${playerBar} [class*='title']`;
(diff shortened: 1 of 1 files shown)


matchilling:
@alexesprit, I don't have permission to add labels to the PR. I'd say it's `minor-change` and `update`.

inverse (approved):
Thanks for the contribution! 

Ran the linter over the changes as it was failing. e.g. `npm run prettierfix`

PR #5190, opened 2025-02-20
Separate image building to separate npm package

yayuyokitano:

By separating image building to a separate folder, we can remove canvas as a requirement to build the extension, and instead just run image gen when necessary. This also should make building a fair bit faster, though I didn't test it.

The cost is that the size for cloning is a decent amount bigger, but hey still nothing compared to npm install.

This might still need some more time to cook, and definitely needs more testing.

**Additional context**


images/package-lock.json
inverse, line 1: Had to run `npm update` as `npm ci` failed

images/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "web-scrobbler-images",
+  "version": "1.0.0",
+  "description": "script for generating images for web scrobbler",
+  "main": "generate-icons.ts",
+  "type": "module",
+  "scripts": {
+    "generate": "tsx generate.ts"
```
inverse, line 8: Should we invoke this on the release pipeline and abort if the working dir is dirty? or run on master builds?

yayuyokitano: The point of this being separated out is that the release build pipeline no longer requires invoking canvas, so mozilla doesnt run into problems reproducing the build.
As a result, we should avoid invoking this in CI for now until theres a better solution

inverse (approved):
Seems reasonable aside the conflicts :+1:

rob9315:
rebased to current `master`, please check if this looks fine @inverse

[diff of everything except generated images](https://github.com/web-scrobbler/web-scrobbler/pull/5190/files/2b33f9461f95ba1a0fb2036cb01a83b9ff030359) in case that helps

PR #5214, opened 2025-03-08
Fix album artist special case in TIDAL.

jaccarmac:
The "canonical" link is gone; The same data is stored in a `apple-itunes-app` meta tag.

The container for artist details is slightly different. See #5213.

The information about an album page is no longer a full URL, so we match on the
last two segments, which should be a literal "album" and then its ID.


jaccarmac:
Fascinating bug with this. It works in Chrome, but Firefox seems to get the `head` update late such that it reflects the last page visited and not the current one.

PR #5232, opened 2025-03-19
fix: radio.de connector

Legiarrd:

This PR adjusts the query selectors, as it seems that radio.de has updated its website. It now uses the `data-testid` attribute since they are using Tailwind, making it difficult to retrieve elements without proper IDs or class names. I believe `data-testid` is more reliable.

**Additional context**

image


Legiarrd:
Any chance this can be reviewed soon?

@jaccarmac (just mentioned you as I saw you merged some PRs in the recent time)

PR #5293, opened 2025-04-21
Add explicit dimensions to SVGs.

jaccarmac:
Unblocks #5153 with the help of Automattic/node-canvas#2486.

node-canvas updated its use of librsvg in a way that removes support for SVGs which have `viewBox` but no `width` or `height`.


jaccarmac:
Initial testing: I can build the extension with updated canvas, but the icons are all strange-sized.

jaccarmac:
The scaling appears to be broken because of a bug in the published version of the package. It's fixed in Automattic/node-canvas#2500, so we should wait for that to land before merging #5153. However, this commit should be fine pending my tests on it.

inverse:
@jaccarmac where you seeing weird icons? It looks fine for me.

PR #5307, opened 2025-05-01
Update VK Music connector

Huterenok:
In VK, most of the music is presented in the form of playlists from users, so it is necessary to provide more information on the album title. Some tracks may have the same title but contain different versions of the tracks (studio/live) and context with the album title should help.

**Additional context**
None


ZeroJine:
It became even worse: now Web Scrobbler identifies album titles as user's playlist names, which are in most cases have nothing to do with real album titles.
shot1
shot2

Is there any opportunity to rollback to previous version or fix it?

ichigopaai:
Please fix it (as ZeroJine wrote). Now I have to change tags manually for each track.

PR #5292, opened 2025-04-19
Update Yandex Music connector

mukhinid:
**Describe the changes you made**
Scrobbler stopped working after Yandex Music redesign, cause they removed js API. I've rewrote the connector using selectors.

**Additional context**
Closes #5254 #5244 #5233 #5194 #5134 #5095 #5025 #4979

PR #5342, opened 2025-05-18
more correct detection of the old design Yandex Music

deevroman:

https://github.com/web-scrobbler/web-scrobbler/pull/5292 added support for the new Yandex Music design, but the implementation of the detection of the old design does not work correctly. WS can be loaded before initializing `window.externalAPI`. So I added an additional check for the presence of a button that suggests switching to a new design:

**Additional context**

@mukhinid since you have implemented a new connector, maybe you can add comments


deevroman:
It seems Yandex has disabled the old design for everyone, so this is no longer relevant

PR #5416, opened 2025-06-22
Add connector for LMS servers

awsms:
Added support for LMS-based servers.

**Additional context**
LMS (Lightweight Music Server) is a Subsonic server for self-hosting a music server. https://github.com/epoupon/lms

### DIFF
--- a/src/connectors/lms.ts
+++ b/src/connectors/lms.ts
@@ -1,0 +1,16 @@
+export {};
+
+Connector.playerSelector = '.container > .d-flex.align-items-center';
+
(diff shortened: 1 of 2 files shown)


src/connectors/lms.ts
```diff
@@ -0,0 +1,16 @@
+export {};
+
+Connector.playerSelector = '.container > .d-flex.align-items-center';
```
inverse, line 3: These seem quite brittle to design changes, wonder if some class selectors could eventually be added to make this easier to target.

inverse (approved):
Thanks for the contribution!

PR #5445, opened 2025-07-07
Add support for gbs-fm

hthief:

Add small connector for gbs-fm, with only 4 selectors, player, artist, track and current time.


inverse (commented):
Thanks for the contribution!

Issue #5341, opened 2025-05-18
YT Music connector sometimes initially reports `"album": null` when opening a song, parses title for artist instead


Lamby777:
**Description**

When I played the song "Weird Fishes / Arpeggi" by Radiohead through the queue last night, it showed up as the nonexistent song "Arpeggi" by Weird Fishes on last.fm. The scrobbler sometimes sees `album` is `null` for whatever reason and decides to filter the title for artist info instead, which leads to completely wrong artist/title info. So I guess the real problem here might actually be whatever is causing `album` to be `null`, but I don't know.

**Reproduce**

Prep:

- New browser profile
- Add the extension with default settings
- Disable all connectors except YouTube Music
- Debug mode enabled

Found out that just refreshing the page on the song itself doesn't work (or maybe I just got unlucky), but the bug seems to happen when going to it from a different song. Go to [a different song 😄](https://music.youtube.com/watch?v=lYBUbBu4W08) and open the console. Refresh the page, and start typing "Weird Fishes / Arpeggi" in the top, and click on the suggestion thing when it shows up.

Assuming the bug happened, it will show in debug output that the very first request made has the wrong info, but all the other ones are correct. I was trying to fix this myself, until I found out it only happens sometimes. You guys probably know better about what's going wrong here, sorry. :/

[Debug Log](https://pastebin.com/tVc7seYx)

**Environment**
-   OS: Arch Linux
-   Browser: Brave
-   Extension version: v3.14.0

1shira:
You're saying this happens initially/on the first request, what I'm seeing is that is fixes itself after that, is that also what you are observing?
If so this is happening because of a bad design choice I made out of laziness last year and should be fixed with #5260.

Lamby777:
> You're saying this happens initially/on the first request, what I'm seeing is that is fixes itself after that, is that also what you are observing? If so this is happening because of a bad design choice I made out of laziness last year and should be fixed with [#5260](https://github.com/web-scrobbler/web-scrobbler/pull/5260).

Dang `setInterval(xxxxxx, 1000);`, the culprit behind so many "we'll fix it later" bugs. 😭

Thanks for your work!

PR #5446, opened 2025-07-07
fix YTMusic Album detection v2

### LINKED ISSUE #5421: App no longer recognizes album from YTM
On YTM website, app stopped tracking album. Now it guesses what album the song might be from based on lastfm, but gets it wrong about 50% of the time and tracks nothing the other 25%

### LINKED ISSUE #5511: Wrong Albums YT Music
**Describe the bug**
Often, the scrobbler will not recognize the correct album on YouTube Music. It will put wrong info instead in the album field. Things get worse if there are hyphens or other punctuation in the album and even the song name.

**How to reproduce**
1: Listen to a song of your choosing
2: Let the song complete and move to the next song

**Expected behavior**
After manually selecting a song in an album and scrobbling it correctly, it will then mess up the album or even the song name after moving onto the next song. 

-   OS: Linux Mint
-   Browser: FireFox
-   Extension version: (e.g. v2.10.0)

**Additional context**
I don't seem to have this issue on my Windows PC using Edge. I was wondering if this is a FireFox thing.

1shira:
moved from obersving html objects to observing the mediaSession directly using a Proxy object

**Additional context**
I've talked with mantainers about this a bit, I still don't think this is an ideal solution, as we're messing with a global object here, but this _will_ detect all changes and not hit any race conditions.

closes #5421, closes #5511


Geroyuni:
I've manually built the extension based on the PR, and some songs don't end up with the wrong album when they would before, although it seems like a majority still ends up with the wrong album.

1shira:
I'm really not sure how to handle YoutubeMusics wirdness with the navigator api, it just changes things entirely inconsistently and for some reason just writes wrong data to it sometimes. 

The implementation that "worked" was just sending the data every second, which while working is not an ideal implementation, not only but also because there may be significant delay in the extention recognizing track changes. (#5341)

A solution I could think of is overwriting the prototype of `navigator.mediaSession` to observe the object directly, but I'm hesitant to mess with api objects. Would that be a prefferable solution? (@ maintiners)

1shira:
edited the original PR message to reflect current changes

Geroyuni:
The addon just says 'Play some music to get started' at all times when building firefox off this PR.

1shira:
I did manage to reproduce what you reported on my pr, but, uh, no idea why.
The data gets send and recieved correctly, but the extention doesn't seem to react.
note that this only happened some of the time for me and fixed itself upon reloading the page.
maybe someone who knows anything about the internals of the extention has an idea? 
I can't think of anything in the connector that could cause it, but then again it hasn't happened in the previous version afaik so it kinda has to be something in there

PR #5456, opened 2025-07-11
Add support for DAB Music Player 

Sheikhlipu123:
This PR adds scrobbling support for [DAB Music Player](https://dab.yeet.su/), available at both dab.yeet.su and dabmusic.xyz.
image


src/connectors/dabmusic.ts
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+Connector.playerSelector = 'div.fixed.bottom-0.left-0.right-0';
+
+Connector.getTrack = () => {
+	const trackEl = document.querySelector('div.fixed.bottom-0.left-0.right-0 h3.truncate');
```
inverse, line 6: These seem quite brittle to any UI changes. Wonder if we could target them in a nicer way?

inverse (approved):
While the selector targeting looks quite brittle to change, lets get this in and  see how we can improve this in the future.

inverse:
@Sheikhlipu123 sorry I mistakenly broke your PR by renaming the filter const. I've PR'd (#5536) a change but I saw this PR was created on your master branch so might cause problems for you to fetch upstream changes.

Issue #5453, opened 2025-07-09
NTS Connector doesn't work on Chrome


michellehwang:
**Describe the bug**

**How to reproduce**
Go to NTS.live and listen. Webscrobbler shows that nothing is playing 

**Expected behavior**
Expect it to scrobble

**Screenshots**

<img width="1721" height="471" alt="Image" src="https://github.com/user-attachments/assets/716a4b90-7206-42af-b1f2-9e179362a3b4" />

**Environment (please complete the following information)**
-   OS: Mac
-   Browser: Chrome
-   Extension version: (e.g. v2.10.0)

**Additional context**

jbwharris:
@michellehwang That extension version is really old, can you check you're running the latest? 3.13.0 from February was when the latest iteration of the NTS connector was released. Latest is 3.14.2.

michellehwang:
Oops I am on 3.14.2! Sorry I just didn't fill it in there.

rob9315:
seems like this was addressed in a PR

Issue #5450, opened 2025-07-08
NTS Connector doesn't seem to work live on Safari


HarrisonWilde:
**Describe the bug**

The NTS connector used to work in the live tracklist pop-up, now it doesn't seem to on macOS Safari.

FI @jbwharris do you know why this might be?

**How to reproduce**

1. Go to https://www.nts.live
2. Open the live tracklist pop-up

**Expected behavior**

Scrobbling

**Screenshots**

<img width="450" height="512" alt="Image" src="https://github.com/user-attachments/assets/1bb084eb-2057-44da-ba7e-4d3937b5136a" />

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   OS: Mac
-   Browser: Safari
-   Extension version: v3.9.0

1shira:
How recently did this work?
The nts connector was updated with 3.12 (Dec 25, 2024) and 3.13 (Feb 19, 2025)

can you try on the newest version?

HarrisonWilde:
@1shira I don't remember exactly, it was a few months ago, I do remember the HTML structure changing at some point. There doesn't seem to be a new version available to download on Mac for Safari Extensions, perhaps this is the source of the issue then, could be that a new build needs to be published in that channel?

1shira:
The Safari extention store currently has v3.9.0, the most recent one is v3.14.2 so yeah that could be the issue, that is why I asked.

I am not sure how it is possible to load unsigned/unpackaged extentions in safari and not able to test myself, but given #5453
(not sure if this is the same issue) it could also have broken again. Could you verify if it works in chrome for you?

jbwharris:
> [@1shira](https://github.com/1shira) I don't remember exactly, it was a few months ago, I do remember the HTML structure changing at some point. There doesn't seem to be a new version available to download on Mac for Safari Extensions, perhaps this is the source of the issue then, could be that a new build needs to be published in that channel?

The iOS Safari extension seems to be pretty out of date. An alternate approach on iOS is using Orion and installing the Chrome extension. See if that works using the latest.

HarrisonWilde:
Ok so weirdly on Chrome v3.14.2 the live scrobbling works but the recorded shows doesn't seem to (the opposite of the old version on Safari)... Is there a reason the safari one cant be published on the app store as its updated?

[...]

PR #5494, opened 2025-07-28
Fix NTS connector

jbwharris:
Related to #5453 #5450 #4518 #3506

I did a significant rethink of this connector compared to what it was previously. In the most recent version it required staying on certain pages, or keeping popups open to keep logging. This is no longer necessary. It now just functions in a more automatic fashion based on whichever player you're currently using. There's 3 different types of players on this site, the live radio steams, the show archives and the infinite mixtapes.  

One thing of note is that I've appended the now playing Web Scrobbler data to the header of the page, so you can always see what's actually happening. This station tends to hide the song and artist from the UI, which has been the biggest struggle with it. How I resolved this was instead of relying on Web Scrobbler to find the selector, which would often be hidden, is to pull the song and artist data into into a variable, then pass that variable. 

**Additional context**
image
Playing Mixtapes in desktop view, with the popup. 

image
Smaller screen showing the mixtape data.

image
Live radio stream with scrobbling data.

image
Archived show with scrobbling data

I still need to test this on iOS, hoping to figure out maybe testing using Orion and potentially being able to load an unsigned extension, but will have to see.


inverse:
@jbwharris sorry for the delay getting back to this, looks like there are conflicts. Is this still valid?

jbwharris:
> @jbwharris sorry for the delay getting back to this, looks like there are conflicts. Is this still valid?

@inverse I don't know how to resolve this. When I look at the conflict, I would resolve it, but the button is greyed out for me. But should be fine to merge.

jaccarmac:
Merged it. Can you give that a test? I'm confused myself about where the conflict came from.

jaccarmac:
I suspect you may have branched off your own version of the unified connector before it landed in this repository.

jbwharris:
> I suspect you may have branched off your own version of the unified connector before it landed in this repository.

@jaccarmac I just pushed a new version with slight CSS tweaks to the now playing output. It seems it doesn't like the `box-orient` value, but it is valid CSS, just probably not in the valid CSS rules list yet. `line-clamp` doesn't work without it, which I feel is necessary for not breaking the site layout with long song names.

Issue #5459, opened 2025-07-12
The Current connector not working


elskootro:
**Describe the bug**

**How to reproduce**

**Expected behavior**

**Screenshots**

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   OS: (e.g. Windows, Mac)
-   Browser: (e.g. Chrome, Safari)
-   Extension version: (e.g. v2.10.0)

**Additional context**

1shira:
Which site????

elskootro:
https://www.thecurrent.org/

1shira:
ah, suddenly the title provides much more information lol

jbwharris:
@elskootro Not a fix, but [The Current](https://scrobblerad.io/#thecurrent) scrobbles fine on my app ScrobbleRadio.

elskootro:
> [@elskootro](https://github.com/elskootro) Not a fix, but [The Current](https://scrobblerad.io/#thecurrent) scrobbles fine on my app ScrobbleRadio.

Oh nice! Thanks. Bookmarked the link. Cheers!

rob9315:
@inverse can be closed, works fine currently

<details>
  <summary>screenshot to prove</summary>
  <img width="1734" height="1012" alt="Image" src="https://github.com/user-attachments/assets/b9fc652f-6f90-4279-9d70-77a80b9ff70f" />
</details>

jbwharris:
Can confirm as well, seems to be working

PR #5498, opened 2025-07-30
The Current connector fix

jbwharris:
The Current seemingly changed some site structure slightly, so I updated them to the new selectors.

#5459


inverse (approved):
Thanks for the contribution!

PR #5522, opened 2025-08-15
Fix self-hosted HTTP-only scrobbling services in Firefox

martin0300:

Disabled 'upgrade-insecure-requests' CSP policy to allow connecting to self-hosted HTTP-only Maloja or other scrobbling servers in Firefox.

**Additional context**

Firefox's default CSP policy automatically upgrades HTTP connections to HTTPS, which prevents self-hosted HTTP-only servers from working.


inverse (approved):
Thanks for the contribution!

PR #5535, opened 2025-08-16
Add connector for sogreatandpowerful.com

VINXIS:
Created a connector for https://sogreatandpowerful.com/

**Additional context**
Fan site of an artist named sgap that was really easy to make a connector for.
Their music is mainly only accessible here online. The site itself hasn't really changed for years, so maintenance shouldn't be an issue

### DIFF
--- a/src/connectors/sogreatandpowerful.ts
+++ b/src/connectors/sogreatandpowerful.ts
@@ -1,0 +1,5 @@
+export {};
+
+Connector.artistSelector = 'h1';
+Connector.trackSelector = '#info';
+Connector.playerSelector = '#player';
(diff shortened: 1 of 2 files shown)


src/connectors/sogreatandpowerful.ts
```diff
@@ -0,0 +1,5 @@
+export {};
+
+Connector.artistSelector = 'h1';
+Connector.trackSelector = '#info';
+Connector.playerSelector = '#player';
```
inverse, line 5: This could be replaced with:

```js
Util.bindListeners(
	['audio'],
	['playing', 'pause', 'timeupdate'],
	Connector.onStateChanged,
);
```

Since we can bind to the `audio` element to watch for events from that.

inverse (changes requested):
Looks good, but suggested a few additional changes.

VINXIS:
Thanks for the quick look! Added the suggestions

PR #5550, opened 2025-08-22
fix: Resolved the issue where album covers could not be scraped from QQ Music

monyuan:

**Additional context**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,10 +4,30 @@ on:
     branches: [master]
   pull_request:
     types: [opened, synchronize]
+  # 添加手动触发选项
+  workflow_dispatch:
+    inputs:
+      build-browser:
+        description: '选择要构建的浏览器 (留空则构建所有)'
+        required: false
+        default: 'all'
+        type: choice
+        options:
+          - all
+          - chrome
+          - firefox
+          - safari
+
 jobs:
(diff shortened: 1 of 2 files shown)


src/connectors/qq-music.ts
```diff
@@ -5,7 +5,7 @@ Connector.playerSelector = '.player__ft';
 Connector.getTrackArt = () => {
 	const src = Util.getAttrFromSelectors('.song_info__pic', 'src');
 	const trackArt = src?.split('?')[0];
-	return trackArt;
+	return 'https:' + trackArt;
```
inverse, line 8: Do we really need to remove everything after the `?` for the image to load?  If not would replacing the whole `Connector.getTrackArt` functiion with `Connector.trackArtSelector = '.song_info__pic';` work?

By doing this we call some internal logic that extracts the image and normalizes it with the protocol.

.github/workflows/build.yml
inverse, line 1: These changes should not be in this pull request

inverse (changes requested):
Looks good but we def need to remove that workflow change.

src/connectors/qq-music.ts
rob9315, line 8: Please try this instead
```suggestion
    return Util.normalizeUrl(src);
```

Issue #5582, opened 2025-09-08
mirlo connector mis-identifies artist compilation tracks as track one artist


ajft:
**Describe the bug**
When playing all tracks from a mirlo compilation album, eg https://mirlo.space/team/release/double-time, the first track is identifed correctly by track name and artist, all subsequent tracks display with their track name, but the artist from track one

**How to reproduce**
1. Go to 'https://mirlo.space/team/release/double-time'
3. Play album
4. observe auto-scrobbled track and artist values

**Expected behavior**
Each track to have its own artist identified

PR #5589, opened 2025-09-11
Fix Mirlo scrobbling the first track of certain releases repeatedly.

jbwharris:
#5582 

Increased specificity of artist and track selectors. It appears Mirlo just had some repeating IDs in the code, so the artist and song were being picked up by the first instance of the ID. Since there is an assumption in valid HTML that there's only one ID per page, and that wasn't true, Web Scrobbler was finding the first instance and stopped looking on the page.

### DIFF
--- a/src/connectors/mirlo.ts
+++ b/src/connectors/mirlo.ts
@@ -1,9 +1,9 @@
 export {};
 
 Connector.playerSelector = '#player';
-Connector.trackSelector = '#player-track-title';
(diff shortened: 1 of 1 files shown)


jaccarmac:
Thanks for the fix. The pattern - using `playerSelector` as a context - is common enough that there's a convention of reusing that variable. Doesn't come close to saving typing here, but I think it's clear enough semantically.

PR #5609, opened 2025-09-18
Add support for Aonsoku

neptunethefox:
Added a connector for [Aonsoku](https://github.com/victoralvesf/aonsoku).

**Additional context**
Example screenshot:
image


jaccarmac:
8080 is a common enough port that I'd prefer not to bind there by default. Do we have a pattern for documenting that adding that pattern could be useful?

PR #5612, opened 2025-09-19
Add support for Feishin

MitenkovIvan:

- Made a new file `connectors/feishin.ts` to support [Feishin](https://github.com/jeffvli/feishin)
- Changed `core/connectors.ts` to include Feishin connector details

**Additional context**
Example screenshot:
webscrobbler-feishin-test

### DIFF
--- a/src/connectors/feishin.ts
+++ b/src/connectors/feishin.ts
@@ -1,0 +1,17 @@
+export {};
+
+Connector.playerSelector = '#player-bar';
+
+Connector.trackSelector = '._line-item_ly86j_53:nth-child(1) > div > a';
+
+Connector.artistSelector = '._line-item_ly86j_53:nth-child(2) > a';
+
+Connector.albumSelector = '._line-item_ly86j_53:nth-child(3) > a';
+
+Connector.currentTimeSelector =
+	'._slider-value-wrapper_jw89i_13:nth-child(1) > div';
+
+Connector.durationSelector =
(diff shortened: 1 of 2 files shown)


inverse:
Thanks for providing support for this! 

Won't this be problematic when changes are made and the selectors get updated as part of that process?

I noticed a similar behaviour with https://archaichorizon.com/ and wrote asking if they would be open to adding semantic elements for us to bind too which the creator did in https://github.com/jscottsmith/archaichorizon.com/pull/1

Wonder if something could be set on this product too.

## Update

Quickly good Google Jules to draft something and it worked locally when changing things too

https://github.com/jeffvli/feishin/pull/1138

```js
Connector.playerSelector = '.media-player';

Connector.trackSelector = '.song-title';

Connector.artistSelector = '.song-artist';

Connector.albumSelector = '.song-album';

Connector.currentTimeSelector = '.elapsed-time';

Connector.trackArtSelector = '.player-cover-art';

Connector.durationSelector = '.total-duration';

Connector.isPlaying = () => {
	return Util.isElementVisible('.player-state-playing');
};
```

What do you think of holding off until we get their feedback and then apply things semantically?

MitenkovIvan:
> Won't this be problematic when changes are made and the selectors get updated as part of that process?

> What do you think of holding off until we get their feedback and then apply things semantically?

Yeah, I agree that the selectors could be better, so I'm definitely looking forward to their feedback on semantic elements.

Issue #5098, opened 2024-12-28
Connector for muztv site


Octasom:
Hi, please consider adding a connector to this site: [https://muztv.org/](https://muztv.org/)
It is free and doesn't require registration.

Thank you in advance!

rob9315:
site has been dead for a while
https://web.archive.org/web/20250320083858/https://muztv.org/

PR #5698, opened 2025-11-01
chore: tidy project locales

inverse:

Part of migrating to a new translation tool (#5098) it feels like the locales should be tidied up:

- Remove untranslated as per Transifex UI
  - en_BR
  - en_US - default one and not needed 
  - id_ID 
  - pt_PT
  - sl_SI
- Renamed for fallback resolution
  - pt_BR -> pt - to cover all Portuguese 
  - sk_SK -> sk - to cover all Slovak 
  -  tr_TR -> tr - to cover all Turkish
  - uk_UA -> uk - to cover all Ukrainian
  - zh -> zh_CN - we have zh_TW and zh_CN makes more sense here

**Additional context**


lucasmz-dev:
pt_BR should continue to be pt_BR. Brazilian Portuguese is very different from the colonizers'.

The rest of the changes seem fine, especially the empty translations

inverse:
@lucasmz-dev thanks for the review - was thinking about that and the other ones as well as they were originally defined with the country. 

I'll revert back the PT one

PR #5706, opened 2025-11-05
chore: use standard multi-artist join for ampwall

inverse:

Other connectors use the `Util.joinArtists` function on Nodes to produce the same `, ` joined string when multiple artists identified.  

@Harjun751 what do you think?

**Additional context**


Harjun751:
Oop, wasn't aware. LGTM!

PR #5727, opened 2025-11-13
Suno Connector: Update selectors for new Playbar markup (Nov 2025)

### LINKED ISSUE #4886: suno.com
A short description of the website, sample URLs, (demo) accounts for testing purposes. If (demo) accounts are not for public use, you can mail them to the project developers.

As you probably know this is a AI music creation and play site where ppopel can  stream musci and even creat playlist and wi; as play songs from accounts pages

my account is (for example) is: https://suno.com/@chandler_songwriting

**Additional context**
Add any other context or screenshots about the request here.

The main address is https://suno.com

TaQuangKhoi:
- Broaden `playerSelector` to match both Play and Pause button variants so the connector observes state in either UI state.
- Replace the old combined time element parsing with direct selectors for the new left/right time labels:
  - Current time: `div.min-w-8.text-right.text-xs`
  - Duration: `div.min-w-8.text-left.text-xs`
- Kept existing `aria-label`-based selectors for artist, title, and cover image since they still match the new markup.
- Left `playButtonSelector` as the Play button; core uses its visibility to infer playing state, which remains correct with a separate Pause button present.

**Additional context**
- Based on the new Playbar HTML shared in the issue/description.
- Local build verification on Windows: `npm run build chrome` completed successfully.
- File touched: `src/connectors/suno.ts`.

If acceptable, please also update the PR title to the one proposed above. If you prefer, I can update the title/body directly given appropriate permissions.

Fixes #4886

### DIFF
--- a/src/connectors/suno.ts
+++ b/src/connectors/suno.ts
@@ -1,6 +1,7 @@
 export {};
 
-const durationSelector = 'span[aria-label^="Playbar: Duration for"]';
+const currentTimeSelector = 'div.min-w-8.text-right.text-xs';
+const durationTimeSelector = 'div.min-w-8.text-left.text-xs';
 
 enum TimePart {
 	Current = 'current',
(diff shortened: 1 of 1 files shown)


src/connectors/suno.ts
```diff
@@ -1,14 +1,15 @@
 export {};
 
-const durationSelector = 'span[aria-label^="Playbar: Duration for"]';
+const currentTimeSelector = 'div.min-w-8.text-right.text-xs';
```
inverse: These seem quite brittle to change as they are not semantic. Wonder what improvements could be made here..

inverse (approved):
While change seems quite brittle, we can adapt in in the future

Issue #5347, opened 2025-05-23
Bandcamp Playlists!


JoshMock:
**Provide information about website**

[Bandcamp playlists](https://bandcamp.com/about_playlists) are now available as a public beta.

An example playlist: <https://bandcamp.com/jzcamp/playlist/doomgaze>

**Additional context**

The only limitation I'm aware of is that you have to be logged in to Bandcamp.com to listen to a playlist.

rob9315:
did not know they had that, sounds cool and looks doable as well

Issue #5499, opened 2025-07-31
New Bandcamp player not scrobbling


JuiceBard:
**Describe the bug**
New bandcamp player used on the Bandcamp homepage and for Bandcamp Radio doesn't scrobble. (Bandcamp Radio did previously scrobble tracks.)

**How to reproduce**
1. Go to [Bandcamp homepage](https://bandcamp.com/) or [Bandcamp radio](https://bandcamp.com/radio)
2. Play something from the [New and notable section](https://bandcamp.com/#v-2-0) or a Bandcamp radio show
3. Doesn't scrobble. 

**Expected behavior**
Expected to scrobble tracks from the homepage and individual tracks from the bandcamp radio shows

**Screenshots**
<img width="1786" height="769" alt="Image" src="https://github.com/user-attachments/assets/7fac2d8c-67eb-45dc-968d-4cb98a70b4a0" />

**Debug logs**
<details><summary>Debug log</summary>

main.js:1 Web Scrobbler: storage.LastFM = {
  "sessionID": "********************************",
  "sessionName": "*********"
}
main.js:1 Web Scrobbler: storage.LibreFM = null
main.js:1 Web Scrobbler: storage.ListenBrainz = null
main.js:1 Web Scrobbler: storage.Maloja = null
main.js:1 Web Scrobbler: storage.Webhook = null
main.js:1 Web Scrobbler: storage.Pleroma = null
main.js:1 Web Scrobbler: storage.Options = {
  "albumGuessingAllTracks": false,
  "albumGuessingDisabled": false,
  "albumGuessingUneditedOnly": true,
  "autoToggleLove": true,
  "debugLoggingEnabled": true,
  "disabledConnectors": "[hidden]",
  "forceRecognize": false,
  "scrobbleEditedTracksOnly": false,
  "scrobblePercent": 50,
  "scrobblePodcasts": false,
  "scrobbleRecognizedTracks": true,
  "showInfobox": true,
  "useNotifications": true,
  "useUnrecognizedSongNotifications": true
}
main.js:1 Web Scrobbler: storage.Connectors = {
  "YouTube": {
    "enableGetTrackInfoFromYtMusic": true,
    "scrobbleEntertainmentOnly": false,
    "scrobbleMusicOnly": true,
    "scrobbleMusicRecognisedOnly": true
  }
}
main.js:1 Web Scrobbler: storage.ConnectorsOverrideOptions = {
  "bandcamp": {},
  "youtube": {
[...]

PR #5736, opened 2025-11-22
Update the Bandcamp connector to work with the new player

adam-blackwater:

Added support for the new Bandcamp player that was introduced with the site update which broke the old connector.

Removed a call to the function getOriginUrl function as the file `src/core/content/connector.ts:627` outlines that this is an internal function that should not be called by connectors.

**Additional context**
Unfortunately if the user collapses the mini player songs are no longer scrobbled. This is due to the now playing dialogue being unloaded from the DOM when the player is collapsed.

I've tried to only use selectors where possible. I only used the get functions were I couldn't get selectors to work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1367,17 +1367,6 @@
         "node": ">=6.0.0"
       }
     },
-    "node_modules/@jridgewell/source-map": {
-      "version": "0.3.6",
-      "resolved": "https://registry.npmjs.org/@jridgewell/source-map/-/source-map-0.3.6.tgz",
-      "integrity": "sha512-1ZJTZebgqllO79ue2bm3rIGud/bOe0pP5BjSRCRxxYkEZS8STV7zN84UBbiYu7jy+eCKSnVIUgoWWE/tt+shMQ==",
-      "optional": true,
-      "peer": true,
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.5",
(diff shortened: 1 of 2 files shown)


JuiceBard:
@adam-blackwater Thanks for looking into this. Any update on getting this merged and into a release?

hornbuckle:
I believe part of the problem here is that 569 lines from package-lock.json were removed and shouldn't have been. Commits are also showing as "Unverified."

rob9315:
unverified doesn't matter, just means you didn't sign it
is this still relevant? because the connector mostly works for me

hornbuckle:
I'm not sure it's still relevant as-is. There are other issues to address in the Bandcamp connector, and this only kind of solves for one section, but also not really because the tracklist it relies on can easily be destroyed in the DOM and break the connection. It may be that we just don't support this section of the site for now.

rob9315:
oh, it tries to fix this: #5499 (edit: i'm blind, it was already reverse linked)

hannahswain:
Opened #6111 to add scrobbling for Bandcamp playlist pages (`og:type=music.playlist`). It plugs into the existing connector dispatch independently of this PR and closes #5347 — credited the new-player markup you mapped here, since that's where I picked up the approach. 👍

package-lock.json
rob9315, line 1: this shouldn't be deleted, please undo this change

PR #5742, opened 2025-12-06
Add support for WPGU

sriya-g:
Added support for scrobbling tracks played on https://wpgu.com/ 

Only gets track and artist. (Only track and artist are provided on the website)

image

### DIFF
--- a/src/connectors/wpgu.ts
+++ b/src/connectors/wpgu.ts
@@ -1,0 +1,7 @@
+export {};
+
+Connector.playerSelector = '.nowPlaying_nowPlayingText__Gszur';
+
+Connector.trackSelector = '.nowPlaying_songTitle__BRb3L';
+
+Connector.artistSelector = '.nowPlaying_songArtist__1S2_y';
--- a/src/connectors/wpgu.ts
+++ b/src/connectors/wpgu.ts
@@ -1,7 +1,14 @@
 export {};
+Connector.isPlaying = () => {
+const playPauseButton = document.querySelector(
+'[class*="livestream_playPauseButton"]',
+);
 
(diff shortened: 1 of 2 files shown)


src/connectors/wpgu.ts
```diff
@@ -0,0 +1,7 @@
+export {};
+
+Connector.playerSelector = '.nowPlaying_nowPlayingText__Gszur';
```
inverse:
```suggestion
Connector.playerSelector = 'body';
```

Wonder if this should be changed to `body` so we can watch for the play button changes

src/connectors/wpgu.ts
```diff
@@ -0,0 +1,7 @@
...
+
+Connector.trackSelector = '.nowPlaying_songTitle__BRb3L';
```
inverse:
```suggestion
Connector.trackSelector = '[class*="nowPlaying_songTitle"]';
```


These feel brittle to any new releases of the site as the last bit is computed as part of the build step from what I understand. Perhaps something more relaxed could help.

inverse (commented):
Looks good but suggested some improvements.

Also perhaps observing the play state could provide more accurate scrobbles

Connector.isPlaying = () => {
	const playPauseButton = document.querySelector(
		'[class*="livestream_playPauseButton"]',
	);

	return playPauseButton?.className.includes('livestream_playing') ?? false;
};

src/connectors/wpgu.ts
```diff
@@ -0,0 +1,7 @@
...
+
+Connector.artistSelector = '.nowPlaying_songArtist__1S2_y';
```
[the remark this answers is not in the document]
sriya-g: I added the suggestions in my new commit, thank you!

PR #5754, opened 2025-12-22
Add support for NPO Radio 2.

MHormes:
Added an entry in the connectors.ts file for NPO Radio 2.
Created npo-radio2.ts file to read currently playing song on NPO Radio 2

### DIFF
--- a/src/connectors/npo-radio2.ts
+++ b/src/connectors/npo-radio2.ts
@@ -1,41 +1,67 @@
 export {};
 // Use a Set for faster, cleaner lookups (Performance/FP)
 const DJ_SET = new Set([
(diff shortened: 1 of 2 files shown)


MHormes:
Tried to fix the failing CI tests. If anything is required before approval please let me know.

rob9315 (changes requested):
the selectors have all changed, but they now retain some general naming from their sources, in addition to that random suffix that you were matching before.

Please change them to reflect the new state of the website, you can match for a specific class without the random part with a selector like `[class*="SongWithArtist__Wrap"]` for a class like `.SongWithArtist__Wrap-sc-d96b906e-0`.

The page also has a popout player which would be nice to support for people using that instead, there you'd probably have to find the info in the recently played list

PR #5771, opened 2026-01-04
Add support for The Indie Beat Television

scy:
Add a new connector for [The Indie Beat Television](https://tv.theindiebeat.fm/).

**Additional context**
Tested and verified locally.

Since this is a livestream with slightly different latency for each visitor, it doesn't have an "elapsed time" indicator (because it wouldn't be precise enough), but Web Scrobbler can deal with that just fine.

We are however provided with the duration of the track, allowing Web Scrobbler to be smart about when to scrobble.

### DIFF
--- a/src/connectors/theindiebeattv.ts
+++ b/src/connectors/theindiebeattv.ts
@@ -1,0 +1,27 @@
+export {};
+
+Connector.playerSelector = '#player';
+
+Connector.pauseButtonSelector = '#player button.vjs-playing';
+
+Connector.artistSelector = '#nowplaying .song-artist';
+
+Connector.trackSelector = '#nowplaying .song-title';
+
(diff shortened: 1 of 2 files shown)


scy:
Rebased onto current `master` to fix conflicts and force-pushed.

rob9315:
i didn't notice this pr, i implemented support in #5977 which just got merged
@inverse i guess we can close this one then?

scy:
Thanks for implementing support. As the person who created the original PR and who implemented TIBtv's "now playing" feature, I've commented on the merged code in its own PR: https://github.com/web-scrobbler/web-scrobbler/pull/5977#issuecomment-4320748705

inverse:
Apologies for overlooking this pull request, def got lost in all the noise that needs to be checked.

Thanks for the contribution!

Issue #5740, opened 2025-11-27
Scrobbler stopped working on safari for Youtube Music


snooppy:
**Describe the bug**
Scrobbler doesn't work on safari for Youtube Music.

**How to reproduce**
Open Youtube Music in Safari.

**Expected behavior**
Music expected to be scrobbled.

**Screenshots**

<img width="1710" height="1078" alt="Image" src="https://github.com/user-attachments/assets/fedfb17f-5c09-46d4-b348-402e9853cf00" />

**Debug logs**
`[Error] XMLHttpRequest cannot load https://music.youtube.com/youtubei/v1/log_event?alt=json due to access control checks.
	Ll (music_polymer_inlined_html.js:1258:163)
	Nl (music_polymer_inlined_html.js:1266)
	fla (music_polymer_inlined_html.js:1553:251)
	ko (music_polymer_inlined_html.js:1552:358)
	Mla (music_polymer_inlined_html.js:1676:493)
	(anonymous function) (music_polymer_inlined_html.js:1576)
	(anonymous function) (music_polymer_inlined_html.js:1673:120)
	q (music_polymer_inlined_html.js:1698:198)
	(anonymous function) (music_polymer_inlined_html.js:1700:98)
[Error] Resource blocked by content blocker (music.youtube.com, line 10)
[Error] XMLHttpRequest cannot load https://googleads.g.doubleclick.net/pagead/id due to access control checks. (music.youtube.com, line 10)
[Error] Source Map loading errors (x4)
[Error] Failed to load resource: the server responded with a status of 404 () (web-animations-next-lite.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (web-animations-next-lite.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (fast-shim.js.sourcemap, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (webcomponents-sd.js.sourcemap, line 0)
[Error] Refused to load safari-web-extension://BFC2538A-B95B-4725-AABC-CF8362C5D2AD/connectors/youtube-music-dom-inject.js because it does not appear in the script-src directive of the Content Security Policy.
[Error] Failed to load resource: the server responded with a status of 401 () (log, line 0)
[Error] Failed to load resource: the server responded with a status of 401 () (log, line 0)`

**Environment (please complete the following information)**
-   OS: Mac OS Tahoe 26.1
-   Browser: Safari
[...]

Issue #5318, opened 2025-05-07
Release Safari Update


bloomkd46:
**Describe the bug**
The extension stopped working for SiriusXM a little while ago, and I notice that it’s fixed in an update on GitHub, but that update is not available in the App Store. I was able to get it working by building the extension from source and running it in Xcode, but I have to re-enable the extension every time I open Safari since it’s unsigned. Please consider pushing the update to the App Store (or even adding a workflow that does it automatically as you’ve already done for FireFox and Chrome)

**How to reproduce**
1. Go to the App Store
2. Click on the App
3. It shows that the latest version is `3.9.0` (should be `3.13.0`)

**Expected behavior**
Updates should be synced to the MacOS App Store

**Screenshots**
N/A

**Debug logs**
N/A

**Environment (please complete the following information)**
-   OS: Mac
-   Browser: Safari

**Additional context**
This issue is a modern version of #4549

inverse:
@yayuyokitano is this one you can push out? will publish a new build as there has been a while since one

bloomkd46:
Any progress?

PR #5778, opened 2026-01-07
Handle Safari CSP restriction for script injection (#5740)

snooppy:
Fixed the Safari CSP issue for the YouTube Music connector.
**The Problem:**
On Safari, injecting scripts via <script src="safari-web-extension://..."> is blocked by the page's Content Security Policy.
**The Solution:**
Modified injectScriptIntoDocument in src/core/content/util.ts to detect Safari (when the script URL starts with safari-web-extension://).
**On Safari:** fetch the script content and inject it as inline code using script.textContent instead of script.src.
**On other browsers:** keep the original method (script tag with src).

### DIFF
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -753,6 +753,23 @@ export function normalizeUrl(url: string | null | undefined): string | null {
  */
 /* istanbul ignore next */
 export function injectScriptIntoDocument(scriptUrl: string): void {
+	// On Safari, script tags with src pointing to safari-web-extension:// URLs
+	// are blocked by Content Security Policy. Fetch and inject inline instead.
+	if (scriptUrl.startsWith('safari-web-extension://')) {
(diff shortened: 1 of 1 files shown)


snooppy:
Hi @yayuyokitano ! Any chance to review?

rob9315:
Are you sure that this fixes the issue #5740 and it's not just due to the outdated published version v3.9.0 on safari due to #5318?

Edit: I read the error messages of that issue and yeah, it sure looks like that is the real issue. A newer version would (soon?) also fix this because the script injection is going to be removed from the youtube-music connector.
[Apple recommends](https://developer.apple.com/documentation/safariservices/injecting-a-script-into-a-webpage#Add-a-script) adding scripts that should be injectable to the `Info.plist` file, is that something we should consider to do instead of creating new elements on the page?

Issue #5759, opened 2025-12-28
YouTube Music connector not using mediaSession album name


jdgarcia:
**Describe the bug**
A did a quick look at the code and it seems the YT Music connector uses the the `navigator.mediaSession` object for track data. However, in practice it seems to frequently use a completely different album name.

**How to reproduce**
It's inconsistent, some songs have the correct album name and some don't.

**Expected behavior**
I expect the detected album name to match the mediaSession data.

**Screenshots**

<img width="304" height="151" alt="Image" src="https://github.com/user-attachments/assets/e35f1fd9-3ad6-4c19-aea4-16a96929a7b0" />
<img width="359" height="175" alt="Image" src="https://github.com/user-attachments/assets/09471866-3e99-4849-bcfa-0f29ae8c7df4" />

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   OS: Mac
-   Browser: Arc (Chrome)
-   Extension version: 3.18.0

jdgarcia:
I dug around a bit and I discovered this is happening because the `mediaSession.metadata` object has empty string for the album info at the moment the MutationObserver callback happens. The album value becomes available slightly after the callback occurs. The connector ends up with no album name in its state so the extension ends up having to guess the album name.

Here's some logs I added to the dom inject script to show the issue:

<img width="661" height="443" alt="Image" src="https://github.com/user-attachments/assets/00c76c99-a987-4106-b5a8-940ad6014408" />

<img width="911" height="220" alt="Image" src="https://github.com/user-attachments/assets/ecf07fc2-865e-49a5-b19a-ffdd23822688" />

ZariTen:
Yeah, it's been happening since August, I think. This doesn't seem to happen to the first song you play; only the next ones in the playlist seem to mess up the album name. Check https://github.com/web-scrobbler/web-scrobbler/issues/5511

UlissesBardini:
One solution I found was to pause and resume the song, it's not perfect of course, but it "solves" the problem. Also, the album it recognizes depends on the album tracklists in Last.fm, because some (most) albums have messed up tracklists, songs that should be in an album end up not being related to it, and, as the album is empty, if you have the "Guess album" option selected, it will get the album info directly from Last.fm, causing this bug.

PR #5785, opened 2026-01-16
Fix Youtube Music connector album info

jdgarcia:

The YTM connector was using a MutationObserver to detect changes to the player. However, it seems YTM does not have the `album` value in the `mediaSession` object available at the moment the mutation observer handler is called. It album name seems to get added a short time afterward (about a second). I replaced the mutation observer with a setInterval, which seems to fix the issue. I could have done a combination of the mutation observer combined with a setTimeout, but this seemed like a simpler solution.

**Additional context**
This addresses #5511 and #5759. See those items for more info.

### DIFF
--- a/src/connectors/youtube-music-dom-inject.ts
+++ b/src/connectors/youtube-music-dom-inject.ts
@@ -17,7 +17,7 @@ if ('cleanup' in window && typeof window.cleanup === 'function') {
 }
 
 (window as unknown as { cleanup: () => void }).cleanup = (() => {
-	const sendData = () => {
+	let pollInterval: number | null = window.setInterval(() => {
 		window.postMessage(
 			{
 				sender: 'web-scrobbler',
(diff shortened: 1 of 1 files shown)


1shira:
This basically reverses #5260, which is fair cause that does solve the album issue, it'll probably reintroduce other issues,  #5446 exists but maintainers don't seem to be looking at it, honestly I don't know what would be better here.

jdgarcia:
@1shira Perhaps some hybrid solution? Keep the mutation observer but instead of immediately sending data, it sets up a setInterval that waits until `album` is available, then sends the data and clears the interval.

Kerkerseele:
I've been manual scrobbling for so long because of this now. It's unpredictable as well because it would scrobble a track correctly, then the next time I listen to the same song it'll either be empty or use a different album.

jdgarcia:
@1shira Any idea what issues this reintroduces? I could try to see if can make sure those stay fixed too. For what it's worth, I've been running this locally for the last few weeks (since the bug is a bit of a deal breaker for me) and so far haven't run into any issues.

1shira:
likely mostly smaller issues like #5341, in my eyes preferable to the current situation too, spamming events is just generally not the way in code, which is why I've been trying to get the current solution working. I can test my own fix more and see if I find any new issues but I haven't in previous tests, we just need maintainers to decide on what to merge I guess.
The whole thing just seems like youtube race conditions to me

Geroyuni:
I've been able to use the extension with jdgarcia's changes for the last few days, but I'll be trying out 1shira's newest changes too.

IMO although it is very understandable to ensure there are no edge cases and maintain good code, the implementation for this streaming service has been entirely unusable for close to half a year. I'd rather have been able to use it at all.

jdgarcia:
Decided to check back on this and see if there's been any update on this.

> spamming events is just generally not the way in code

That's very fair, so I've updated the code a bit so it only calls `postMessage` if anything has actually changed. I've been running my patched version this whole time and I haven't noticed any issues with it, for what it's worth 🤷

PR #5818, opened 2026-02-09
Fix Niconico connector

### LINKED ISSUE #5036: Please make Niconico videos available for Scrobbling again.
https://www.nicovideo.jp/

I was able to scrobble on Nico Nico Douga before, but since the Nico Nico Douga update, scrobbing is no longer possible. Please do something about it.

jedenastka:

Fix ニコニコ/Niconico (https://nicovideo.jp) connector, as it is currently broken.

Selectors aren't the best, but the UI changed a lot, and it no longer has nice class names for each element.

Fixes #5036.

**Additional context**

None.


package-lock.json
inverse: Can we remove this from the PR

inverse (changes requested):
Looks good aside the additional file.

PR #5896, opened 2026-03-13
Fix bbc-sounds connector

### LINKED ISSUE #5730: BBC Sounds - website changed so doesn't work online.
BBC sounds (browser version) has changed layout. Scrobbling doesn't work on there any more.

https://www.bbc.co.uk/sounds/ the webscrobbler say it is supported and has access  - but nothing actually comes through. 

Can it be updated to accommodate for the change on the website?

### LINKED ISSUE #5895: Not scrobbling BBC Sounds
When listening to catch up on demand shows, the extension doesn't scrobble tracks played on the shows. Using chrome. Version 145.0.7632.117 (Official Build) (64-bit)

website link: https://www.bbc.co.uk/sounds/play/m002skfg

rob9315:
this is not a final fix, when you navigate away from the source page this does not detect anything playing anymore. I will have a fix for this but it will take some time to implement, this should be a good compromise until then.

also, the previous one only worked on the first track for some reason?

fixes #5895
fixes #5730

### DIFF
--- a/src/connectors/bbc-sounds.ts
+++ b/src/connectors/bbc-sounds.ts
@@ -1,4 +1,4 @@
-export { };
+export {};
 
 Connector.playerSelector = '[data-testid*="tracks"]';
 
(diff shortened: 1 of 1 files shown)


onedouglas-stack:
Thank you for this. Much appreciated.

Issue #5696, opened 2025-10-30
kcrw.com/shows/eclectic24


wososh:
**Describe the bug**
A streaming for page https://www.kcrw.com/shows/eclectic24/latest is appears to be supported

**How to reproduce**
Click Play on Eclectic24

**Expected behavior**
The scrobbler starts with recognizing the playback and submit scrobbles

<img width="1697" height="260" alt="Image" src="https://github.com/user-attachments/assets/3e3701a2-fe0b-4ca7-b942-9d61d15b90d3" />

-   OS: (Windows 11)
-   Browser: (Edge)
-   Extension version: (3.16.0)

swimtrunks:
Looks like KCRW has gone through a redesign of the entire site. Shouldn't be to terrible to go in and change the selectors in order to get this to work again. 

I have included a screenshot of the DOM Tree just in case some one has the time to change it:

<img width="961" height="589" alt="Image" src="https://github.com/user-attachments/assets/c27edc86-f36f-44a6-bcf7-2c2026edb123" />

swimtrunks:
Update : it looks like the new player doesn't show the artist for the currently playing song (that I can see). This isn't going to work, unless you are on a given shows episode page, in which case we could use that episodes playlist on the page as a work around. 

But unless they change this by showing metadata in the DOM again we are kinda stuck here for player scrobbling.

PR #5900, opened 2026-03-14
KCRW Connector : Support scrobbling on show pages

swimtrunks:
Changes : 
The KCRW site has been redesigned with a new player that doesn't supply the artist name to the DOM. The old connector relied on selectors (.site-player, .meta, .song-name, etc.) that no longer exist in the DOM. Will need an update if metadata is fed to DOM correctly. (submitted support ticket for this and will edit if changes are made)

Added support for` /shows` page(s) : tested on dj show episode pages and eclectic24 page.

addresses : #5696

### DIFF
--- a/src/connectors/kcrw.ts
+++ b/src/connectors/kcrw.ts
@@ -1,71 +1,70 @@
 export {};
 
 const filter = MetadataFilter.createFilter({
-	artist: [removeShowName, replaceSmartQuotes],
+	artist: [replaceSmartQuotes],
 	track: [
-		removeSmartQuotes,
 		replaceSmartQuotes,
 		MetadataFilter.removeVersion,
 		MetadataFilter.removeCleanExplicit,
 	],
 });
-
-Connector.playerSelector = '.site-player'; // supports main and mini players
+// we are using fuzzy matching for the classes in this connector because I think they are generated at run time and thus can potentially change on each build. This would make the only consistent part the suffix of the class name, ex: "AudioPlayer_audioController".
+Connector.playerSelector = '[class*="AudioPlayer_audioController"]';
 
 Connector.getTrackInfo = () => {
-	const metaSelector = '.site-player .meta';
-	const timeSelector = '.site-player .duration';
-	const showNameText = Util.getTextFromSelectors(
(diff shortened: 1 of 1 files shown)


hornbuckle:
Unfortunately, this update requires staying on a single show page to capture the playlist in order to work. But a user can easily navigate away from it and even go to a different show's playlist, and the player will still be playing the same show without capturing any info to scrobble anymore. It's a creative solution, but it doesn't work the way a connector needs to work.
This update also removes support for Today's Top Tune, which is still available, regularly updated, and may be the only thing that can still be fully supported by a connector for this site, since it includes both the artist and track info.
In this case, I wonder if it would be worth reaching out to KCRW's web team with a general inquiry about adding the artist back to the player somehow—maybe the same way they do for Today's Top Tune, or even just adding as an alt to the cover image instead of the track title would be helpful. Otherwise, their site can't really be supported by the extension anymore. It seems like they're pushing users more toward their app now, so they may be purposely making the site less useful. But it really is an odd choice to completely omit the artist altogether.

swimtrunks:
Thanks for this feedback! Yea I am not holding my breath waiting for the web team to get back to me on this lol I honestly forgot about this PR and meant to mark it as a draft.

hornbuckle:
No problem. I was just reviewing it to see why it was stuck in the queue and if I could help, since I had worked on this connector before. But I'm really not sure what to do for this one now. 😕

swimtrunks:
unfortunately this is becoming increasingly common in the connectors that i
have been looking to fix where there is a redesign that breaks integration
and there seems to be no way forward. heres hoping things don't trend that
way too much.

rob9315:
it really seems to become time to build some framework to inspect http requests with web-scrobbler

<details><summary>tracklist-api</summary>
image
</details>

hornbuckle:
Yep! I figured the info had to still be there _somewhere_ but didn't dig very far when I last looked.

(As an aside, I was going to mention that KCRW is on iheart.com—which is a whole other thing—but they aren't supplying any artist/track info there either. They also don't seem to transmit the info over FM either.)

PR #5905, opened 2026-03-16
Add support for Radio Nemiers

demiters:
Added a connector for radionemiers.com

### DIFF
--- a/src/connectors/radionemiers.ts
+++ b/src/connectors/radionemiers.ts
@@ -1,0 +1,9 @@
+export {};
+
+Connector.playerSelector = '.player-container';
+
+Connector.artistSelector =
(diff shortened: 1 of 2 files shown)


src/connectors/radionemiers.ts
```diff
@@ -0,0 +1,9 @@
+export {};
+
+Connector.playerSelector = '.player-container';
+
+Connector.artistSelector =
+	'.player-container .interface .area-metadata .artist a';
+Connector.trackSelector = '.player-container .interface .area-metadata .title';
```
rob9315: seems a bit overly specific
the player selector and artist/title should be enough

inverse (approved):
Thanks for the contribution!

PR #5906, opened 2026-03-16
Add support for Radio NABA

demiters:
Added a connector for the radio at naba.lv

**Additional context**
While the artist + title is shown on the main page, the player itself opens in a separate pop-up and does not contain the artist/title. Which means the playing state can not be detected in the same tab as where we read what is now playing, so the player has to be assumed to be always playing.

### DIFF
--- a/src/connectors/naba.ts
+++ b/src/connectors/naba.ts
@@ -1,7 +1,67 @@
+import type { ArtistTrackInfo } from '@/core/types';
+
 export {};
 
-Connector.playerSelector = '#tx_nabalive_live-song';
+const API_URL = 'https://www.lu.lv/api/?r=naba-live/search/';
+const POLL_INTERVAL = 30_000;
 
-Connector.artistTrackSelector = '#tx_nabalive_live-song';
+let currentSong: ArtistTrackInfo | null = null;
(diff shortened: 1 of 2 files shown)


rob9315:
the api url is `https://www.lu.lv/api/?r=naba-live/search/` if you maybe wanna implement that for the website playing 
audio instead?
if you have multiple tabs open this solution results in duplicate scrobbles, no?

demiters:
> the api url is `https://www.lu.lv/api/?r=naba-live/search/` if you maybe wanna implement that for the website playing audio instead? if you have multiple tabs open this solution results in duplicate scrobbles, no?

Yes that seems to be true. And I was hoping for a suggestion like that, will look into it shortly, thanks.

rob9315:
I tried doing something similar with a different connector and did not find a good way to do it with the current Connector API that isn't regular polling, just fyi

demiters:
API polling-based solution implemented, hopefully robust enough (polls every 30 seconds, all the while tracking play state and selected channel state to only scrobble when listening to the correct station)

rob9315:
it seems like the (player popup) website changed and it's not scrobbling anymore (`.jwplayer`)

PR #5907, opened 2026-03-16
Add support for Radio SWH

demiters:
Added a connector for play.radioswh.lv which contains 6 different radio stations, all of them supported with differences accounted for.

**Additional context**
Exceptions are handled to skip scrobbling status messages (commercial breaks etc).

### DIFF
--- a/src/connectors/radioswh.ts
+++ b/src/connectors/radioswh.ts
@@ -1,23 +1,30 @@
 export {};
 
-Connector.playerSelector = 'p.music-title';
+Connector.playerSelector = '#page-footer';
 
 Connector.getTrack = () => {
 	const span = document.querySelector('p.music-title span');
-	return span ? span.textContent?.trim() ?? null : null;
+	return span ? (span.textContent?.trim() ?? null) : null;
 };
 
 Connector.getArtist = () => {
 	const p = document.querySelector('p.music-title');
-	if (!p) return null;
+	if (!p) {
+		return null;
+	}
 	if (p.classList.contains('noafter')) {
-		const text = document.querySelector('p.music-sub-title span')?.textContent?.trim() || null;
+		const text =
+			document
+				.querySelector('p.music-sub-title span')
+				?.textContent?.trim() || null;
 		return text?.startsWith('SWH') ? null : text;
 	}
 	for (const node of p.childNodes) {
 		if (node.nodeType === Node.TEXT_NODE) {
 			const text = node.textContent?.trim();
-			if (text) return text.startsWith('SWH') ? null : text;
+			if (text) {
+				return text.startsWith('SWH') ? null : text;
+			}
 		}
 	}
 	return null;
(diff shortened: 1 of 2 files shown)


src/connectors/radioswh.ts
```diff
@@ -0,0 +1,31 @@
+export {};
+
+Connector.playerSelector = 'p.music-title';
```
rob9315:
```suggestion
Connector.playerSelector = '#page-footer';
```
`p.music-title` is not broad enough to capture changes like play/pause

src/connectors/radioswh.ts
```diff
@@ -0,0 +1,31 @@
...
+
+Connector.getTrack = () => {
+	const span = document.querySelector('p.music-title span');
+	return span ? span.textContent?.trim() ?? null : null;
```
rob9315:
```suggestion
	return span ? (span.textContent?.trim() ?? null) : null;
```
prettier lint will scream after the other lints are fixed

src/connectors/radioswh.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+Connector.getArtist = () => {
+	const p = document.querySelector('p.music-title');
+	if (!p) return null;
+	if (p.classList.contains('noafter')) {
+		const text = document.querySelector('p.music-sub-title span')?.textContent?.trim() || null;
```
rob9315:
```suggestion
		const text =
			document
				.querySelector('p.music-sub-title span')
				?.textContent?.trim() || null;
```
prettier lint will scream after the other lints are fixed

src/connectors/radioswh.ts
```diff
@@ -0,0 +1,31 @@
...
+Connector.playerSelector = 'p.music-title';
+
+Connector.getTrack = () => {
+	const span = document.querySelector('p.music-title span');
+	return span ? span.textContent?.trim() ?? null : null;
+};
+
+Connector.getArtist = () => {
+	const p = document.querySelector('p.music-title');
+	if (!p) return null;
```
rob9315:
```suggestion
	if (!p) {
		return null;
	}
```
eslint fix

src/connectors/radioswh.ts
```diff
@@ -0,0 +1,31 @@
...
+	if (p.classList.contains('noafter')) {
+		const text = document.querySelector('p.music-sub-title span')?.textContent?.trim() || null;
+		return text?.startsWith('SWH') ? null : text;
+	}
+	for (const node of p.childNodes) {
+		if (node.nodeType === Node.TEXT_NODE) {
+			const text = node.textContent?.trim();
+			if (text) return text.startsWith('SWH') ? null : text;
```
rob9315:
```suggestion
			if (text) {
				return text.startsWith('SWH') ? null : text;
			}
```
eslint fix

rob9315 (changes requested):

PR #5909, opened 2026-03-18
remove uses of innerHTML

rob9315:
uses of innerHTML seem to slow down firefox addon validation (maybe?), and it's basically only been used for links

also it nicely unifies what the links are like because `id`'s, and `rel` attributes sometimes differed.

i split the translation strings into their parts which aren't `<a>` tags, it's not pretty, yes.

**Additional context**
for release builds there is one remaining innerHTML use, in `context-menu`... this is from the solidjs template function and can't be avoided (?)

(yes i know `_locales` shouldn't be edited manually, i only did that so the PR can be tested and demonstrated)

(small helper function for splitting the strings by `<a>` tags: [trennen.js](https://github.com/user-attachments/files/26072495/trennen.js))

### DIFF
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -519,16 +519,40 @@
         "message": "Erstelle ein URL-Muster, das auf die spezielle Bandcamp-Domain-URL zeigt und übernimm die Änderungen.",
         "description": "Answer #4"
     },
-    "faqAnswer4c": {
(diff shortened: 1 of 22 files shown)


rob9315:
bad solution, should be using something similar to TPopupAnchor
<https://github.com/web-scrobbler/web-scrobbler/blob/master/src/ui/components/util.tsx>

Issue #5821, opened 2026-02-14
Error


c849839012:
My extension keeps on logging off, it appeared to me to report here with this following error:

href='https://github.com/web-scrobbler/web-scrobbler/issues' target='_blank'

rob9315:
Do you have a screenshot to look at? have you tried restarting everything, clearing the cache?
What browser do you use? What version of the extension, please provide more information.

rob9315:
no additional information has been given, there are other similar issues. if you can provide us information, don't hesitate to reopen the issue

PR #5918, opened 2026-03-22
translate anchor

rob9315:
currently `innerHTML` is used to display links, this removes many unnecessary uses of it

- changes regex for anchors in translation strings to support both `'` and `"` as `href` delimiters
- introduces a new TAnchor similar to TPopupAnchor which sanitizes translation strings

**Additional context**
- Popup Anchor was broken as seen in #5821, #5970 because of `'` instead of `"` for delimiting the href url in a translation string

### DIFF
--- a/src/ui/components/util.tsx
+++ b/src/ui/components/util.tsx
@@ -69,6 +69,7 @@ export function Anchor(props: {
 	href: string;
 	class?: string;
 	title?: string;
+	target?: '_blank' | '_top' | '_self' | '_parent';
 }) {
 	return (
 		<a
@@ -146,14 +147,25 @@ export function TPopupAnchor(props: Localization): JSXElement {
 	);
 }
 
-export function TAnchor(props: Localization): JSXElement {
+export function TAnchor(
+	props: Localization & {
+		class?: string;
+		target?: '_blank' | '_top' | '_self' | '_parent';
+	},
+): JSXElement {
 	const [res, setRes] = createSignal<(string | [string, string])[]>([]);
 	onMount(() => anchorOnMount(props, setRes));
 	return (
 		<For each={res()}>
 			{(item) => (
 				<Show when={typeof item !== 'string'} fallback={item}>
-					<Anchor href={item[0]}>{item[1]}</Anchor>
+					<Anchor
+						href={item[0]}
+						class={props.class}
+						target={props.target}
+					>
+						{item[1]}
+					</Anchor>
 				</Show>
 			)}
 		</For>
--- a/src/ui/components/util.tsx
+++ b/src/ui/components/util.tsx
@@ -1,5 +1,5 @@
-import type { JSXElement, Setter } from 'solid-js';
-import { For, Show, createSignal, onMount } from 'solid-js';
+import type { JSXElement } from 'solid-js';
+import { For, Show, createMemo } from 'solid-js';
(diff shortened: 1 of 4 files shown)


src/ui/components/util.tsx
```diff
@@ -64,52 +64,77 @@ export function PopupAnchor(props: {
 	);
 }
 
+export function Anchor(props: {
+	children: JSXElement | string;
+	href: string;
+	class?: string;
+	title?: string;
+	target?: '_blank' | '_top' | '_self' | '_parent';
```
inverse, line 72: Looks like we need to handle the target too, I dont think this is being passed.

src/ui/components/util.tsx
```diff
@@ -64,52 +64,77 @@ export function PopupAnchor(props: {
...
-			// eslint-disable-next-line
-			setRes((prev) => [
-				...prev,
-				message.slice(0, match.index),
-				[href, content],
-			]);
-			message = message.slice((match.index ?? 0) + str.length);
-		}
-	});
+	onMount(() => anchorOnMount(props, setRes));
```
inverse: Asked copilot and it suggested using a memo here instead to have a more deterministic behaviour and avoiding additional rendering.

src/ui/components/util.tsx
```diff
@@ -64,52 +64,77 @@ export function PopupAnchor(props: {
...
+			{props.children}
+		</a>
+	);
+}
+
 /**
  * Regex that finds anchor tag
  */
-const anchorRegex = /<a.*?href="([^"]+)".*?>([^<]+)<\/a>/;
+const anchorRegex = /<a.*?href=("|')((?:(?!\1).)+)\1.*?>([^<]+)<\/a>/;
```
inverse:
```suggestion
const anchorRegex = /<a.*?href=("|')((?:(?!\1).)+)\1.*?>([^<]+)<\/a>/g
```

What about using a global regex to match all URLs and then iterate over those instead of using maxIterations?

inverse (commented):
Left some feedback

src/ui/components/util.tsx
rob9315: forgot that one, good catch

src/ui/components/util.tsx
rob9315: not really a frontend guy, i just copied what was already there. good call tho, now it's much cleaner

src/ui/options/components/connector-override.tsx
```diff
@@ -58,7 +59,7 @@ export default function ConnectorOverrideOptions(props: {
 		<>
 			<h1>{t('optionsSupportedWebsites')}</h1>
 			<p>{t('optionsEnableDisableHint')}</p>
-			<p innerHTML={t('optionsCustomPatternsHint')} />
+			<TAnchor messageName="optionsCustomPatternsHint" target="_blank" />
```
inverse: Is this change needed? The text doesn't look like it has any anchors that need fixing

rob9315:
```suggestion
			<p>{t('optionsCustomPatternsHint')}</p>
```
no probably doesn't need that, i just blindly replaced all uses of innerHTML. it didn't need innerHTML either so i'll just make it into a paragraph

inverse (approved):
Thanks for the contribution!

PR #5919, opened 2026-03-22
Fix kinknl connector

### LINKED ISSUE #5911: Kink.nl player not scrobbling
The Kink.nl player is not scrobbling 

Go to
https://kink.nl/player/kink-distortion

Or the other channels ;-)

Click on play

Expected it to scrobble, like it used to

-   Windows 11
-   Browser: Google Chrome
-   Extension version: 3.20.0

rob9315:
they seem to be using their own player now? but in the network tab there's additionally still requests made to tritondigital and they do also still find metadata ... who knows
the connectors are brittle, there aren't any good classes or ids to target.

I experimented with network requests, but the current connector framework isn't set up to handle that well...

fix #5911

### DIFF
--- a/src/connectors/kinknl.ts
+++ b/src/connectors/kinknl.ts
@@ -1,22 +1,25 @@
 export {};
 
-Connector.playerSelector = '[class^=PopupPlayer_popupPlayer__player__]';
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks for the contribution!

PR #5922, opened 2026-03-22
feat: add poeba.lu connector

EndlessHallucination:
Adds a connector for poeba.lu, a music review site.

The site uses an embedded Bandcamp player. Since the player runs 
in a cross-origin iframe, metadata is read from the page's own 
album info section and floating player dock.

Currently scrobbles artist and album. Track-level scrobbling is 
not possible without iframe injection support, as the track name 
is only available inside the Bandcamp embed.

**Additional context**
Tested manually on multiple review pages. Scrobbles correct 
artist and album on each page.

### DIFF
--- a/src/connectors/poeba.lu.ts
+++ b/src/connectors/poeba.lu.ts
@@ -1,0 +1,26 @@
+export {};
+
+/**
+ * This connector supports poeba.lu music review site.
+ * Reads metadata from the album info section and floating player.
+ */
+
+Connector.playerSelector = '#floating-player-dock';
+
+Connector.artistSelector = '.album-meta dd:nth-child(2)';
+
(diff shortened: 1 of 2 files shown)


hornbuckle:
After reviewing and confirming by testing this locally, unfortunately, it does not satisfy the most basic requirements for a new connector. Since there is no actual player on the page, the extension thinks it is playing just by opening an album page, whether playback is initiated or not. Also, since there is no song title on the page, substituting the album title as a song title is an inaccurate scrobble and just creates bad data on Last.fm and other supported sites. Additionally, that would only equal a single scrobble, even if someone listened to an entire album, or it would repeat the same incorrect album as a song title until the page was closed.
There is another PR in the queue #5941 that was submitted for poeba.lu, which, with additional updates and testing, could potentially be used for all sites with Bandcamp embeds.

PR #5930, opened 2026-03-27
Fix hoopla connector

hornbuckle:
Hi! Long time, no see. I was unable to set up the V3 project locally without dependency errors and eventually gave up after a year or so of several tries. However, earlier today I noticed that the hoopla connector was no longer working and decided to give it another shot. Thankfully, it finally worked, and here we are again!

Anyway, hoopla made some changes to their site that required some selector updates. No real changes to the connector setup other than selectors, which are hopefully a bit less fragile and will work for a while again.

Since hoopla access is limited, here is a screenshot of the connector working again with the selector updates:

Since I haven't submitted anything in a long time, please let me know if there is anything else needed or anything I should keep in mind for next time. Thanks!

### DIFF
--- a/src/connectors/hoopladigital.ts
+++ b/src/connectors/hoopladigital.ts
@@ -16,7 +16,7 @@ Connector.getTrackArt = () => {
 	);
 
 	if (trackArtUrl) {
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks for the contribution!

PR #5934, opened 2026-03-29
Fix Audacy connector

### LINKED ISSUE #4789: The Audacy connector isn't working
I am currently unable to scrobble any music from Audacy.

Radio stations play with the mini-player by default, which is in the bottom right hand corner.

**Additional context**
Sample stations:
https://www.audacy.com/stations/metallicaradio?action=AUTOPLAY_FULL&actionContentId=102-7198 (Song Title - Artist)
https://www.audacy.com/search?action=AUTOPLAY_FULL&actionContentId=101-299 (Song Title - Artist)

hornbuckle:
Finally,* a fix for the long-broken Audacy connector! This will resolve #4789.

Station example: https://www.audacy.com/stations/live105
Podcast example: https://www.audacy.com/podcast/calm-it-down-354c0

- Got it working again, first of all. 😅
- Simplified connector functionality.
- Added podcast recognition.
- Updated disabled scrobbles to prevent more bad scrobbles.
- Added filter to remove dashes from artists.

**Additional context**
- After rebuilding the Audacy connector in #3550, they made an update that removed the mini player from the DOM when the large player was active, which broke it again. It was then fixed in #3595 to support both players. After trying to get that version to work again now and being unable to do so, I realized they have since reverted the update that removed the mini player from the DOM, so I could simplify the connector functionality again and only focus on that player. They also seem to have removed their playlist stations.
- Podcasts are signified by having forward and rewind options, which also ends up including some newscasts, but there is no way to differentiate one from the other.
- In addition to the previous disabled scrobbles, I noticed that there is often a delay before the artist loads, so it will now ensure scrobbling something very wrong like "Smooth Operator" by 94.7 The WAVE does not happen.
- As noted in my comments in #5935, some stations have inexplicably added dashes to artist names that are causing issues with lookups and possibly creating bad artists on Last.fm. Adding a simple filter to remove dashes from artist names seems to fix it, while most artists that should have dashes in their names should still scrobble correctly—blink-182 and The B-52's are a couple that specifically came up while testing.

*I have wanted to fix this for years but was unable to set up the project locally without errors after the V3 revamp until just this week when I tried again, and it finally worked for me.

### DIFF
--- a/src/connectors/audacy.ts
+++ b/src/connectors/audacy.ts
@@ -45,5 +45,8 @@ Connector.scrobblingDisallowedReason = () => {
 Connector.applyFilter(filter);
 
 function removeDashes(text: string) {
+	if (Connector.isPodcast()) {
+		return text;
+	}
 	return text.replace(/-/g, ' ');
 }
--- a/src/connectors/audacy.ts
+++ b/src/connectors/audacy.ts
@@ -1,79 +1,42 @@
 export {};
 
-const audioPlayer = 'div[aria-label="Audio player"]';
-const buttonOpenFullscreen = 'button[aria-label="Open full-screen player"]';
-const buttonCloseFullscreen = 'button[aria-label="Close full-screen player"]';
-const fullArtistTrackSelector = `${buttonCloseFullscreen} + div span`;
+const audioPlayer = '[aria-label="Audio player"]';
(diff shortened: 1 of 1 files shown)


inverse (approved):
Unable to verify due to geo-blocks, but looks good. Thanks for the contribution!

PR #3345, opened 2022-08-31
Add MUSICat connector with support for 14 local library sites

hornbuckle:
Added support for 14 local library sites, which all use the MUSICat platform: https://musicat.co/libraries

One connector file (musicat.js) is used for all sites. Tested each site and scrobbling is successful.

Unlike other library sites, such as Freegal or hoopla, no library card login is required to listen to the available music. Example album links from each site provided below for additional testing/verification.

- Amplify 817 - https://amplify817.org/albums/driving-slow-motion-arda
- Capital City Records - https://capitalcityrecords.ca/albums/golden-grey-8-bit-machine
- Electric Lady Bird - https://atxlibrary.musicat.co/albums/john-dixon-attack-of-the-demons-soundtrack
- FlipSide - https://flipside.apl.org/albums/james-hall-lattice
- HUM (Hear Utah Music) - https://hum.slcpl.org/albums/ava-lux-princess-heaven
- KDL Vibes - https://vibes.kdl.org/albums/eli-kahn-how-are-you-no-really-how-are-you
- Library Music Project - https://librarymusicproject.com/albums/boink-something-colorful-for-sure
- MNspin - https://hclib.musicat.co/albums/ie-pome
- Nashville BoomBox - https://boombox.library.nashville.org/albums/jay-leo-phillips-one-million-one-million-one-million
- QC Beats - https://qcbeats.org/albums/kumate-69-chambers
- Sawdust City Sounds - https://sawdustcitysounds.org/albums/tonebox-last-encryption
- Seattle PlayBack - https://playback.spl.org/albums/acid-tongue-bullies
- STACKS - https://stacks.carnegielibrary.org/albums/else-collective-else-collective
- Tracks Music Library - https://tracksmusiclibrary.org/albums/dustlights-in-a-stillness

Issue #2853, opened 2021-05-08
Brooklynvegan.com


IndieMonster:
Can you add support for them?  Most of there content is embed youtube or bandcamp tracks so instead of having to open a new window to capture the scrobble, it would be great to just get the scrobble there.  Same for Stereogum and Paste if all those could be added.

rob9315:
youtube embeds are broken right now, bandcamp embeds will be fixed soon

PR #5941, opened 2026-04-04
Add connector for poeba.lu bandcamp embedded player

a4180p:

Add scrobbling support for https://poeba.lu
The embedded player runs inside an iframe at `bandcamp.com/EmbeddedPlayer/...`

**Additional context**

### DIFF
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -38,7 +38,7 @@ export default <ConnectorMeta[]>[
 		allFrames: true,
 	},
 	{
-	        label: 'poeba.lu (embedded bandcamp player only)',
+		label: 'poeba.lu (embedded bandcamp player only)',
 		matches: ['*://bandcamp.com/EmbeddedPlayer/*'],
 		js: 'poeba.lu.js',
 		id: 'poeba.lu',
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -38,6 +38,13 @@ export default <ConnectorMeta[]>[
 		allFrames: true,
 	},
 	{
+		label: 'poeba.lu',
+		matches: ['*://bandcamp.com/EmbeddedPlayer/*'],
+		js: 'poeba.lu.js',
+		id: 'poeba.lu',
+		allFrames: true,
+	},
+	{
 		label: 'Bandcamp',
 		matches: ['*://*.bandcamp.com/*', '*://bandcamp.com/*'],
 		js: 'bandcamp.js',
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -38,7 +38,7 @@ export default <ConnectorMeta[]>[
 		allFrames: true,
 	},
 	{
-		label: 'poeba.lu',
+	        label: 'poeba.lu (embedded bandcamp player only)',
 		matches: ['*://bandcamp.com/EmbeddedPlayer/*'],
 		js: 'poeba.lu.js',
 		id: 'poeba.lu',
--- a/src/connectors/poeba.lu.ts
+++ b/src/connectors/poeba.lu.ts
@@ -2,9 +2,6 @@ export {};
 
 /**
  * Connector for the Bandcamp embedded player on poeba.lu.
- *
- * The embedded player runs inside an iframe at bandcamp.com/EmbeddedPlayer/...
- * State: the root #player element gets the class "playing" when audio is playing.
  */
 
 Connector.playerSelector = '#player';
--- a/src/connectors/poeba.lu.ts
+++ b/src/connectors/poeba.lu.ts
@@ -1,0 +1,23 @@
+export {};
+
+/**
+ * Connector for the Bandcamp embedded player on poeba.lu.
+ *
+ * The embedded player runs inside an iframe at bandcamp.com/EmbeddedPlayer/...
+ * State: the root #player element gets the class "playing" when audio is playing.
+ */
+
+Connector.playerSelector = '#player';
+
+Connector.trackSelector = '#currenttitle_title';
+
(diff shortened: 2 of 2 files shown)


hornbuckle:
Thinking about this, and it seems like there should be a better way to do this, in order to scale a connector to support Bandcamp embeds on other sites.
There is currently a "Bandcamp Daily" connector named "[bandcamp-embed](https://github.com/web-scrobbler/web-scrobbler/blob/master/src/connectors/bandcamp-embed.ts)," but it seems like that one should be renamed to "bandcamp-daily" because it is for a specific player that is only used on https://daily.bandcamp.com/
Then this connector—with further testing—could become a more appropriately-named "bandcamp-embed" and reused on other sites that rely on embeds from Bandcamp, as suggested in #2853 and similar to #3345. (Additionally, the notes included in the connector file would be more appropriate in the initial PR comment instead of in the file itself.)
Would want to get @inverse thoughts on this before making any additional changes, of course.

a4180p:
@hornbuckle thanks for the review!
I've removed extra comments from the file.

hornbuckle:
Looking at this again, and I believe there are at least a couple of reasons why something like this hasn't been done successfully before:
1. Like the sites suggested in the earlier request mentioned above, this site uses various embeds—including not just Bandcamp, but also Spotify and Tidal. But the extension can only support one at a time.
2. This doesn't work the way it's been submitted. The extension needs the site URL to be activated when on that site, so it only works when opening the embedded Bandcamp player in a new tab.

I believe, while I appreciate the intention here, that this request will unfortunately not be supported—mostly because of the problem of trying to support multiple embeds on one site. And while my comments yesterday were imagining a blanket "bandcamp-embed" connector, the reality is that there is an endless list of sites that could be added to try to support that, and it's just not really scalable or practical.

a4180p:
> This doesn't work the way it's been submitted. The extension needs the site URL to be activated when on that site, so it only works when opening the embedded Bandcamp player in a new tab.

I don't know how, but it works.
image

hornbuckle:
Ah, I see it is working locally for me now too. (I had removed `allFrames: true` while testing, so it wasn't working without that but worked after putting it back.) So, maybe my comments from yesterday are possible after all, without having to support each site URL individually…

a4180p:
> not just Bandcamp, but also Spotify and Tidal

I’ve changed the label of the connector to clearly indicate that it supports Bandcamp only.

hornbuckle:
Let's wait for @inverse to review before making any other changes. This is potentially bigger than poeba.lu, and I just realized there is already another PR specifically for that site #5922.

src/core/connectors.ts
```diff
@@ -37,6 +37,13 @@ export default <ConnectorMeta[]>[
 		id: 'bandcamp-embed',
 		allFrames: true,
 	},
+	{
+	        label: 'poeba.lu (embedded bandcamp player only)',
```
a4180p:
```suggestion
		label: 'poeba.lu (embedded bandcamp player only)',
```

src/core/connectors.ts
```diff
@@ -37,6 +37,13 @@ export default <ConnectorMeta[]>[
 		id: 'bandcamp-embed',
 		allFrames: true,
 	},
+	{
+		label: 'poeba.lu (embedded bandcamp player only)',
+		matches: ['*://bandcamp.com/EmbeddedPlayer/*'],
+		js: 'poeba.lu.js',
+		id: 'poeba.lu',
+		allFrames: true,
+	},
```
hornbuckle, line 46: To clarify, this is part of what I am proposing in order to proceed with this new connector:
- Existing Bandcamp Daily connector is renamed more appropriately to "bandcamp-daily"
- New connector is renamed to "bandcamp-embed" (it has nothing specifically to do with poeba.lu anymore)

```
	{
		label: 'Bandcamp Daily',
		matches: ['*://daily.bandcamp.com/*'],
		js: 'bandcamp-daily.js',
		id: 'bandcamp-daily',
		allFrames: true,
	},
	{
		label: 'Bandcamp Embed',
		matches: ['*://*.bandcamp.com/EmbeddedPlayer/*'],
		js: 'bandcamp-embed.js',
		id: 'bandcamp-embed',
		allFrames: true,
	},
```
I have an additional suggestion to get the track art in the connector file, but for now please hold off on pushing any more changes to this PR (including the above suggestion). I want to get @inverse opinion on whether or not it is okay to proceed with a Bandcamp Embed connector that will work on any site that has one. I have to wonder if this was tried before and there were some issues, or else why hasn't it been done before?

a4180p:
@inverse @hornbuckle are there any chances of moving forward with this PR?

hornbuckle:
Hopefully you don't mind, but I've submitted a new PR #6009 to clarify the intended behavior for a new Bandcamp embed connector, since this one still needs updates to proceed and has so far been overlooked.

a4180p:
> this one still needs updates to proceed

well, it needs updates only because you've asked not to update, but if you think that PR with another author has more chances to be noticed, I'm ok.

hornbuckle:
Thanks. Yes, I know I did ask that, but only because I was hoping for an additional review sooner and didn't want to test your patience any further with additional updates (of which there could potentially be more than I'd already mentioned).

rob9315:
@hornbuckle how about you rebase your pr off this one?
then we can keep proper attribution, we can even merge this one if bandcamp support needs more discussion
you would do a git mv on the connector file to correct the file naming in your branch

hornbuckle:
@a4180p A full rebase wasn't working, so I've added you as co-author on the initial connector file commit in the new PR. Closing this PR may be best now, so we can focus on moving the other one forward.

PR #5943, opened 2026-04-06
Fix Tidal connector

Kahlis:
- Fixed the `artistSelector` for Tidal's new UI
- Added a new variable called `albumArtistSelector` that extends the `artistSelector` with `:first-child` so it captures the first artist in the artist row _(which match the albumArtist by the platform's design)_
- Cleaned up `Connector.getAlbumArtist` by refactoring the extraction; now it is cleaner and effective, without over-engineering logic

**Additional context**
`Connector.getAlbumArtist` edition must be validated before merged; I may be missed some business rules from Tidal and it might borderline a issue. In my general view it doesn't, but double check is never too much.

### DIFF
--- a/src/connectors/tidal.ts
+++ b/src/connectors/tidal.ts
@@ -27,7 +27,8 @@ Connector.getUniqueID = () => {
 	return null;
 };
 
-const artistSelector = `${Connector.playerSelector} span.artist-link a`;
+const artistSelector = `${Connector.playerSelector} a[data-test="grid-item-detail-text-title-artist"]`;
+const albumArtistSelector = `${artistSelector}:first-child`;
 
 Connector.getArtist = () => {
 	const artistNodes = document.querySelectorAll(artistSelector);
@@ -40,25 +41,8 @@ Connector.albumSelector = [
 ];
 
 Connector.getAlbumArtist = () => {
-	const albumUrlSegments = Util.getAttrFromSelectors(
-		Connector.albumSelector,
-		'href',
-	)?.split('/');
-	const pageUrlSegments = Util.getAttrFromSelectors(
-		'head meta[name="apple-itunes-app"]',
-		'content',
-	)?.split('/');
-	if (
-		'album' === albumUrlSegments?.at(-2) &&
-		'album' === pageUrlSegments?.at(-2) &&
-		albumUrlSegments?.at(-1) === pageUrlSegments?.at(-1)
-	) {
-		const albumArtistNode = document.querySelectorAll(
-			'main div[class^="_detailContainer"] .artist-link a',
-		);
-		return Util.joinArtists(Array.from(albumArtistNode));
-	}
-	return null;
+	const albumArtistNode = document.querySelectorAll(albumArtistSelector);
+	return Util.joinArtists(Array.from(albumArtistNode));
 };
 
 Connector.trackArtSelector = `${Connector.playerSelector} figure[data-test="current-media-imagery"] img`;


Kahlis:
I can close this PR and send clear tree one instead if requested

inverse (approved):
Thanks for the contribution!

jaccarmac:
Sorry that I've been inactive and am late to this, but

> the first artist in the artist row (which match the albumArtist by the platform's design)

is not correct. This creates incorrect scrobbles for certain collaborative projects. The old logic, while tortured, is about as well as we can do for features, which I think is what this change targets.

jaccarmac:
Thanks for the selector updates! I do always appreciate additional eyes on this connector.

src/connectors/tidal.ts
```diff
@@ -27,7 +27,8 @@ Connector.getUniqueID = () => {
 	return null;
 };
 
-const artistSelector = `${Connector.playerSelector} span.artist-link a`;
```
jaccarmac, line 30: This old selector still works for me. Why did it need to change to use `[data-test]`?

src/connectors/tidal.ts
```diff
@@ -40,25 +41,8 @@ Connector.albumSelector = [
...
-		albumUrlSegments?.at(-1) === pageUrlSegments?.at(-1)
-	) {
-		const albumArtistNode = document.querySelectorAll(
-			'main div[class^="_detailContainer"] .artist-link a',
-		);
-		return Util.joinArtists(Array.from(albumArtistNode));
-	}
-	return null;
+	const albumArtistNode = document.querySelectorAll(albumArtistSelector);
+	return Util.joinArtists(Array.from(albumArtistNode));
```
jaccarmac, line 45: Specifying `:first-child` means there will only be one artist in this selection and `joinArtists` is a no-op.

jaccarmac (commented):
Other than changing the determination of album artist (in a way I believe is incorrect), this doesn't appear to solve any issues. I suggest a full revert.

src/connectors/tidal.ts
Kahlis: Yes, only one artist since `albumArtist` means the the artist that owns the album, not all the artists that are part of it (such as feat or any special participating). The first artist is always the albumArtist by default in Tidal, so thats why I choose this approach.

src/connectors/tidal.ts
Kahlis: If the change still functional for you I see no point in criticism. It needs to work for everyone, not only for you, that's the point here

src/connectors/tidal.ts
Kahlis: I did understand now, I've used incorrectly the joinArtists, I'm going to fix this asap

src/connectors/tidal.ts
jaccarmac: Apparently I needed to hard-refresh to get the new UI, so I was indeed missing the buggy behavior.

src/connectors/tidal.ts
jaccarmac: > The first artist is always the albumArtist by default in Tidal.

That is not true. The album artist for compilations is "Various Artists"; The album artist for the recent NIN collab is not NIN but "Nine Inch Nails, Boys Noize".

jaccarmac:
I was incorrect to identify the entire PR as unnecessary.

The redesign changes more than the location of the artist, and it might not be
possible to restore the old album behavior in its current state. I do want to
finish refactoring for the redesign, but we should open new issues as
necessary.

PR #5960, opened 2026-04-15
Fix iHeart connector

### LINKED ISSUE #5809: iHeartRadio no longer scrobbling
It appears iHeartRadio redesigned their radio pages and the extension no longer scrobbles from the site. Here's WFUV's page as an example.

https://www.iheart.com/live/907-wfuv-6430/

-   OS: Mac
-   Browser: Chrome
-   Extension version: 3.19.0

### LINKED ISSUE #5951: iheart site changed
new site means need new connector. if anyone can provide new url pattern for iheart radio before the app updates that would be cool

hornbuckle:
Fixes for iHeart connector after site updates. Resolves #5809 and resolves #5951.

In addition to updating the connector for the current site, I also added podcast and track art support.

Example station: https://www.iheart.com/live/alt-987-201
Example playlist: https://www.iheart.com/playlist/newly-released-312064750-JwZCfk5kxHoV7u2yFQ7hQC
Example podcast: https://www.iheart.com/podcast/1119-stuff-you-should-know-26940277

**Additional context**
The player does this buggy thing where it randomly switches between the artist/track info and the station information. When this switch happens, the track goes null and the artist changes to the station name. Sometimes a song can play all the way through without it happening, in which case a scrobble can be completed without issue. But often, the switch will happen every few seconds, making the connector continually go into an empty/reset state and making completing a scrobble impossible. Playlists and podcasts don't seem to have this issue.
In an attempt to mitigate this issue, a 10 second delay has been added. If, after 10 seconds, the track is still null and artist is the station name, empty/reset state can proceed because it typically means either a DJ is talking or an ad is playing and nothing should scrobble anyway. Sometimes the switch can be longer than 10 seconds, and sometimes it never goes back to the artist/track info while the song is playing, so there will occasionally be missed or duplicate scrobbles, but there's only so much we can do with the connector when the site is wack.
Thanks very much to @rob9315 for providing a solution to implement the delay that was needed for this connector.
(There was a temporary glitch where Firefox and Safari were getting the old version of the site while Chrome was still getting the new site, which seems to be resolved now.)

Also, while digging around the site, I noticed that full albums are available to stream with a paid subscription. So, I wonder if the album title is made available in the player when listening to one. Unfortunately, there is no way to even preview that without an account.

### DIFF
--- a/src/connectors/iheart.ts
+++ b/src/connectors/iheart.ts
@@ -5,6 +5,13 @@ export {};
 const playerBar = '[data-test=player-container]';
 const controlBar = '[data-test=player-controls]';
 const artistSelector = `${playerBar} [data-test=description-link]`;
+const filter = MetadataFilter.createFilter({
+	track: [
+		MetadataFilter.removeRemastered,
+		MetadataFilter.removeVersion,
+		MetadataFilter.removeLive,
+	],
+});
 let artistTrack: ArtistTrackInfo | null = null;
 let timeout: ReturnType<typeof setTimeout> | undefined;
 
@@ -49,6 +56,8 @@ Connector.isPlaying = () => {
 	return svgLabel === 'Stop';
 };
 
+Connector.applyFilter(filter);
+
 function getArtistTrack() {
 	const artist = Util.getTextFromSelectors(artistSelector);
 	let trackSelector = `${playerBar} [data-test=title-link]`;
@@ -57,7 +66,10 @@ function getArtistTrack() {
 		trackSelector = `${playerBar} [data-test=subtitle-link]`;
 	}
 
-	const track = Util.getTextFromSelectors(trackSelector)?.replace('•', '-');
+	const track = Util.getTextFromSelectors(trackSelector)?.replace(
+		'\u2022',
+		'-',
+	);
 
 	return { artist, track };
 }
--- a/src/connectors/iheart.ts
+++ b/src/connectors/iheart.ts
@@ -1,8 +1,12 @@
+import type { ArtistTrackInfo } from '@/core/types';
+
 export {};
 
 const playerBar = '[data-test=player-container]';
 const controlBar = '[data-test=player-controls]';
 const artistSelector = `${playerBar} [data-test=description-link]`;
+let artistTrack: ArtistTrackInfo | null = null;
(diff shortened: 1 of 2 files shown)


hornbuckle:
@inverse This should be good to go now.

src/connectors/iheart.ts
```diff
@@ -1,19 +1,79 @@
...
+	if (Connector.isPodcast()) {
+		trackSelector = `${playerBar} [data-test=subtitle-link]`;
+	}
+
+	const track = Util.getTextFromSelectors(trackSelector);
+
+	return { artist, track };
+}
+
+function isArtistTrackValid(artistTrack: ArtistTrackInfo) {
```
inverse: This could be simplified too 

```js
function isArtistTrackValid(artistTrack: ArtistTrackInfo): boolean {
  return !!artistTrack.track;
}
```

src/connectors/iheart.ts
```diff
@@ -1,19 +1,79 @@
...
+	if (!artistTrack.track) {
+		return false;
+	}
+
+	return true;
+}
+
+function renewTimeout() {
```
inverse, line 77: Is this needed? won't the default handling of watching for DOM changes handle this?

inverse (commented):
Looks good, left some comments

src/connectors/iheart.ts
rob9315: some radio stations (specifically one that was requested and tested on) change between their station name and the actual track being played frequently, also in the middle of the track, leading to missed scrobbles or multi-scrobbles

src/connectors/iheart.ts
hornbuckle: Updated with this suggestion. Thanks!

src/connectors/iheart.ts
hornbuckle: Yes, normally the default handling would suffice, but this site behavior is abnormal. Adding the delay helps a scrobble complete as expected in most cases.

src/connectors/iheart.ts
hornbuckle: Realized there is an existing Util function that will work and updated to use that instead of a custom function.

PR #5963, opened 2026-04-18
handle last.fm error 9

rob9315:
error means: Invalid session key - Please re-authenticate
see docs: https://www.last.fm/api/errorcodes

which means that last.fm or the user revoked our session key and we have to make the user go through the sign-in flow again.

call `signOut` for error 9

**Additional context**
- should this check that we are talking to Last.fm or is this also valid for other AudioScrobblers?

### DIFF
--- a/src/core/scrobbler/audio-scrobbler/audio-scrobbler.ts
+++ b/src/core/scrobbler/audio-scrobbler/audio-scrobbler.ts
@@ -329,6 +329,13 @@ export default abstract class AudioScrobbler extends BaseScrobbler<'LastFM'> {
 
 		if (!response.ok) {
 			this.debugLog(`${params.method} response:\n${debugMsg}`, 'error');
+			if ('error' in responseData && responseData.error === 9) {
+				// Invalid session key - Please re-authenticate
+				await this.signOut();
+				this.debugLog(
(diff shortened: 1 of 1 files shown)


src/core/scrobbler/audio-scrobbler/audio-scrobbler.ts
```diff
@@ -329,6 +329,13 @@ export default abstract class AudioScrobbler extends BaseScrobbler<'LastFM'> {
 
 		if (!response.ok) {
 			this.debugLog(`${params.method} response:\n${debugMsg}`, 'error');
+			if ('error' in responseData && responseData.error === 9) {
```
rob9315, line 332:
```suggestion
			if (
				this instanceof LastFmScrobbler &&
				'error' in responseData &&
				responseData.error === 9
			) {
```
if it should be limited to last.fm, then this should do that

inverse: I think it should be fine to leave the code 9 check in place in this layer as it's only used in librefm and lastfm scrobblers and from briefly checking it should support this error code.

inverse (approved):
Thanks for the contribution!

PR #5964, opened 2026-04-18
add support for escradio.com

### LINKED ISSUE #5954: ESC Radio
The #1 Eurovision radio station. Doesn't require accounts.

https://www.escradio.com/player/

rob9315:
fixes #5954


rob9315:
yeahhhhh.... no. let's try that again

PR #5969, opened 2026-04-21
Refactor for the TIDAL design update.

jaccarmac:
# Remove selectors which no longer apply to the TIDAL redesign.

Note that there is no longer any way to get the album for the current track
unless it is being played from that album.

Note that the site appears to no longer install a service worker to operate
offline and manage updates. However, the old service worker will not uninstall
itself, which means that a new tab will open the old UI. A hard refresh loads
the new site; Removing the service worker from the browser prevents the old
code from loading. I noticed this switching between TIDAL versions 2026.03.03
and 2026.04.16.

# Restore old album artist behavior.

Partially reverts #5943.

### DIFF
--- a/src/connectors/tidal.ts
+++ b/src/connectors/tidal.ts
@@ -11,10 +11,7 @@ Connector.pauseButtonSelector = `${Connector.playerSelector} button[data-test="p
 Connector.scrobblingDisallowedReason = () =>
 	Util.queryElements(Connector.playButtonSelector) ? null : 'ElementMissing';
 
-Connector.trackSelector = [
-	'#nowPlaying span[data-test="now-playing-track-title"]',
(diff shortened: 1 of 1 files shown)


jaccarmac:
Logging out and in does not force the new UI. I have confirmed using another
Firefox that had a logged-in session.

I'm interested to see if there's any way to manually install the service worker
once unregistered, but my web API knowledge has rusted too much to do that
quickly.

jaccarmac:
Revisiting the YouTube Music connector has reminded me that the MediaSession
API exists. Might be about time to switch this connector over, it carries album
information.

inverse:
@jaccarmac do you wanna explore that before applying this change?

jaccarmac:
No, this works as a unit (really two but it's dependent on consensus on the album artist discussion).

Issue #4644, opened 2024-04-13
Show default edits in the page and allow modification


soreikomori:
**Is your feature request related to a problem? Please describe.**
web-scrobbler has a few default edits like removing "(feat. x)" or "(Music Video)" from scrobbles. This is great in most cases to keep last.fm clean, but it can cause some issues sometimes, especially when web-scrobbler is used in tandem with another scrobbler in another platform (say in Android), since one scrobbler might remove those tags and the other won't. In the end, this should be left to the discretion of the user- If I want my track to have the "(feat. x)" then web-scrobbler should have an option to stop executing those default edits.

**Describe the solution you'd like**
To give the user more control over what _exactly_ gets edited out, the default web-scrobbler edits such as that one should be shown in the regex/bulk edits page in the extension settings, and simply be turned on by default. The user then should be able to remove them if need be.

**Describe alternatives you've considered**
Maybe a single slider button that can enable or disable "recommended edits" or "default edits" or something of the sort.

jonnybel:
+1, same issue on YouTube Music — the connector pulls metadata from `navigator.mediaSession` (which YT Music already strips) instead of the DOM like most other connectors. The concrete pain: scrobble a whole album and the tracks whose Last.fm canonical has "(feat. X)" don't match, so they end up as orphan singles detached from the album — splits play counts, breaks the album stats. I'm working on a fix, notifying the maintainers on discord.

jonnybel:
Follow-up after testing more connectors: turns out `(feat. X)` actually isn't stripped by Web Scrobbler's filter layer on most services — Bandcamp, Tidal, Spotify, and YT Music all preserve it fine. The exception is Amazon Music, which runs `removeFeature` in its default filter. So Amazon scrobbles strip the suffix while every other service leaves it alone. Same user-visible symptom, different cause — probably its own discussion.

For YT Music specifically, the strip happens *before* any filter layer: the connector reads the title from `navigator.mediaSession.metadata.title`, which YouTube Music itself pre-cleans. Fix is to read it from the DOM like most other connectors.

Issue #4991, opened 2024-10-20
Tracks titles edited despite no regex/bulk edits on YouTube Music


Alespren:
**Describe the bug**
When scrobbling Charli XCX's music on YouTube Music, tracks that feature other artists are being edited to remove the featured artist. For example "Guess (feat. Billie Eilish)" is scrobbled as "Guess". I have no regex/bulk filters, and no edited tracks. This is a problem because Guess and Guess (feat. Billie Eilish) are distinctly different songs and there's no way to distinguish that if the full title isn't scrobbled.

**How to reproduce**
1. Play "Guess (feat. Billie Eilish)" or any other song on the brat remix album on YouTube Music
2. Check what is scrobbled

**Expected behavior**
The track should be scrobbled as "Guess (feat. Billie Eilish)", which is what the track is titled on YouTube Music.

**Screenshots**

**Debug logs**
It's strongly recommended to grab [debug logs](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) after reproducing the issue.

**Environment (please complete the following information)**
-   OS: Windows 11
-   Browser: Arc Browser
-   Extension version: v3.10.0

**Additional context**

jaccarmac:
Do you have the "Use Youtube Music to get track info" option enabled or disabled? It's on the main page of options.

Alespren:
I've tested it with the option enabled and disabled, with no change in behavior.

jaccarmac:
Thanks for checking that. I can duplicate the behavior in Chrome, though it was a little hard to pull up the non-MV version of the song. At first blush the problem seems to be in the media state information coming from Chrome. If you visit the song on plain old YouTube do you also get a bad scrobble? https://www.youtube.com/watch?v=Q4d7SV9LYz4 for me results in "Guess feat. billie eilish (feat. Billie Eilish)"

jaccarmac:
I can confirm in incognito Chrome (that is, no extension) that the `mediaSession` is the problem. So while I understand this is a problem with the track, it is technically the expected behavior. That object usually provides better information than the DOM and I don't see a way to detect that the behavior should change here. If you edit the scrobbling information, the extension should save it for the video.

![image](https://github.com/user-attachments/assets/5c04ca59-d258-4b1a-aa13-2cd74d96532b)

PR #5972, opened 2026-04-21
Fix track title scrobbling on YouTube Music

jonnybel:
## Problem

On YouTube Music, tracks with `(feat. X)` in the title were scrobbling as the stripped version (`Song` instead of `Song (feat. Artist)`), so they wouldn't match the canonical track on Last.fm. Tracks from the same album ended up as orphan singles detached from the album — splits play counts, breaks album stats.

## Investigation

Started out thinking this was the default-filter issue described in #4644 and was planning to add a toggle to disable it. But the default filter (the one `BaseConnector` applies on top of each connector's filter) only does `trim` + `&nbsp;` normalization, nothing that would strip `(feat. X)`. Checked the per-connector filters too, and none of the major ones (Bandcamp, Tidal, Spotify, YT Music) strip it either.

Real cause on YT Music: the connector was reading the track title from `navigator.mediaSession.metadata.title`, which YouTube Music pre-cleans before exposing. So the stripped title never hit WS's filter layer.

## Fix

Read the track title from `.ytmusic-player-bar .title` in the DOM, same source most other connectors use. Worth flagging: YouTube Music is internally inconsistent here — the DOM element has the raw title with `(feat. X)` preserved, while MediaSession has the stripped version. Switching to the DOM isn't a workaround, just picking the source they themselves expose with the raw label.

Artist and album stay on MediaSession — they're structured fields, less fragile than parsing the byline, and don't get the same suffix-strip treatment.

## Related observation: Amazon Music filter inconsistency

While investigating, noticed that across the built-in filters in `@web-scrobbler/metadata-filter`, **`createAmazonFilter()` is the only one that uses `removeFeature`**. So Amazon Music scrobbles strip `(feat. X)` at the filter layer while every other service preserves it. Same user-visible symptom as this YT Music bug, different mechanism.

Happy to also drop `removeFeature` from `createAmazonFilter()` in this PR to make scrobble behavior consistent across services, or open a separate PR — whichever you prefer.

## Refs

- #4644 — broader default-filter concern this PR doesn't fully address (only fixes the YT Music case users have actually been hitting)

### DIFF
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -22,20 +22,7 @@ export {};
 
 const adSelector = '.ytmusic-player-bar.advertisement';
 
-/**
- * Track title is read from the DOM rather than `mediaSession.metadata.title`
- * because YouTube Music sets a pre-cleaned title on the MediaSession object
- * (e.g. strips "(feat. X)"). The DOM element preserves the raw title so
- * the scrobble matches the canonical track on the scrobbling service.
- *
- * If the DOM selector stops matching (YT Music UI changes, element missing),
- * `getArtistTrack` falls back to `mediaSession.metadata.title` so scrobbling
- * still works — just with the pre-cleaned title until the selector is
- * updated.
- *
- * Artist and album still use MediaSession — they're structured fields and
- * don't get the same suffix-strip treatment.
- */
+// track title in `mediaSession` is stripped compared to DOM
 const titleSelector = '.ytmusic-player-bar .title';
 
 const mediaInfo = {
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -27,6 +27,12 @@ const adSelector = '.ytmusic-player-bar.advertisement';
  * because YouTube Music sets a pre-cleaned title on the MediaSession object
  * (e.g. strips "(feat. X)"). The DOM element preserves the raw title so
  * the scrobble matches the canonical track on the scrobbling service.
+ *
+ * If the DOM selector stops matching (YT Music UI changes, element missing),
+ * `getArtistTrack` falls back to `mediaSession.metadata.title` so scrobbling
+ * still works — just with the pre-cleaned title until the selector is
+ * updated.
+ *
  * Artist and album still use MediaSession — they're structured fields and
  * don't get the same suffix-strip treatment.
  */
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -22,6 +22,16 @@ export {};
 
 const adSelector = '.ytmusic-player-bar.advertisement';
 
+/**
+ * Track title is read from the DOM rather than `mediaSession.metadata.title`
+ * because YouTube Music sets a pre-cleaned title on the MediaSession object
+ * (e.g. strips "(feat. X)"). The DOM element preserves the raw title so
+ * the scrobble matches the canonical track on the scrobbling service.
+ * Artist and album still use MediaSession — they're structured fields and
+ * don't get the same suffix-strip treatment.
+ */
+const titleSelector = '.ytmusic-player-bar .title';
+
 const mediaInfo = {
 	playbackState: 'none',
 	metadata: {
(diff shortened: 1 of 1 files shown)


rob9315:
I'd suggest falling back to the old version of getting the title for when the selectors change.
As for removing feat., i think a global setting for that would be ideal to toggle having it on all or none of the connectors, but it's probably worth looking into why the amazon filterset has that

jonnybel:
@rob9315 Yes, good call — the fallback's actually already there via the `||` in `getArtistTrack`: if the selector returns nothing, it falls through to `metadata?.title` (the pre-cleaned MediaSession value). 

Just pushed a commit that expands the surrounding comment so the fallback is explicit from reading the source.

rob9315:
ah i see, i must've missed that viewing the pr on my phone :+1:

jaccarmac:
Using `mediaSession` is by design.

That said, the option we have for YouTube - "Use YouTube Music to get track info" - is slightly confusing IMO and it might be worth it to use that option to toggle YouTube Music behavior as well.

The standard for adding new options is pretty high: They involve a maintenance burden and can be hard to understand for users.

Please open other PR(s) for other connector(s).

> they wouldn't match the canonical track on Last.fm

There is no consistent notion of "canonical track" on Last.fm. We occasionally have discussions about what the best extension behavior is, but we can't just assert a preference in every PR.

> ~5 lines of real code plus an explanatory comment. No new settings, no new deps, no regression. All existing tests pass.

Keep the LLM-ese out of PR submissions, please. This is information I could already get by glancing at the GitHub UI.

src/connectors/youtube-music.ts
```diff
@@ -22,6 +22,22 @@ export {};
...
+ * the scrobble matches the canonical track on the scrobbling service.
+ *
+ * If the DOM selector stops matching (YT Music UI changes, element missing),
+ * `getArtistTrack` falls back to `mediaSession.metadata.title` so scrobbling
+ * still works — just with the pre-cleaned title until the selector is
+ * updated.
+ *
+ * Artist and album still use MediaSession — they're structured fields and
+ * don't get the same suffix-strip treatment.
+ */
```
rob9315:
```suggestion
// track title in `mediaSession` is stripped compared to DOM
```
honestly this should be enough to explain why the line has to exist

jaccarmac:
ref. #4991

jaccarmac:
The `mediaSession` approach was introduced in #4587, and we should test this against its linked bugs.

jaccarmac:
While it doesn't normalize track names, #5785 probably addresses the "orphan single" issue in part, keeping in mind that the fix here sounds like it relies on album guessing.

jonnybel:
I hadn't seen those Issues/PRs, thanks for the pointers!

I was imprecise with "canonical", I meant the arguably more accurate one, since the stripped title doesn't match the last.fm / ListenBrainz-MBID entry that actually has the (feat. X) included, so scrobbles don't associate with the known track and get stranded.

Indeed this might also be compounded by the issue from #5785, so both these fixes together probably mitigate many wrong scrobbles from YT Music. 

I understand that by design it makes more sense to use the MediaSession, since in theory it's much more reliable than parsing DOM, but it's unfortunate that YT Music itself has this inconsistency: the title of the track has one value in the UI, and a different one on the MediaSession. 

That being said, in this case I felt it could still be acceptable since many other "major" connectors also use the DOM-parsing strategy, and especially since with this approach, when YT decides to update the UI again it falls back to MediaSession.

> keeping in mind that the fix here sounds like it relies on album guessing

What do you mean by album guessing? In this PR there are no changes to Connector.getAlbum, only Connector.getArtistTrack, where the DOM title gives `processYtVideoTitle` better input than the stripped MediaSession one and the logic is unchanged.

jonnybel:
> but it's probably worth looking into why the amazon filterset has that

I looked into why Amazon has removeFeature. It was added on purpose in #2193. I don't know if this is still an issue 6 years later but I also don't have Amazon Music to test it, so I'll leave that alone.

jaccarmac:
Thanks for finding the Amazon PR. That was going to be the next thing I peeked at. Personally, I am in favor of removing that special case: I don't really care if "feat." is in the artist and the track title, but that's a taste thing.

> I meant the arguably more accurate one...
> I understand that by design it makes more sense to use the MediaSession, since in theory it's much more reliable than parsing DOM, but it's unfortunate that YT Music itself has this inconsistency: the title of the track has one value in the UI, and a different one on the MediaSession.

My nitpick is that this is true for *some* tracks. Last.fm isn't as strict as MusicBrainz, so while the tracks you care about might scrobble better with this change, other tracks might not. This may be less true for YouTube Music than in the past, but I want to do some testing to see if I find any counterexamples.

We also definitely can't regress the narrow-window fix, though I also need to test that.

> What do you mean by album guessing? In this PR there are no changes to Connector.getAlbum

If album guessing is turned off, the name of the track has no impact on whether it is scrobbled as an orphan single or part of the correct album.

PR #5977, opened 2026-04-21
Add support for tv.theindiebeat.fm

### LINKED ISSUE #5973: Support for tv.theindiebeat.fm
TIBtv is a 24-hour video streaming channel. Like MTV but for indies!  It's quite active and gaining in popularity.

URL is https://tv.theindiebeat.fm

You don't need an account to access.

**Additional context**
If the format of the page isn't conducive to scrobbling, then the devs are lovely folk - I'm sure they'd consider some tweaks to help it be supported.

rob9315:
connector for https://tv.theindiebeat.fm/

fixes #5973

gets the time information from data tags on the #nowplaying element which also contains the artist and title info

### DIFF
--- a/src/connectors/theindiebeat-owncast.ts
+++ b/src/connectors/theindiebeat-owncast.ts
@@ -6,7 +6,7 @@ Connector.isPlaying = () => {
 	const video = document.querySelector('#player video');
 
 	if (!(video instanceof HTMLVideoElement)) {
-		Util.debugLog(`video player not found`, 'error');
+		Util.debugLog('video player not found', 'error');
 		return;
 	}
 
@@ -29,7 +29,7 @@ Connector.getTimeInfo = () => {
 		return null;
 	}
 
-	const { stream_delay, started, duration } = nowplaying.dataset;
+	const { stream_delay: streamDelay, started, duration } = nowplaying.dataset;
 
 	if (!started || !duration) {
 		return null;
@@ -39,7 +39,7 @@ Connector.getTimeInfo = () => {
 	const nowUnixTime = +new Date();
 
 	const currentTime =
-		(nowUnixTime - startedUnixTime) / 1000 - +(stream_delay ?? 0);
+		(nowUnixTime - startedUnixTime) / 1000 - +(streamDelay ?? 0);
 
 	return { currentTime, duration: +duration };
 };
--- a/src/connectors/theindiebeat-owncast.ts
+++ b/src/connectors/theindiebeat-owncast.ts
@@ -1,0 +1,45 @@
+export {};
+
+Connector.playerSelector = ['#player', '#nowplaying'];
+
+Connector.isPlaying = () => {
+	const video = document.querySelector('#player video');
+
+	if (!(video instanceof HTMLVideoElement)) {
+		Util.debugLog(`video player not found`, 'error');
+		return;
+	}
+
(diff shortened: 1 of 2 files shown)


rob9315:
i suspect that https://live.theindiebeat.fm/ is going to be the same or similar, however that hasn't been live for 5 days it says and codeberg where they supposedly host their code is having an outage

inverse (approved):
Thanks for the contribution!

scy:
Well this is a bit disappointing.

For context: I'm the guy who created the "now playing" display on The Indie Beat TV, and one of my main design goals was to have WebScrobbler support TIBtv ASAP, which is why I created #5711 almost four months ago and which is why there are these additional `data-` attributes in the player on the TIBtv site.

It's nice that support for it got merged, thanks @rob9315 for that. However, I see two issues with your implementation here:

1. You should ignore the `streamDelay` value. It's a rough estimate of the delay between when the streaming server starts playing something and when the average client might see it on their screen. However, clients can be faster or slower by several seconds, probably up to half a minute. WebScrobbler does just fine if it only knows the `duration` of the track.
2. I like that you implemented `scrobblingDisallowedReason`, didn't know that's a thing. However, you're assuming that every time the "now playing" element is empty is because an "ad" is playing. That's not the case. First of all, TIBtv plays very little "ads" in the traditional sense; it's a non-commercial project. We mostly play what we call "interstitials" like station or program ID clips ("you are watching the Electronic hour"). Additionally, metadata may be missing or suppressed for a couple more reasons, for example when regular programming is interrupted for a live event. In other words, we can't really know the reason why the "now playing" element is empty, but in most cases it won't be because an ad is playing.

Addressing these two things will probably have the implementation end up looking pretty much like mine, I guess…? Of course it's up to you, the maintainers, to choose how to implement support.

rob9315:
1. That's interesting! streamdelay matched extremely accurately for me which is why i didn't question it any more than that - but if you say this will deviate more than i experienced usually we can obviously switch to your implementation of only the duration. (and for people with their time set wrong i guess - "i was so eager that i could that i didn't stop to question if i should")

2. there are only a few different allowed values for `scrobblingDisallowedReason` ([see here](https://github.com/web-scrobbler/web-scrobbler/blob/bdca7880cc90e006f224180f5c66af897f4b0b0d/src/core/object/disallowed-reason.ts#L4)) currently, they effectively all do the same thing (safe for a bit of [flavor text](https://github.com/web-scrobbler/web-scrobbler/blob/bdca7880cc90e006f224180f5c66af897f4b0b0d/src/_locales/en/messages.json#L690-L729) in the popup implemented [here](https://github.com/web-scrobbler/web-scrobbler/blob/bdca7880cc90e006f224180f5c66af897f4b0b0d/src/ui/popup/disallowed.tsx#L37)). I guess we can switch to [`ElementMissing`](https://github.com/web-scrobbler/web-scrobbler/blob/bdca7880cc90e006f224180f5c66af897f4b0b0d/src/_locales/en/messages.json#L715)? though accurate it implies in the text that the connector is likely broken - then maybe just `Other`? but that doesn't really tell the user anything.. ¯\_(ツ)_/¯ 

thanks for making the now playing bar, it's great!
(out of curiosity, is the Live site (which seems to have the same player) when it's live, also have the bar?)
i think the maintainers were just quite busy with other things for a longer time, it's great that now things are moving more again :)

PR #5987, opened 2026-04-30
Fix KEXP connector

### LINKED ISSUE #5986: Web scrobbling for kexp.org stopped working
This issue started yesterday. When I go to http://kexp.org and press play, it says "This site is supported for scrobbling" but it doesn't actually scrobble. Eventually it switches to "Web Scrobbler does not support this site".

hornbuckle:
Fixed [KEXP](https://kexp.org/) connector. Resolves #5986.

**Additional context**
Thankfully this player is mostly still the same, but they made some updates to the artist/track display area from how it was [before](https://web.archive.org/web/20260415174544/https://www.kexp.org/). Track now comes before artist and is separated by a bullet. They also added some new child elements, but I left the selector the same to hopefully avoid the connector breaking if they add/remove any of them, rather than change it to one of the children.
Also updated the filter to use MetadataFilter, since it's been added there now.

### DIFF
--- a/src/connectors/kexp.ts
+++ b/src/connectors/kexp.ts
@@ -16,7 +16,7 @@ Connector.getArtistTrack = () => {
 
 Connector.albumSelector = '.Player-album';
 
-Connector.trackArtSelector = '.Player-coverImage.image-is-loaded';
+Connector.trackArtSelector = '.Player-coverImage';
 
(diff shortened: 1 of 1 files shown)


inverse (approved):
Thanks for the contribution!

PR #5988, opened 2026-05-01
Add support for musicMe

### LINKED ISSUE #5985: Support scrobbling from musicme.com
[Musicme](https://www.musicme.com/) is a music streaming service used in France. 

Part of the catalog is available for free. Creating an account is not necessary to access the free catalog. Everybody can create a free account to manage their preferences.

A paid service is also available through direct subscriptions, or through **specific access via subscriptions to certain libraries.** For these specific cases, the url of the app can be different, for example the url to access the musime catalog through an [BNF (Bibliotèque Nationale de France)](https://www.bnf.fr) subscription looks like that : https://bnf-bib-musicme-com.bnf.idm.oclc.org

hornbuckle:
Added connector for [musicMe](https://www.musicme.com/). Resolves #5985.

**Additional context**
Artist sometimes contains multiple names separated by a slash, but they're often the songwriter or producer and not the performer. The connector only captures the first name, which seems to often be the performing artist but not always. However, it seemed best to use that instead of inadvertently creating a bunch of incorrect multiple artists on Last.fm. Occasional manual edits may be required when scrobbling from this site.
Request mentions additional URLs, but they are not accessible without having an account with the participating library. Hopefully, those URLs actually resolve back to musicme.com when accessed.
Also, I was seeing some weird behavior with this while testing it, where the extension would just stop registering anything while playing. I'm not sure if it's due to the iframe or Firefox or something else. But as far as I can tell, this is a pretty straightforward player and should also be a simple connector setup as submitted.

One additional note: I added MetadataFilters to this, but they won't all work until the case insensitive edits submitted in https://github.com/web-scrobbler/metadata-filter/pull/394 are merged, as this site uses lowercase text for additional title info. Since that is also my first submission there, please let me know if there is anything I missed or need to update.

### DIFF
--- a/src/connectors/musicme.ts
+++ b/src/connectors/musicme.ts
@@ -1,0 +1,26 @@
+export {};
+
+const player = '#audio-player';
+const filter = MetadataFilter.createFilter({
+	track: [
+		MetadataFilter.removeRemastered,
+		MetadataFilter.removeVersion,
+		MetadataFilter.removeLive,
+	],
+});
+
+Connector.playerSelector = player;
+
+Connector.artistSelector = `${player} .meta-main > a:first-of-type`;
+
+Connector.trackSelector = `${player} .meta-sub`;
+
+Connector.trackArtSelector = `${player} img.cover`;
+
+Connector.currentTimeSelector = `${player} .timing-elapsed`;
+
+Connector.durationSelector = `${player} .timing-total`;
+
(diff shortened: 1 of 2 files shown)


inverse:
Thanks for the contribution!

hornbuckle:
@inverse No problem! (BTW not sure if you can change the labels after merging, but this was a new connector.)

PR #5990, opened 2026-05-01
Scrobble BBC Sounds playlists with longer lists

neanias:

If a playlist on BBC Sounds have greater than 9 tracks, the track selector would report the track as `["10.", "<track>"]` rather than `["<track>", "<artist>"]. This changes the regex to accept 1 or more digits rather than just a single number.

**Additional context**

An example of what happens currently:

Screenshot 2026-05-01 at 13 40 34


inverse (approved):
Thanks for the contribution!

PR #5981, opened 2026-04-26
Poolsuite scrobbling fixed via Media Session API (fixes #5375)

edelstone:
**Describe the changes you made**

Replaced the old Poolsuite DOM-scraping connector with Media Session API support. Poolsuite now exposes track and artist via `navigator.mediaSession.metadata`, so Web Scrobbler can read it directly without brittle selectors.

**Additional context**

Fixes #5375.

Tested on [poolsuite.net](https://poolsuite.net) in Chrome/Brave for Mac. Track and artist update correctly when skipping tracks and scrobble normally.

Apologies for the [duplicate PR](https://github.com/web-scrobbler/web-scrobbler/pull/5779); I deleted my fork before you completed review. This is an identical fix as previously submitted.

Issue #4613, opened 2024-03-26
music.youtube.com connector doesn't work for a few days


NorthernBlow:
**Describe the bug**
music yotube connector works fine until last three days and after upgradee broken

**Environment**
-   OS: Arch Linux
-   Browser: Firefox
-   Extension version: Latest

flavioislima:
I can confirm it doesn't work for me as well for both Linux (Arch and OpenSuse) on Firefox and Vivaldi.
But also doesn't work on Windows as well for same browsers:


Music is playing but still cannot recognize it:
![image](https://github.com/web-scrobbler/web-scrobbler/assets/26871415/575a7403-a132-4256-bb03-2c4c52a5d201)

Tried restarting the browser, reinstalling the extension, logging out and in again from Last.Fm and still nothing works.

Kellen-wq:
for me it works, but adds comma after almost every artist name
![image](https://github.com/web-scrobbler/web-scrobbler/assets/76692731/23d8aa4a-1487-438a-9ba7-f3653814fef9)

TheColorman:
Probably related to #4587. Works fine for me on Windows/Google Chrome + Chrome desktop PWA. The rework made the connector use the javascript api for playing media, is your browser correctly detecting the music?

zbigniewzolnierowicz:
Can confirm the same error on macOS + Firefox.

vpzomtrrfrt:
I see information in `navigator.mediaSession.metadata` while playing so it does seem to exist

NorthernBlow:
on Windows 10 + Firefox same problem, Windows 11 + Firefox same

CECTPAyTkyHECu:
Have the same problem too on Windows 11 and Firefox

abrown19:
Oddly enough, this is working fine for me in Microsoft Edge only (Windows 10 and Windows 11) using the plugin version from the Edge Add-ons site. The other browsers I've tested (firefox, chrome, vivaldi) using the Mozilla or Chrome Web Store versions are reporting errors.

TheColorman:
Maybe related, but I get a "The track is not recognized" notification for every song, but it detects just fine.

1shira:
I did some testing an discovered, that the `navigator` Object that the extension accesses is not the same that the page accesses in Firefox. 

[...]

PR #5991, opened 2026-05-01
TIDAL: Get album from navigator.mediaSession.

jaccarmac:
Unedited scrobbles from playlists now have proper albums.

Album artist detection is still squirrelly since we match URLs and need the album in the playing bar.

Only switch the album for now since, no matter what, the first artist on a track is listed as that track's artist. I may end up eating some of my words on #5972, as only one artist is ever listed. I don't think it's correct to scrobble tracks like that.

### DIFF
--- a/src/connectors/tidal.ts
+++ b/src/connectors/tidal.ts
@@ -34,12 +34,12 @@ Connector.getArtist = () => {
 Connector.albumSelector = `${Connector.playerSelector} a[href^="/album/"]`;
 
 Connector.getAlbum = () => {
-	const {mediaSession} = navigator;
+	const { mediaSession } = navigator;
 	if (mediaSession && mediaSession.metadata) {
 		return mediaSession.metadata.album;
 	}
-	return Util.getTextFromSelectors(Connector.albumSelector)
-}
+	return Util.getTextFromSelectors(Connector.albumSelector);
+};
 
 Connector.getAlbumArtist = () => {
 	const albumUrlSegments = Util.getAttrFromSelectors(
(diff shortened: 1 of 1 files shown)


jaccarmac:
Summoning @rob9315 from #4613.

Maybe I don't fully understand the interactions between various connectors and DOM injection and `navigator.mediaSession`, but I see the current `mediaSession` code for YouTube as slightly tortured compared to a PR like this or #5981.

rob9315:
the YTM mediasession code is solving a problem that should be addressed in the core code
i don't fully understand when the mediasession is available to the extension and when it isn't
on firefox, and i would assume on other browsers as well but i haven't checked, the extension runs in its own javascript context but the same dom. that is also the reason i was having trouble interacting with the youtube player api, which is just a collection of properties on a dom object. they exist in the console, they exist to dom injected code, but they don't exist to the extension
now i don't know when the changes to the mediaSession are replicated for the extension's javascript context, or even what happens when an iframe also plays something. this is all easier to comprehend on chromium based browsers because you can easily manually set the console's context there

inverse (approved):

rob9315:
so, if you don't have an extension that tries to hook mediaSession installed that breaks the object for all extensions, then you should be just fine accessing it from the content script and you should not need any script injection or such

jaccarmac:
That's what I figured, and, frankly, I'm not sure it's worth it to carry the maintenance burden of the injection, duplicated code, and YouTube connector delays to fix other extensions.

jaccarmac:
But could you open an issue to track that with links esp. to the KDE issue you found?

rob9315:
I even made a merge request with them and it seems to have been a known issue.
There does however not seem to be a real solution to do what they are doing there correctly but that is neither here nor there

PR #5992, opened 2026-05-02
chore: fix circular imports

inverse:

- utils was depending on options and options on utils and this change removes that by dropping a debug log in options
- moves the debug logic to its own file for easier management
- exports these in original places to avoid bigger refactoring for now
- Added madge for checking and enforcing via lint
- Fix other instances with utils and action

**Additional context**


rob9315:
nice! 🎉

PR #5995, opened 2026-05-02
restore dev convenience

rob9315:

- fixes file watching
- generates sourcemap files when not running `npm run dist <browser>`

**Additional context**

there are essentially three "build modes" now

| mode | watch | sourcemap | command |
| ---- | ----- | --------- | ------- |
| dev | yes | yes | npm run dev <browser> |
| normal | no | yes | npm run build <browser> |
| prod | no | no | npm run dist <browser> |

i have found that i was missing a source map badly for tracing errors that users are encountering

### DIFF
--- a/scripts/set-env.ts
+++ b/scripts/set-env.ts
@@ -1,8 +1,8 @@
-import { isDev, releaseTarget } from './util';
+import { isDev, isProd, releaseTarget } from './util';
 
 if (isDev()) {
 	process.env.VITE_DEV = 'true';
-} else {
+} else if (isProd()) {
 	process.env.VITE_PROD = 'true';
 }
 
--- a/vite.configs.ts
+++ b/vite.configs.ts
@@ -4,7 +4,7 @@ import solid from 'vite-plugin-solid';
 import { viteStaticCopy } from 'vite-plugin-static-copy';
 import compileConnectors from './scripts/compile-connectors';
 import makeManifest from './scripts/make-manifest';
-import { isDev, releaseTarget, resolvePath } from 'scripts/util';
+import { isDev, isProd, releaseTarget, resolvePath } from 'scripts/util';
 import minifyImages from 'scripts/minify-images';
 import generateIcons from 'scripts/generate-icons';
 import ConditionalCompile from 'vite-plugin-conditional-compiler';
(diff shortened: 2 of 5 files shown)


scripts/set-env.ts
```diff
@@ -1,8 +1,8 @@
-import { isDev, releaseTarget } from './util';
+import { isDev, isProd, releaseTarget } from './util';
 
 if (isDev()) {
 	process.env.VITE_DEV = 'true';
-} else {
+} else if (isProd()) {
```
inverse, line 5: Lets explicitly handle this now if it's neither to raise an error to avoid weird behaviour

vite.configs.ts
```diff
@@ -73,6 +79,7 @@ export const buildContent: UserConfig = {
 			output: {
 				entryFileNames: 'main.js',
 				extend: true,
+				sourcemap: !isProd(),
```
inverse:
```suggestion
				sourcemap: isDev(),
```

Would this make more sense to avoid the negation

vite.configs.ts
```diff
@@ -85,14 +92,16 @@ export const buildStart: UserConfig = {
...
+		watch: isDev() ? watchConfig : null,
 		sourcemap: isDev(),
 		rollupOptions: {
 			input: {
 				popup: resolvePath(root, 'ui', 'popup', 'index.html'),
 				options: resolvePath(root, 'ui', 'options', 'index.html'),
 			},
-			output: {},
+			output: {
+				sourcemap: !isProd(),
```
inverse:
```suggestion
				sourcemap: isDev(),
```

Same for this one

inverse (commented):
Nice one - minor suggestions

rob9315:
i disagree with the changes. dev vs build vs dist are explicitly three separate things with this PR
`isDev` is not the opposite of `isProd`, so that you can do `build` (which is neither dev nor prod) and end up with the same js as the users, but with sourcemaps

vite.configs.ts
```diff
@@ -51,6 +56,7 @@ export const buildBackground: UserConfig = {
 			output: {
 				entryFileNames: 'main.js',
 				extend: true,
+				sourcemap: !isProd(),
```
inverse: Are these correct? When running I get

```
vite v7.3.2 building client environment for production...
Vite does not support "rollupOptions.output.sourcemap". Please use "build.sourcemap" instead.
vite v7.3.2 building client environment for production...
Vite does not support "rollupOptions.output.sourcemap". Please use "build.sourcemap" instead.
vite v7.3.2 building client environment for production...
Vite does not support "rollupOptions.output.sourcemap". Please use "build.sourcemap" instead.
```

inverse:
Ahh I see now - was originally thinking it could get messy mixing these behaviours and relying on negations vs potentially adding a new releaseType that outlines the behaviour explicitly.

vite.configs.ts
rob9315: oh, i did not notice that. it worked for me but let's change that

PR #6009, opened 2026-05-03
Re-add support for Bandcamp embeds

### LINKED ISSUE #2853: Brooklynvegan.com

### LINKED ISSUE #3067: thequietus.com - Scrobbling of bandcamp
thequietus.com is a website dedicated to music, book, films, etc. 
I mostly read the music reviews. Most of the music they present is available via bandcamp. Often they include multiple bandcamp players in one article. It's in a quite similar to bandcamp daily. 
Sometimes they also include Youtube and Spotify links in the same article. But i guess that's just too much. I can imagine that this would be pretty complex to parse correctly. 

Example links: 
Multiple bandcamp players:
https://thequietus.com/articles/30859-converge-chelsea-wolfe-review

One bandcamp player:
https://thequietus.com/articles/30806-portico-quartet-monument-review

LOTS of bandcamp players:
https://thequietus.com/articles/30903-the-quietus-top-100-albums-of-2021-norman-records

**Additional context**

hornbuckle:
Adding support for standard Bandcamp embeds, as suggested and discussed in #5941.
Submitting a new PR to be clear of the intention with this new connector. This will allow the standard Bandcamp embedded player to scrobble from any site, without having to explicitly include the site itself in the connectors list.

**Additional context**
Initially submitted in an earlier PR for a specific site, it was realized that this had bigger potential. Similar to #6004, adding each site that uses a particular embedded player to the connectors is unnecessarily complicated, and with the ubiquitous use of Bandcamp embeds, it's unrealistic and unscalable to even do so. This allows all of those embeds to scrobble without adding multiple sites. So, not only does [poeba.lu](https://poeba.lu/blog/aplacetoburystrangers-rare-and-deadly/) work, but so does [BrooklynVegan](https://www.brooklynvegan.com/american-football-kacey-musgraves-emily-nenni-reviews/) and any other site that has a Bandcamp embed.
One minor issue I'm seeing, which I don't believe should be a blocker and can potentially be resolved in a separate PR, is if multiple players are on a page and one is played then paused and another is played after it, the extension pop-up still shows the track from the first player. But the active track is still being scrobbled as expected. This seems like a fix could be made for the pop-up to update to display the actively playing track. Additional filters may also be needed. Otherwise, this seems to work just fine.
This is very different than what was submitted in #2026 because this has the Bandcamp embedded player URL to match. This would also resolve #2853 and resolve #3067.
Lastly, I updated the Bandcamp Daily connector ID and JS to the more appropriate name of "bandcamp-daily" for clarity and also removed `allFrames` because it is not an embedded player. @rob9315 did some digging and found #2253, which appears to have broken the original "bandcamp-embed" implementation and changed it to work only for Bandcamp Daily instead of creating a separate connector. It appears that this was working as it would be with this update until that happened, since the selectors that were used then are similar. If following the [naming conventions](https://github.com/web-scrobbler/web-scrobbler/wiki/Connectors-development#how-to-name-connector-file), it should have been called "bandcamp-daily" at that point. This connector, reclaiming "bandcamp-embed," is for the default embedded player used across the web.


rob9315:
I would just do the id and js changes tbh, maybe with a patchnote

jaccarmac:
> the extension pop-up still shows the track from the first player

AFAICT, that is a general property of our pop-up and not even specific to embeds. Shouldn't block.

a4180p:
@hornbuckle are there any chances that this PR will be merged?

hornbuckle:
> @hornbuckle are there any chances that this PR will be merged?

Yes, but there just haven't been any new merges in nearly a month.

PR #6017, opened 2026-05-05
Add support for Spotify embeds

### LINKED ISSUE #2024: Add support for any Spotify embeds

hornbuckle:
Adding support for Spotify embeds, including podcast and playlist recognition. This will allow Spotify embedded players to scrobble from any site, without having to explicitly include the site itself in the connectors list.

**Additional context**
Building off of the ideas in #6004 and #6009, this adds support for Spotify embedded players. As noted in the latter PR, there is a minor issue where if multiple players are on a page ([example](https://www.brooklynvegan.com/classic-goths-13-greatest-albums/)) and one is played then paused and another is played after it, the extension pop-up still shows the track from the first player. But the active track is still being scrobbled as expected. This seems like something that could potentially be fixed in a separate PR.
Additionally, here is a [standard playlist](https://www.brooklynvegan.com/our-favorite-songs-of-the-week-playlist-150/) embed and an [artist playlist](https://superdragtour.com/#spotify) embed to see the difference in behavior from an album.
Resolves #2024.

### DIFF
--- a/src/connectors/spotify-embed.ts
+++ b/src/connectors/spotify-embed.ts
@@ -1,0 +1,75 @@
+export {};
+
+Connector.playerSelector = '[data-testid=embed-widget-container]';
+
+Connector.getArtist = () => {
+	const tracklistArtistSelector =
+		'li[class*=TracklistRow_isCurrentTrack__] h4';
+	const tracklistArtistElement = document.querySelector(
+		tracklistArtistSelector,
+	);
+	const trackArtistElements = document.querySelectorAll(
+		'[class*=_metadataWrapper__] h2 a',
+	);
+
+	if (tracklistArtistElement) {
+		return tracklistArtistElement?.lastChild?.textContent;
+	}
+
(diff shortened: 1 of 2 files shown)


rob9315:
You're on a roll with the embeds! 🎉

inverse (approved):

PR #6019, opened 2026-05-05
Add support for The Lot Radio

cubeghost:
Added connector for [The Lot Radio](https://www.thelotradio.com/)

**Additional context**
The Lot Radio's markup is unfortunately not semantic at all, so the selectors used here depend on their Tailwind classes and are quite brittle. I also couldn't find a way to re-initialize on navigation change (they use `history.pushState` type navigation), which is why the `isMissingTracklist` check happens inside `getTrack`, `getArtist`, and `getDuration`.


inverse (approved):

inverse:
Thanks for the contribution!

PR #6028, opened 2026-05-09
add privacy url fallback

### LINKED ISSUE #6027: Privacy Policy does not forward to correct address on Firefox
On Firefox when the user clicks on Extension on the top of browser
Goes into Settings
Goes into Contact
Clicks on Privacy Policy
It forwards to Show Us Some Love page

**Expected behavior**
Linking to the correct policy

-   OS: Windows 10
-   Browser: Firefox
-   Extension version: 3.18.0

rob9315:

add a fallback for when the privacy url has not been translated, to open the english one as a text file.
not the prettiest but i don't know how to make a proper ui for this

**Additional context**

fixes #6027

<details><summary>alternative?</summary>

alternatively maybe we should use this?
https://github.com/web-scrobbler/web-scrobbler/blob/ea09c2dfb8b46a9ba473b3e4b108b7cfed6aa46b/src/util/util-browser.ts#L51
but it's async and that sucks for ui, someone else would have to do that but in that case i'd suggest that someone who knows web technologies better make a builtin ui which actually renders the markdown? (or link to the exact commit on github?)
</details>

### DIFF
--- a/src/ui/options/components/contact.tsx
+++ b/src/ui/options/components/contact.tsx
@@ -61,7 +61,10 @@ export default function ContactComponent() {
 					</li>
 				</a>
 				<a
-					href={t('contactPrivacyPolicyUrl')}
+					href={
+						t('contactPrivacyPolicyUrl') ||
+						'/_locales/en/privacy.md'
+					}
 					target="_blank"
 					rel="noopener noreferrer"
 				>


pedenys:
> i'd suggest that someone who knows web technologies better make a builtin ui which actually renders the markdown

I can take the task if you want.

src/ui/options/components/contact.tsx
```diff
@@ -61,7 +61,10 @@ export default function ContactComponent() {
 					</li>
 				</a>
 				<a
-					href={t('contactPrivacyPolicyUrl')}
+					href={
+						t('contactPrivacyPolicyUrl') ||
+						'/_locales/en/privacy.md'
```
pedenys, line 66: `browser.i18n.getUILanguage()` can provide the UI language of the browser in a synchronous way. It could be used to link to the appropriate translation of the 
`privacy.md` or fallback to english.

pedenys (approved):

src/ui/options/components/contact.tsx
rob9315: something like that is already being used in this function over here which i mentioned in the PR:
https://github.com/web-scrobbler/web-scrobbler/blob/ea09c2dfb8b46a9ba473b3e4b108b7cfed6aa46b/src/util/util-browser.ts#L51-L69

but it's async because it has to check whether the file exists

PR #6035, opened 2026-05-11
fix globalplayer

### LINKED ISSUE #2244: Radio player
Tried to scrobble radioplayer e.g https://www.classicfm.com/radio/player/ and getting the "play some music" prompt even though the API does tell you which track is playing.

rob9315:

it was using a page element that is only there when you're focused on the specific station but you could navigate away. now it uses only info from the playerbar.

**Additional context**

<details><summary>checks for element count in the info panel to detect moderators speaking</summary>
image
</details>

<details><summary>working outside of radio page, with only info from the playbar</summary>
image
</details>

fixes #2244


inverse (commented):
Thanks for the contribution!

PR #6041, opened 2026-05-12
Add support for Mixcloud embeds

### LINKED ISSUE #5959: Sohoradio.com Mixcloud
Sohoradio.com is an online radio site - it doesn't provide live tracklistings, but playing an archived programme uses Mixcloud, and the track details are shown, but not scrobbled.

For example: https://sohoradio.com/show/insomniac-monthly-with-jon-more-17-03-2026/ (press the Mixcloud play icon, rather than the main one at the top of the page)

**Additional context**

Image

hornbuckle:
Adding support for Mixcloud embeds. This will allow Mixcloud embedded players to scrobble from any site, without having to explicitly include the site itself in the connectors list.

**Additional context**
Supports Mixcloud streams with playlist metadata and without; the latter will be considered a podcast.
Example with playlist: https://sohoradio.com/show/insomniac-monthly-with-jon-more-14-04-2026/
Example without playlist: https://sohoradio.com/show/soph-nathan-and-mina-wood-09-02-2026/
Example with multiple embeds of both types: https://arts.britishcouncil.org/selector-radio/highlights
Logic and selectors may be a bit fragile due to use of randomly generated classes and lots of nested divs and spans, but this is the best I could do with what their player provides. Unfortunately, not all embedded players have the extra child element I was initially using to differentiate between players with and without playlists, so, like the main Mixcloud connector, there may be a temporary switch between podcast detection on some shows.
Resolves #5959.

### DIFF
--- a/src/connectors/mixcloud-embed.ts
+++ b/src/connectors/mixcloud-embed.ts
@@ -1,0 +1,58 @@
+export {};
+
+const playerSelector = '[data-testid=widgetTopClass]';
+const playButtonSelector = `${playerSelector} [data-testid=widget-play-button]`;
+const trackSelector = `${playerSelector} span strong:has(+ span)`;
+const artistSelector = `${trackSelector} + span span:last-of-type`;
+
+Connector.playerSelector = playerSelector;
+
+Connector.isPodcast = () => {
+	const playerWrap = document.querySelector(playerSelector);
+
+	return Boolean(playerWrap && playerWrap?.childElementCount < 2);
+};
+
+Connector.getTrackInfo = () => {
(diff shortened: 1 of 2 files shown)


rob9315:
multiple embeds seems to run into issues with the core code not distinguishing between the different embeds as their own "instances" or smth like that. (play one, pause it, play a different one)
i think we merge this anyways, since that seems like more of a core issue?

hornbuckle:
Yes, unfortunately, there has been a similar issue with all of the embedded connectors, but I believe the correct songs still get scrobbled even if the active song isn't always shown in the extension.

PR #5801, opened 2026-01-28
Add support for Subvert.fm

AramZS:
**Describe the changes you made**

This adds the correct tracking to capture song information for scrobbles on the new Subvert.fm co-op bandcamp replacement platform. 

**Additional context**

Apologies that this is basically impossible to test for anyone who is not a member, since the player is behind a paywall. I have tested on Firefox and Chrome successfully and it is working as intended. 

I have run `npm run prettierfix`

I have confirmed everything is working locally across multiple pages and songs.

PR #6044, opened 2026-05-12
Fix Subvert connector

hornbuckle:
Subvert connector previously added in #5801, but connector setup was overly complicated and appears to have already broken before being released. This update is a more simplified connector that only utilizes selectors and has no convoluted logic.

The site is no longer behind a paywall (it seems to have officially opened [today](https://bsky.app/profile/subvert.fm/post/3mlnzh2kqbk22)), and music can now be accessed at https://www.subvert.fm/discover

**Additional context**
Also updated name to just "Subvert," as that is how it is referred to in their [FAQ](https://www.subvert.fm/docs/faq) and elsewhere. And only one URL match is needed.

### DIFF
--- a/src/connectors/subvert.ts
+++ b/src/connectors/subvert.ts
@@ -1,66 +1,17 @@
 export {};
 
-const player = '[class*=globalPlayerContainer]';
-const artistSelector = `${player} .globalPlayerCoverImage > div:nth-child(2) > a:nth-child(2)`;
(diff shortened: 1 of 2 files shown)


rob9315:
fixes #6059

AramZS:
This looks good to me, one of the explicit support requests I made to the Subvert engineers was to make clearer div labels and classes to hook stuff like this on, and they have done so, which enables this.

seathebeauty:
I can't wait to start scobbling my Subvert listens!

PR #6052, opened 2026-05-19
fix Indie88 connector

rob9315:

indie88 has changed their URL for the web-player. both of the old supported styles didn't work anymore with the websites listed.
I've replaced it with a single new connector for the updated main url.
There was a second service listed but they don't have the Track Info anywhere on the page so i didn't bother preserving it

**Additional context**

not really, no

### DIFF
--- a/src/connectors/indie88.ts
+++ b/src/connectors/indie88.ts
@@ -1,31 +1,12 @@
+// https://www.indie88.com/player/
+// managed with https://www.socastdigital.com/
+
 export {};
 
-function setupConnector() {
-	if (isCobrPlayer()) {
-		setupCobrPlayer();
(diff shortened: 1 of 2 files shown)


rob9315:
they changed the url again, to https://indie88.leanplayer.com/CINDFM now and their homepage is a 403 on all paths

rob9315:
leanplayer is something else, from https://www.leanstream.net/
there are others using this, for example [MBS Radio](https://www.mbsradio.com/stations) (taken from the leanstream site), one of their webplayer urls is https://player.fx1019.ca/

rob9315:
gonna keep this PR around for if they bring back their other player, for now

PR #6054, opened 2026-05-19
fix allclassical.org

rob9315:

fixed allclassical.org scrobblingDisallowedReason

**Additional context**

every song was registering as not allowed to play because of 'Other' because the now playing text was now stylized as "NOW PLAYING" when we were checking for "Now Playing". an effect of using .innerText instead of .textContent in the getTextForSelector function. this instead checks whether the `#song-info` element is currently hidden or not (it gets switched out for the `#recently-played` element if the host is talking at the moment

### DIFF
--- a/src/connectors/allclassical.org.ts
+++ b/src/connectors/allclassical.org.ts
@@ -9,6 +9,4 @@ Connector.artistSelector = '#composer';
 Connector.playButtonSelector = '#live-play';
 
 Connector.scrobblingDisallowedReason = () =>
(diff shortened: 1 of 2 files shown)


rob9315:
<details><summary>scrobbling when "NOW PLAYING" (screenshot)</summary>
image
</details>
<details><summary>not scrobbling when "RECENTLY PLAYED" (no screenshot)</summary>
trust me, i think it works
</details>

PR #6055, opened 2026-05-19
Fix scrobbling issues on ytm

Rhylanscript:
Fix issue where songs that have multiple artists without anchor tags do not scrobble correctly. Will only scrobble first artist now.

**Issue**
In the case that a song being scrobbled on youtube music had multiple artists, none of which had an anchor tag, the song wouldn't scrobble correctly. Below is a comparison of the correct scrobble (done with the changes) (first) and the incorrect scrobble (second).

The fix included instead of just looking for anchor tags, it would firstly try for anchor tags, however if not found would resort to common splitters, and would take the first artist to scrobble the song correctly.

This problem of dual instances for albums and artists was present too, and was also fixed with this update.

**Additional context**

Correct scrobble:
image
Incorrect scrobble:
image

### DIFF
--- a/src/connectors/youtube-music.ts
+++ b/src/connectors/youtube-music.ts
@@ -56,18 +56,38 @@ Connector.getTrackArt = () => {
 	return artworks?.[artworks.length - 1].src;
 };
 
+// BETTER ARTIST PARSER - FIXES MISSING ANCHOR TAGS FOR ARTISTS
+function getCleanArtist(rawArtist: string | undefined): string {
+	if (!rawArtist) return '';
+
+	// trim leading / trailing spaces of an artist name, and replace misc whitespace
+	// 	   (e.g. double spaces) with a single space
+	let artist = rawArtist.trim()
+		.replace(/\s+/g, ' ')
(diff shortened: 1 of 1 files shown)


rob9315:
I don't exactly see why this is needed and why you're not doing it with `MetadataFilter`s
I feel like it may want to be behind an option? but wait on adding that for some feedback from actual maintainers
since you seem to intend to always run the cleaning, run it after the if tree when artist has been set

Rhylanscript:
> I don't exactly see why this is needed and why you're not doing it with `MetadataFilter`s I feel like it may want to be behind an option? but wait on adding that for some feedback from actual maintainers since you seem to intend to always run the cleaning, run it after the if tree when artist has been set

It's really just a very minor change for youtube music users. I know some others who have asked for this feature. As for making it an option, I didn't see a reason to have this disabled so I felt making it an option would be somewhat pointless. But if people would choose to turn it off I suppose it wouldn't be too difficult to add.

jaccarmac:
Some people prefer the first artist, some people prefer scrobbling multiple artists; We definitely can't hardcode it like this.

PR #6025, opened 2026-05-09
[yandex] complete overhaul

rob9315:
**Describe the changes you made**

i switched to some more sensible selectors and made the code less ai-generated.

**Additional context**

old layout removed, doesn't seem to exist anymore
when loading the website it has a fake playbar that gets removed later. in order to mitigate the playerSelector falsely selecting that element for observing which gets later removed, the playerselector filters for only a playbar with track text. there's going to be a warning in the console because of that but it works.

close #5832: this PR removes reliance on language
close #5624: ai PR that also would have fixed the issues probably but is way too convoluted

fixes #5702: "it don't work"
fixes #5506: new/old design
fixes #5233: not working since ...
fixes #5134: not working on new and old site

<details><summary>handles variations/subtitle:</summary>
image
</details>

fixes: #4911: subtitle
fixes: #4530: subtitle

*please someone test this*

PR #6060, opened 2026-05-20
Fix Yandex Music connector for new player layouts

Alexsab:

Restore Yandex Music scrobbling after the latest player DOM changes (2026-05-20)

The connector now detects the active Yandex Music player layout instead of relying on theme-based detection and a single hardcoded player selector. It supports the current Vibe player, mobile PlayerBar, desktop background-progress PlayerBar, current collapsed PlayerBar, and keeps the legacy `.player-controls` layout as a fallback.

The connector re-detects the active layout during SPA navigation, so switching between Vibe and regular pages can update the metadata getters. PlayerBar metadata reads are scoped to the active player root to avoid picking up unrelated page content.

**Additional context**

Vibe stores some metadata outside the player bar, so it has a dedicated path for artist/title extraction. It also handles the idle preview title format shown as `Artist — Track` before `SeparatedArtists` is rendered, avoiding `"Моя волна"` being returned as the artist.

**Manual testing completed:**

- Opened Yandex Music home page / Vibe ("Моя волна"), started playback, and verified artist and track are detected.
- Used Next/Prev on the Vibe page and verified metadata updates.
- Navigated from Vibe to a regular Yandex Music page and verified the regular PlayerBar is detected.
- Navigated back to Vibe and verified scrobbling resumes.
- Switched between Vibe and regular pages several times and verified scrobbling continues to update correctly.
- Verified mobile `PlayerBarMobile` layout.
- Verified desktop `PlayerBarDesktopWithBackgroundProgressBar` layout.

<details><summary>Current Vibe layout</summary>
Yandex Music new vibe layout
</details>

Related to #6025 which addressed the same Yandex Music breakage. This change expands the fix to the current player layouts, including Vibe and SPA navigation between layouts

### DIFF
--- a/src/connectors/yandex-music.ts
+++ b/src/connectors/yandex-music.ts
@@ -2,66 +2,131 @@ export {};
 
 const currentCollapsedPlayerSelector =
 	'section[class*="PlayerBar_root__"]:has([class*="Meta_titleContainer__"])';
+const vibePageSelector = 'div[class*="VibePage_meta__"]';
 const vibePlayerSelector = 'section[class*="VibePlayerBar_root__"]';
+const mobilePlayerSelector = 'section[class*="PlayerBarMobile_root__"]';
 const backgroundProgressPlayerSelector =
 	"section[class*='PlayerBarDesktopWithBackgroundProgressBar_root']";
 const veryOldPlayButtonSelector = '.player-controls__btn_play';
+const stateObserverKeys = new Set<string>();
+
+type ConnectorLayout =
+	| 'current-collapsed'
+	| 'vibe'
+	| 'mobile-player'
+	| 'background-progress'
+	| 'very-old';
+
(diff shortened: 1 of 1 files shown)


rob9315:
I don't have a premium account to test with, that's why i didn't know about the different vibe layout.

Is there a reason you are keeping the very-old layout?
As far as I am aware it has been removed

I assume you want to detect state changes because of the vibe layout?
In my PR i removed this completely because only mobile and "new" layout actually incur a new element, full-screen retains the playerbar in the background.

this pr creates a bunch of extra tech-debt seemingly, and even though you say your changes are expanding on the fix, your PR isn't based on mine..?
I don't like that this creates a bunch of new code where it could probably be done way more cleanly. You set up both a timer and an observer with a bunch of management logic attached.

PR #6062, opened 2026-05-21
switch to Jest

rob9315:

this PR would switch the repo from vitest to jest

i had massive problems with vitest failing in tests when working on the storage abstraction code. there is a slight chance that there is an actual timing problem here but i doubt it, It seems to be because of the hoisting that vitest is doing in conjunction with vite

**considerations**

\+ might be more reliable? (yet to be tested on github runners)
\+ line numbers work correctly on test failure
\- it's slightly slower (1.8s vs 1.3s for me locally)
\- not integrated with vite plugins
\+ jsdom mocking (location)
\- significantly more dependencies

might be worth it to switch over to if we change away from vite, should we change away from vite? idk.


pedenys:
As the Javascript ecosystem is finally moving to ES modules, I think it’s safest to keep using Vitest and fix the existing tests instead of falling back to Jest, which is known to have poor support of ESM (please correct me if I’m wrong). Vitest is also becoming the _de facto_ test framework as npm metrics seem to suggest.

PR #6081, opened 2026-06-03
Fix Poolsuite metadata parsing for most common track formats

edelstone:
## Describe the changes you made

Updates the Poolsuite connector to correctly parse artist and track metadata across the three most common track formats the site uses.

### 1. Hyphenated titles: Splits on the dash using `Util.processSoundCloudTrack`

poolsuite-hyphenated
scrobbler-hyphenated

### 2. Standard titles: Falls back to `MediaSession` artist and title fields directly

poolsuite-standard
scrobbler-standard

### 3. Live station a.k.a. "Poolsuite ON AIR": `MediaSession` contains no useful track data, so artist/track is scraped from the DOM

poolsuite-live
scrobbler-live

## Additional context

Poolsuite was broken entirely until my earlier PR #5981 (released in v3.22.0), which restored playback detection using `useMediaSessionApi()`. However, that approach mapped metadata sub-optimally. The full `Artist - Track` string was coming in as the track title, and the Soundcloud username came in as the artist.

This PR replaces `useMediaSessionApi()` with a custom `getArtistTrack` implementation that handles the three primary formats Poolsuite uses, leveraging the existing `Util.processSoundCloudTrack` utility since Poolsuite streams SoundCloud content.

### DIFF
--- a/src/connectors/poolsuite.ts
+++ b/src/connectors/poolsuite.ts
@@ -1,5 +1,18 @@
 export {};
 
 Connector.playerSelector = 'body';
-Connector.useMediaSessionApi();
 Connector.useTabAudibleApi();
+
+Connector.getArtistTrack = () => {
+  const title = navigator.mediaSession.metadata?.title ?? '';
+  const artist = navigator.mediaSession.metadata?.artist ?? '';
+
+  if (title === 'Poolsuite ON AIR') {
+    const text = document.querySelector('h2.font-everyday')?.textContent ?? '';
+    return Util.processSoundCloudTrack(text);
+  }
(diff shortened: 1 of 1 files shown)


edelstone:
Worth noting: `processSoundCloudTrack` uses a greedy regex that splits on the *last* dash rather than the first.

So, for example, a track like `Tony Allen – Road Close – Dance Dub` results in `Tony Allen – Road Close` being scrobbled as the artist. This may affect the SoundCloud connector as well, and might be worth a separate fix.

inverse (approved):

PR #6096, opened 2026-06-09
Fix Mixcloud connector

### LINKED ISSUE #6095: Mixcloud, existing connector? not scrobbling
Hello. I'm discovering Selector Radio, and while their own page does not have a connector, their shows are also available on Mixcloud. But the Mixcloud connector is not picking up the scrobble data. So maybe this is a bug/fix?

https://www.mixcloud.com/SelectorRadio/playlists/the-selector/

In the example screenshot, the current track shows "Chalk by Antony Szmierek" in the player bar at the bottom.

Image

hornbuckle:
An update to the Mixcloud connector (#6045) just went out in the last release, but apparently they have since removed the classes that were used to capture the artist and track information when provided. As a result, the connector is unable to get that info and considers everything playing on Mixcloud to be a podcast.
Updated the selectors for artist/track to get to the correct nested divs without classes.

**Additional context**
Resolves #6095. Here is the show they were trying to play: https://www.mixcloud.com/SelectorRadio/earth-day-2026-shivani-sen-papaoul-baobei-1-may-2026/

### DIFF
--- a/src/connectors/mixcloud.ts
+++ b/src/connectors/mixcloud.ts
@@ -1,8 +1,8 @@
 export {};
 
 const playerSelector = '[data-testid=player-container]';
-const artistSelector = '[class*=styles__Artist-]';
-const trackSelector = '[class*=styles__Track-]';
+const artistSelector = `${playerSelector} p + div > div:last-of-type > div:last-of-type`;
+const trackSelector = `${playerSelector} p + div > div:last-of-type > div:first-of-type`;
 const filter = MetadataFilter.createFilter({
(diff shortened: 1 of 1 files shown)


rob9315:
Navigating the page to a live stream destroys the playerbar and thus breaks the `MutationObserver`, resultig in no further updates being registered -> same song forever, play/pause do nothing.
Works perfectly otherwise, will merge as is :+1:

hornbuckle:
Ah, thanks! May need to update with a different `playerSelector`. I'll look into it again soon.

PR #6102, opened 2026-06-14
Add support for Lyda

loudar:
Added a connector for [Lyda](https://lyda.app), a music streaming and buying service.


loudar:
Would this be possible to merge?

PR #6133, opened 2026-07-14
Fix Meloradio connector

kubabisi:
Updated selectors to select song info instead of the audition name.

**Additional context**
Previous selector was too generic and was applying for both the on air audition name and the current song, because of that audition name was being chosen as it is first.

screenshot


rob9315:
LGTM!

PR #6141, opened 2026-07-20
Fix Radio Nova connector against the current player

### LINKED ISSUE #6140: Radio Nova connector broken: site replaced the jPlayer player
**Describe the bug**

The Radio Nova connector (`nova.ts`) loads, then never scrobbles.

It still targets the jPlayer-based player that `nova.fr` dropped. Every selector resolves to nothing on the current site:

| Selector                               | Purpose          | Status on current site |
|----------------------------------------|------------------|------------------------|
| `#jp_container_1` + `jp-state-playing` | playback state   | element does not exist |
| `.player`                              | player container | does not exist         |
| `.player .artiste`                     | artist           | does not exist         |
| `.player .titre`                       | track            | does not exist         |

`Connector.isPlaying()` looks for `jp-state-playing` on a `#jp_container_1` that the site no longer renders, so it returns `false` at all times. The extension loads the connector, sets up the observer, concludes that nothing is playing, and submits no track. It raises no error, so the failure passes unnoticed.

**How to reproduce**

1. Go to https://www.nova.fr/radios/radio-nova/
2. Start playback
3. Open the extension debug logs
4. Observe `Successfully loaded Radio Nova connector` and `Setting up observer`, but no now-playing and no scrobble, whatever the playback duration

**Expected behavior**

While a Nova stream plays, the connector reports the broadcast track as now-playing and scrobbles it once the threshold is reached.

**Debug logs**

```
Web Scrobbler: Successfully loaded Radio Nova connector
Web Scrobbler: Controller: Created controller for Radio Nova connector
Web Scrobbler: Setting up observer
```

Nothing follows: no state change, no now-playing, no scrobble.

**Environment**

- OS: Linux
- Browser: Firefox
- Extension version: 3.22.0

**Additional context**

The site runs a Dailymotion-backed player (`audio_player.js`) and puts **no `<audio>` or `<video>` element in the DOM**, so a media element cannot supply playback state. The DOM controls have to.

Those controls expose what the connector needs, under stable ids:

- Playback state: the player flips the `display` of two images. During playback, `#player-play-btn` is `display: none` and `#player-pause-btn` is visible. Stopped, the two swap.
- Metadata: `.current-track.title` carries the track as `ARTIST - TITLE`. `#player-title` and `#player-author` survive in the markup but stay empty, so they are dead ends.

A fix along the lines of `radiofrance.ts`, which also reads playback state off the controls instead of a media element:

```ts
export {};

Connector.playerSelector = '#audio_player';

Connector.isPlaying = () => Util.isElementVisible('#player-pause-btn');

Connector.getArtistTrack = () =>
	Util.splitArtistTrack(Util.getTextFromSelectors('.current-track.title'));
```

The default separator list already covers the ` - ` the site uses, so `splitArtistTrack` needs no explicit separator.

I checked this against the live site with playback started: `isPlaying` resolves to `true`, and splitting `.current-track.title` yields the same artist and track that the site's own metadata feed reports at `https://www.nova.fr/radios-data/www.nova.fr/all.json`, including on a collaboration title where the artist field itself contains punctuation.

I left `getTrackArt` out. `#player-image` holds a plain `` rather than a `background-image`, and it keeps showing the station logo during playback instead of per-track art.

All Nova streams (`radio-nova`, `nova-hip-hop`, and the rest) share this player, so one fix covers the whole `*://www.nova.fr/*` match.

I can open a PR with this fix.

mabhub:

Fixes #6140

Rewrites the Radio Nova connector against the player `nova.fr` runs today. 
The old selectors targeted a jPlayer-based player the site dropped, so `isPlaying()` returned `false` at all times and the connector submitted no track.

- `playerSelector` now points at `#audio_player`, the current player container.
- `isPlaying` reads the visibility of `#player-pause-btn`.  
  The player flips the `display` of a play and a pause image, so the pause image shows exactly while a stream plays.
- `getArtistTrack` splits `.current-track.title`, which holds the track as `ARTIST - TITLE`.  
  `defaultSeparators` already covers that ` - `, so the call passes no separator.
- Drops `artistSelector` and `trackSelector`, whose nodes no longer exist.

**Additional context**

The site runs a Dailymotion-backed player and puts no `<audio>` or `<video>` element in the DOM, so no media element can report playback state. `radiofrance.ts` solves the same problem by reading the controls, and this follows it.

I ran a local Firefox build with this connector on top of the DOM checks in #6140.
The extension reports now-playing and scrobbles, neither of which it did before.

I left `getTrackArt` out. `#player-image` holds a plain `` rather than a `background-image`, and it keeps showing the station logo during playback instead of per-track art.

All Nova streams (`radio-nova`, `nova-hip-hop`, and the rest) share this player, so this covers the whole `*://www.nova.fr/*` match.

### DIFF
--- a/src/connectors/nova.ts
+++ b/src/connectors/nova.ts
@@ -1,10 +1,8 @@
 export {};
 
-Connector.playerSelector = '.player';
+Connector.playerSelector = '#audio_player';
 
-Connector.artistSelector = `${Connector.playerSelector} .artiste`;
+Connector.isPlaying = () => Util.isElementVisible('#player-pause-btn');
 
-Connector.trackSelector = `${Connector.playerSelector} .titre`;
-
-Connector.isPlaying = () =>
-	Util.hasElementClass('#jp_container_1', 'jp-state-playing');
+Connector.getArtistTrack = () =>
+	Util.splitArtistTrack(Util.getTextFromSelectors('.current-track.title'));


rob9315:
Thank you for the great PR! Works perfectly :+1:

PR #6146, opened 2026-07-22
fix archive.org connector (#5984)

### LINKED ISSUE #5984: web scrobbler no longer scrobbles audio from archive.org
Scrobbling with music on archive.org doesn't work anymore. It used to do that. 

yeah

sorusari343:
- Fixed the Internet Archive connector by reading from the jwplayer object instead of piercing the shadow DOM and reading from the DOM
- Added album artist support
- Added an example link including both item author and track author
- Removed unusable example links (audio unavailable / item unavailable)

fixes #5984

### DIFF
--- a/src/connectors/archive-dom-inject.ts
+++ b/src/connectors/archive-dom-inject.ts
@@ -26,9 +26,9 @@ interface JwplayerApi {
 }
 
 type JwplayerPlaylistItem = {
-	title: string,
-	artist: string,
-}
+	title: string;
+	artist: string;
+};
 
 if ('cleanup' in window && typeof window.cleanup === 'function') {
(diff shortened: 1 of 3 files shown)


sorusari343:
It seems that jwplayer is not initialized until a webpage is viewed. This causes the script to fail to initialize in background tabs.

> Uncaught TypeError: jwplayer().on is not a function

I am not sure how to handle this.

rob9315:
I renamed the `archive-dom-inject` to `jwplayer-dom-inject` because there are some other sites we currently have with separate connectors that use jwplayer elements as well.

and some typescript stuff for reusability / cleanup

Thank you for your contribution! :+1:

PR #6151, opened 2026-07-24
fix(listen.moe): update connector for redesigned site layout

fumireko:
The listen.moe website has been redesigned and the old selectors no longer work.

### Changes
- **playerSelector**: changed from #app to .glass
- **artistSelector**: changed from .player-song-artist to .glass .text-xs (artist now uses 	ext-xs class, while the source link uses 	ext-sm)
- **getTrack**: custom function that extracts only the direct text node, ignoring the nested source link inside the title
- **pauseButtonSelector**: replaces the old isPlaying that relied on the now-removed #audio-player element

### Testing
- Verified the connector loads without type errors
- Track name no longer includes the source text
- Artist is correctly extracted

### DIFF
--- a/src/connectors/listen.moe.ts
+++ b/src/connectors/listen.moe.ts
@@ -13,16 +13,29 @@ const filterRules = [
 	{ source: /\s+/g, target: ' ' },
 ];
 
-Connector.playerSelector = '#app';
+Connector.playerSelector = '.glass';
 
-Connector.artistSelector = '.player-song-artist';
+Connector.artistSelector = '.glass .text-xs';
 
-Connector.trackSelector = '.player-song-title';
-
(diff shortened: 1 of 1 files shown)


rob9315:
only gripe in testing is that it doesn't cope with page navigation

the other tabs / pages of the pwa have a footer instead of the card in the center.
the footer version isn't recognized because of different selectors, and navigating back to the main page it's still broken because the playerSelector can't be reacquired, tho this could be fixed in core instead and is a problem on many pages, not only this one

your call, if it's easy to add the selectors for the footer in addition to the current ones and change the player to probably just listen on the entire main page content, or not. if not then this can also just be merged as is, with known limitations

rob9315:
merging with known limitations

PR #6156, opened 2026-07-28
Fix qobuz.ts Connector

CalendricalHeresy:

They updated the app so that the track title is nestled in a few more divs, so the class name of the TrackSelector changed.

Also apparently the .player element gets thrown away when you select a new track on the homepage, so moving the playerSelector up one element fixed it. 

Tested on Edge and working. I of course do not know if this new player is a slow rollout and some people might still have the old player? I suspect not though. 

**Additional context**

### DIFF
--- a/src/connectors/qobuz.ts
+++ b/src/connectors/qobuz.ts
@@ -5,7 +5,7 @@ const timeInfoSelector = '.player__track-time-content';
 
 Connector.playerSelector = '.player';
 
-Connector.trackSelector = '.player__track-title';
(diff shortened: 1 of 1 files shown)


CalendricalHeresy:
Okay I tested a different case and it still doesnt scrobble properly. I give up.

CalendricalHeresy:
For anyone interested: I seem to have fixed it, but by treating most of the DOM as the player: Connector.playerSelector = '.ui-responsive-window';

that works and fixes the problems without changing anything else

PR #6171, opened 2026-08-08
Add support for Outra Hora

alvarovotan:
Adds a connector for https://radio.outrahora.com, an independent Brazilian online radio. The site exposes MediaMetadata via \
avigator.mediaSession\, so the connector points at the player container and the artwork element and delegates metadata extraction to the MediaSession API.

### DIFF
--- a/src/connectors/outrahora.ts
+++ b/src/connectors/outrahora.ts
@@ -1,7 +1,7 @@
 export {};
 
 /**
- * Outra Hora ships two players that both fill the Media Session API:
+ * outrahora rádio ships two players that both fill the Media Session API:
  * the station site at radio.outrahora.com, and an embed bar injected on
  * every page of outrahora.com.
  *
--- a/src/connectors/outrahora.ts
+++ b/src/connectors/outrahora.ts
@@ -1,13 +1,32 @@
 export {};
 
-Connector.playerSelector = '#playerView';
+/**
+ * Outra Hora ships two players that both fill the Media Session API:
+ * the station site at radio.outrahora.com, and an embed bar injected on
+ * every page of outrahora.com.
+ *
+ * The bar keeps its player inside a Shadow DOM and creates the audio
+ * element with `new Audio()`, so it is reachable neither by
+ * `document.querySelector('audio')` nor through the shadow root. Both
+ * the audio listeners and the `paused` check only apply to the station
+ * site; on the bar we fall back to the tab audible API.
+ */
+const isEmbedBar = window.location.hostname !== 'radio.outrahora.com';
 
-Util.bindListeners(
-	['audio'],
-	['playing', 'pause', 'waiting'],
-	Connector.onStateChanged,
-);
+if (isEmbedBar) {
+	Connector.playerSelector = '#outrahora-radio-widget';
 
-Connector.isPlaying = () => !document.querySelector('audio')?.paused;
+	Connector.useTabAudibleApi();
+} else {
+	Connector.playerSelector = '#playerView';
+
+	Util.bindListeners(
+		['audio'],
+		['playing', 'pause', 'waiting'],
+		Connector.onStateChanged,
+	);
+
+	Connector.isPlaying = () => !document.querySelector('audio')?.paused;
+}
 
 Connector.useMediaSessionApi();
--- a/src/connectors/outrahora.ts
+++ b/src/connectors/outrahora.ts
@@ -1,32 +1,12 @@
 export {};
 
-/**
- * Connector for Rádio Outra Hora (https://radio.outrahora.com).
- *
- * The site is a static SPA that exposes a single primary player
- * (`#playerView`) on both the home page and the TV layout. Track info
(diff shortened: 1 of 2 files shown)


alvarovotan:
Hi @rob9315 

Just wanted to check in — I pushed a fix for the failing test earlier (the matches array had an invalid URL pattern with no path, which the connector test rejects). It should be all clear now on my end.

Let me know if there's anything else you'd like me to adjust!

rob9315:
the playbutton is never removed from the page or made invisible, as such the current code never thinks that music is playing.
I'd suggest checking the `<audio>` element in the DOM like other connectors already do, here's an example:
https://github.com/web-scrobbler/web-scrobbler/blob/25f9bffbb4910ea57bd5198a2ba0c9290662aecd/src/connectors/theindiebeat.ts#L31

alvarovotan:
Hi @rob9315 , thanks for catching that!

You're right, #playButton only swaps its icon and never gets hidden. I pushed a fix that reads playback state from the <audio> element instead (!document.querySelector('audio')?.paused), same approach as theindiebeat.ts.

Really appreciate how quickly you're reviewing this and all the helpful pointers, makes contributing a lot easier. Let me know if this looks good now!!


> the playbutton is never removed from the page or made invisible, as such the current code never thinks that music is playing. I'd suggest checking the `<audio>` element in the DOM like other connectors already do, here's an example:
> 
> https://github.com/web-scrobbler/web-scrobbler/blob/25f9bffbb4910ea57bd5198a2ba0c9290662aecd/src/connectors/theindiebeat.ts#L31

rob9315:
thanks!

i found one more thing actually,
there is a fullscreen mode on the website (bottom right) which prevents the trackArt from updating on song changes.

as an aside, the big comment at the beginning is not necessary, at least not that detailed, and is more of a maintenance hassle than help.

if you use mediaSession, you also don't need the selectors for the track, artist or trackart since the website seems to fill them in correctly.

the file could probably be reduced to only the player selector, the isPlaying logic, and the library call to use mediaSession to get the rest of the info :)

rob9315:
or if you desire, you could hook the audio element some more to trigger the onStateChanged action only when the audio element pauses, buffers or similar actions using `Util.bindListeners`

alvarovotan:
@rob9315 
 I won’t be able to make this adjustment right now—only in about 15 hours once I get back home. I really hope that’s in time to get everything sorted out and make it into the next app update :(

Please check back here tomorrow, as I’ll already have the fix done by then!



> or if you desire, you could hook the audio element some more to trigger the onStateChanged action only when the audio element pauses, buffers or similar actions using `Util.bindListeners`

alvarovotan:
Thanks again @rob9315 , all three points are addressed 🙏

- Dropped the track/artist/trackArt selectors and the big header comment — the file is now just the player selector, the isPlaying logic and useMediaSessionApi(), as you suggested.
- That also fixes the fullscreen issue: the artwork no longer depends on reading #coverArtwork (which gets hidden in ambient mode), it comes from MediaSession now.
- Took your Util.bindListeners suggestion too, hooking the <audio> element for playing / pause / waiting so onStateChanged fires on buffering as well.
- Tests, typecheck, prettier and eslint all pass locally.

By the way, do you have a rough idea of when the next Web Scrobbler release might go out? Just curious about when the station would actually start scrobbling for listeners, no pressure at all, I know releases happen when they happen. Thanks for all the guidance on this one, it's been a genuinely nice first contribution experience!


Update: tested it locally on Edge and everything's working, playback is detected and the artwork updates correctly in fullscreen now

alvarovotan:
@rob9315 
Thanks again for the review, both suggestions are in, and I found a third issue while testing.

**What changed since your review:**

1. `isPlaying` now reads the `<audio>` element instead of the play button, as you suggested
2. Added `Util.bindListeners` on the audio element, as you suggested
3. The station also injects an embed bar on every page of `outrahora.com`, which I had missed. Scrobbling silently did nothing there, so the connector now covers that domain too.

**About the embed bar:** it keeps its player inside a Shadow DOM and creates the audio element with `new Audio()`, so it's reachable neither via `document.querySelector('audio')` nor through the shadow root, `isPlaying` would return `true` even while paused. The connector branches on hostname: the station site keeps the audio listeners you reviewed, and the bar falls back to `useTabAudibleApi()`.

**One thing I fixed on the station side:** the site was filling MediaSession's `album` with the station name. Last.fm picks the artwork it shows from the album in the scrobble, so every track fell back to the placeholder icon, and it would have polluted every listener's library with an album that doesn't exist. The site now resolves the real album and leaves the field empty when there's no match.

**Tested end to end** on every entry point, the station site, its fullscreen mode, and the embed bar on `outrahora.com`. Scrobbles land correctly with artwork on all of them.

rob9315:
I simplified your url match to just `*://*.outrahora.com/*` since that should match everything
seems good, i don't see anything else in need of changing :+1:

alvarovotan:
Thanks so much for the review, @rob9315 really appreciate the help and the merge!

PR #6176, opened 2026-08-12
Fix qobuz.ts connector

CalendricalHeresy:

See my struggles in #6156. I finally solved it by changing nothing but the playerSelector to use the highest element that isn't discarded by qobuz every time the track changes. Fun times. Does not seem like an elegant solution AT ALL though so if anyone has a better idea I am all ears.

**Additional context**

### DIFF
--- a/src/connectors/qobuz.ts
+++ b/src/connectors/qobuz.ts
@@ -3,7 +3,7 @@ export {};
 const artistSelector = '.player__track-album [href*=artist]';
 const timeInfoSelector = '.player__track-time-content';
 
-Connector.playerSelector = '.player';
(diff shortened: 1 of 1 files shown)


CalendricalHeresy:
Ive been running this basically since the PR and it seems to work perfectly.

rob9315:
yeah, websites removing and rebuilding the dom on navigation / changes is super annoying. I've been looking at fixing this in the core of the extension instead, allowing connectors to opt into having a chain of mutationobservers listening for element removal and then going back to searching for the element like when it's not initially found.

I can't test this PR but your explanation seems sound so I'll merge this as is. I also don't know of a better way to fix this right now.

PR #6179, opened 2026-08-14
Add support for SoundScout

Cyclomactic:

Adds support for scrobbling from SoundScout (soundscout.com).

The new connector detects:
- Track title
- Artist
- Track duration
- Play/pause state
- Track changes

SoundScout dynamically adds its player when playback begins, so the connector observes a persistent player layout container to detect when the player and track information become available.

**Additional context**

Tested manually in Firefox with the development build.

Verified:
- Starting playback detects the current track
- Pausing and resuming updates the playback state
- Changing tracks detects the new track
- Track duration is detected correctly
- Now Playing is submitted successfully
- Tracks scrobble successfully

Validation:
- `npm test` — 2088 tests passed
- `npm run lint` — passed

### DIFF
--- a/src/connectors/soundscout.ts
+++ b/src/connectors/soundscout.ts
@@ -1,0 +1,15 @@
+export {};
+
+Connector.playerSelector = '[data-sidebar-collapsed][data-player-active]';
+
+Connector.artistSelector = '[data-testid="player-bar-track-artist"]';
+
+Connector.trackSelector = '[data-testid="player-bar-track-title"]';
+
+Connector.durationSelector = '[data-testid="player-bar-duration"]';
+
+Connector.isPlaying = () =>
+	Util.getAttrFromSelectors(
+		'[data-testid="btn-player-bar-play"]',
+		'aria-label',
+	) === 'Pause';
--- a/src/core/connectors.ts
+++ b/src/core/connectors.ts
@@ -87,6 +87,12 @@ export default <ConnectorMeta[]>[
 		id: 'soundcloud',
 	},
 	{
+		label: 'SoundScout',
+		matches: ['*://www.soundscout.com/*'],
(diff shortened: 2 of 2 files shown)


Cyclomactic:
This is my first PR, and it looks like I missed the required added label. I don't appear to have permission to add labels to the PR myself. Thanks for your patience!

rob9315:
sorry, i missed your pr and wrote my own in #6182 

it seems we wrote a very similar connector, was there a reason you selected the whole page and not just the player-bar?

Cyclomactic:
Initially I attempted to use the bar with:

Connector.playerSelector = '[data-testid="player-bar-track-title"]';

but got:

Element '[data-testid="player-bar-track-title"]' is missing

I then tested SoundScout after a fresh page load, before starting a track, and searched the DOM for player-bar. It returned zero results. From that testing, it appeared that SoundScout doesn't create the player bar until playback starts, while Web Scrobbler was trying to attach its observer when the connector initialized.

As a test, I changed the selector to:

Connector.playerSelector = 'body';

That immediately worked — when SoundScout later inserted the player, Web Scrobbler detected the change and the track. I then looked for something narrower and found SoundScout's persistent shell, which existed both before and after playback, so I ended up with:

Connector.playerSelector =
    '[data-sidebar-collapsed][data-player-active]';

Full disclosure, I wrote this with a lot of help from an LLM, so I'm definitely open to there being a better way to handle it. That said, I've been running this version locally since submitting the PR and it has worked reliably.

rob9315:
Alright, web-scrobbler has the ability to target elements that are only added later to the DOM, as long as they are never removed after being added. With that said I'll close this PR then, you can reopen it if you find any problems with the version that has been merged into `master` though

Issue #6033, opened 2026-05-10
Last.FM / account login problems


rob9315:
Many users are experiencing problems with Last.FM (or scrobbling accounts in general). We are having trouble reproducing some of these issues, and we need your help, if you encounter these

 there are a few different issues at play, so please read carefully which one applies to you before opening a new one:

### #6030: LastFM is geoblocking Russia
This is actually not an issue with web-scrobbler, but if you are in russia then Last.FM probably doesn't load for you at all, see that issue for how to circumvent it.

<details><summary>symptoms:</summary>

- infinite loading empty popup
  - firefox: <img width="196" height="82" alt="Image" src="https://github.com/user-attachments/assets/4d2b373a-cddb-4ae7-8172-8e50712bbf6a" style="display:inline-block" />
  - chromium: <img width="188" height="87" alt="Image" src="https://github.com/user-attachments/assets/1de3c049-4819-443b-aa9b-dc6337db4888" style="display:inline-block" />
- network errors like these in the extension's page: <img width="435" height="77" alt="Image" src="https://github.com/user-attachments/assets/78882d53-7623-487f-aa59-b168cca0d508" />
</details>

### #4351: Buttons on Account Page don't open any page

Caused by localstorage not being available to the extension, try running
```js
await browser.storage.local.get('LastFM')
```
in a console (Ctrl+Shift+I) on the extension's options page, if you do not get any value <details><summary>(like this:)</summary><img width="403" height="105" alt="Image" src="https://github.com/user-attachments/assets/d3ab9947-0e00-43a4-b647-7061b1f189ad" /></details> then it is definitely the same issue as #4351 **and we would appreciate information about your setup!** so that we can reproduce it and fix it

<details><summary>other symptoms:</summary>

- Buttons look normal but pressing them doesn't do anything: <img width="1876" height="1336" alt="Image" src="https://github.com/user-attachments/assets/80b462ac-515d-4f03-b0ba-f0553a01198f" />
- When scrobbling something you it tells you that you aren't signed in (duh) <img width="429" height="409" alt="Image" src="https://github.com/user-attachments/assets/24a833bb-c3c4-4c64-a966-7f56bb8b0b51" />
</details>

duplicates: #5170 (brave), #5989 (also firefox)

### #4783 uBlock Origin Lite complications

[...]

Issue #5989, opened 2026-05-01
connect account to last.fm button is not clickable


dagath614:
After reinstalling the application i needed to re-connect the  account, but as i click the button it does nothing, whether i use it with or without VPN. Last.fm itself is accessible with VPN, but the link won't work. Other options with clickable buttons (Libre.fm, MusicBrainz) are working as intended.

charblob:
I am getting the same issue as well. I am using the firefox browser.

rob9315:
can we get some [debug information](https://github.com/web-scrobbler/web-scrobbler/wiki/Debug-the-extension) please?

charblob:
When I try to click on "Sign In", it doesn't do anything and here is the console log:

Uncaught (in promise) Error: An unexpected error occurred 17 undefined
Last.fm: Error while fetching session [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
Error: An unexpected error occurred [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
Libre.fm: Error while fetching session [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
Error: An unexpected error occurred [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
ListenBrainz: Error while fetching session [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
Error: An unexpected error occurred [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
Uncaught (in promise) Error: An unexpected error occurred 3 undefined
Last.fm: Error while fetching session [context-menu-DebIYzG8.js:1:103099](moz-extension://bf272a20-951f-4b18-90e6-d70ce9088e64/assets/context-menu-DebIYzG8.js)
[...]

Issue #5839, opened 2026-02-27
Cannot log in to lastfm, help?


sammori03:
Help. Using Firefox. Was connected just fine for the past couple of days, then suddenly noticed nothing was scrobbling.

Refreshed the web scrobbler page and it said I wasn't connected to my lastfm account

Tried to log in to lastfm account from the account page. The "sign in" button might as well be a graphic. it doesn't redirect to any link

https://imgur.com/YjCcQ8O

  "manifest_version": 3,
  "name": "Web Scrobbler",
  "default_locale": "en",
  "description": "__MSG_extDescription__",
  "version": "3.18.0",
  "permissions": [
    "storage",
    "contextMenus",
    "notifications",
    "scripting"
  ],
  "host_permissions": [
    "http://*/",
    "https://*/"
  ],
  "content_scripts": [
    {
      "matches": [
        "<all_urls>"
      ],
      "js": [
        "content/main.js"
      ],
      "all_frames": true
    }
  ],
  "web_accessible_resources": [
    {
      "resources": [
        "connectors/*"
      ],
      "matches": [
        "<all_urls>"
      ]
    },
    {
      "resources": [
        "icons/*"
      ],
      "matches": [
        "<all_urls>"
      ]
    }
  ],
  "options_ui": {
    "page": "src/ui/options/index.html",
    "open_in_tab": true
  },
  "icons": {
    "16": "icons/icon_main_16.png",
    "48": "icons/icon_main_48.png",
    "96": "icons/icon_main_96.png",
    "128": "icons/icon_main_128.png",
    "256": "icons/icon_main_256.png",
    "512": "icons/icon_main_512.png"
  },
  "action": {
    "default_icon": {
      "16": "icons/action_unsupported_16_light.png",
      "19": "icons/action_unsupported_19_light.png",
      "32": "icons/action_unsupported_32_light.png",
      "38": "icons/action_unsupported_38_light.png"
    },
    "default_title": "__MSG_pageActionUnsupported__",
    "default_popup": "src/ui/popup/index.html"
  },
  "commands": {
    "toggle-connector": {
      "description": "__MSG_hotkeyToggleConnector__"
    },
    "love-song": {
      "description": "__MSG_hotkeyLoveSong__"
    },
    "unlove-song": {
      "description": "__MSG_hotkeyUnloveSong__"
    }
  },
  "background": {
    "scripts": [
      "background/main.js"
    ]
  },
  "browser_specific_settings": {
[...]

Issue #5170, opened 2025-02-04
I can't log into last.fm account


User-444:
When I tried to log into last.fm via the extension it didn't work

**How to reproduce**
1. Go to accounts
2. Click on sign in for last.fm

**Expected behavior**
A redirect to sign in to last.fm with the Web Scrobbler

Screen recording:
https://github.com/user-attachments/assets/166bc3e2-a89d-4f4e-8b08-96c614fcaa3b

**Debug logs**
[
](https://pastebin.com/ucB6jTpB)

**Environment**
-   OS: MacOS Sonoma 14.7
-   Browser: Brave: Version 1.74.51 Chromium: 132.0.6834.160
-   Extension version: 3.12.0

thekingrat:
I'm seeing this on Ubuntu 22.04, Vivaldi 7.1,  Extension 3.12.0.

EpicestGamer:
A friend and I are seeing a similar issue. Web Scrobbler was refusing to recognize or scrobble any tracks, eventually tried reinstalling, and the sign in button for last.fm is completely unresponsive.

uzkahya:
Same problem for me!

uzkahya:
> A friend and I are seeing a similar issue. Web Scrobbler was refusing to recognize or scrobble any tracks, eventually tried reinstalling, and the sign in button for last.fm is completely unresponsive.

I had totally same experience, it seems doesn't recognize any songs and showing the ''exclamation mark '' ( using safari btw)
and after that I went to settings on the extension and login button doesn't response at all.

u1traviolet:
Same issue here as well, works fine in Edge but in Brave unable to login

DolceJ0Lly:
doesn't work in windows 10 home edition, chrome.

kaikkeus:
I've had this experience for a long time. 

And this is kind of unrelated, but it's weird how for example VLC media player has never lost the connection (like you have to re-login with Web Scrobbler every now and then, apparently as designed by last.fm). VLC uses post.audioscrobbler.com BTW, which is old protocol but apparently might not get discontinued, for legacy reasons. But I guess it's better to use the new protocol especially since that's how it's working right now. But perhaps for some users, the old protocol could be available too? Oh, "we will not be issuing new client ids for this protocol ", so I guess not.

Ao1Pointblank:
I also have this issue on Brave.
[...]

PR #6186, opened 2026-08-21
core: fix circularity

rob9315:

- add `tsConfig` setting for `lint:circular` (`madge`)
- remove the dynamic import of Options from `@/core/util/debug` in favor of reading from `browser.storage.(sync||local).get('Options')` directly

**Additional context**

the CI for checking for circular dependencies had a lot of ignored files because it wasn't pointed to the `tsConfig` and as such it ignored all `@/` imports, which are the ones that have the circularity

to our analysis tools the dynamic import seemed to fix the issues with circularity, which it doesn't do in reality due to [vite not creating a new chunk just for that dynamic import][vite-chunk].

[vite-chunk]: https://github.com/vitejs/vite/discussions/13903

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lint:stylelint": "stylelint \"**/*.scss\"",
     "lint:prettier": "prettier . --check",
     "lint:markdown": "remark .",
-    "lint:circular": "madge --extensions ts,tsx --circular src",
+    "lint:circular": "madge --circular src",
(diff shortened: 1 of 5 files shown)


rob9315:
It is quite likely that this fixes the part of #6033 that is actually a problem with the extension.
concrete issues: #4351, #5170, #5989, #5170, #4922, #5839

Issue #5870, opened 2026-03-08
Single artist scrobbling


The-HungryCaterpillar:
**Single artist scrobbles are inconsistent**
Single artist scrobbling is desirable for a cleaner last.fm library and services such as Pano Scrobbler are able to reliably determine and select the first artist. For example, Pano Scrobbler correctly lists the single artist to this [song](https://music.youtube.com/watch?v=Wn0pp_KqaeE&si=p8Kv9ufxA6Q8wU1U) as Thaiboy Digital but Web Scrobbler records it as both artists (Thaiboy Digital & Bladee).

**Use Pano Scrobbler's System**
Pano Scrobbler appears to clean the string and then match it against some database to ensure it is valid, this is extended to ensure artists with more complicated names such as Earth, Wind & Fire are still correctly scrobbled. The link to the file that handles this is [here](https://github.com/kawaiiDango/pano-scrobbler/blob/main/composeApp/src/commonMain/kotlin/com/arn/scrobble/utils/FirstArtistExtractor.kt).

**Consequences**
I haven't looked but I assume Web Scrobbler compares the song to another, larger database and tries to ascribe artists to scrobbles from there. [Sun Goddess by Earth, Wind & Fire & Ramsey Lewis](https://music.youtube.com/watch?v=hbmCDMdBnPY&si=bPVE8LIKWFM_uGcp) is scrobbled as just Ramsey Lewis, this seems more correct as the song is also available on Ramsey Lewis' page and the album is credited in his name. Therefore it maybe wise to have a separate option for this type of first artist scrobbling in case it interrupts other users' setups.

Thank you for reading this.

rob9315:
it uses musicbrainz artists and the main parsing they are doing is actually from here

https://github.com/kawaiiDango/pano-scrobbler/blob/9cc12c375cfe2482acfe5ed284de12292baf21b3/py-scripts/collect_musicbrainz_artist_names.py

SamuelSHrk:
i have a similar issue when playing on web spotify https://open.spotify.com/track/2PMzMHzHpeNVn4KNryqKLy?si=3dfeb960aedb4fee

Numerosityy:
There shouldn't be any issue making this happen, right? There are settings for this on other scrobbling apps that function properly (ex: Pano Scrobbler)

Issue #5500, opened 2025-07-31
Artist tag on some services


audienciaon:
Is there a way to identify only the main (or first) artist?

On Spotify, only the first or main artist of a song is considered for scrobbling. But on other streaming services like Tidal and YouTube Music, two or more artists are counted, which creates an error tag and messes up the database and user profile.

Issue #4382, opened 2023-12-26
Standardize syntax for songs with multiple artists and adapt manual forms accordingly


AdamantConlanger:
**Is your feature request related to a problem? Please describe.**
When a song has multiple artists, and especially if a single album contains collab tracks whose listed artists may differ from song to song, Web Scrobbler often has trouble identifying and registering it properly. A good example is [Epic: The Troy Saga on Youtube Music](https://music.youtube.com/playlist?list=OLAK5uy_k5KIaPWJf8oxbr1W2AxY0VbHDOxcahoHg&si=LbXhPWyy4UXY5Oyv). Not only does this compromise the reliability of Web Scrobbler, it also invites different notational conventions into the data, because people decide on different approaches when manually entering the missing artists. If there were an agreed-upon syntax for songs with multiple artists, this would be much easier to reconcile, and an official implementation for automatically processing multi-artist tracks could be developed.

**Describe the solution you'd like**
The main goal would be to have the program automatically detect all the artists listed for a track, and store them in an unordered sequence data structure that I'll just call a list for simplicity's sake. Apps using Web Scrobbler to collect data should then be able to associate songs to each of the artists in that list individually instead of to the entire group of artists as a whole. That would also alleviate possible issues cropping up when we consider what happens when a user manually enters only a single artist instead of the whole caboose. Ideally, there'd be an "add extra artist" button in Web Scrobbler's GUI so no human element comes into play when delimiting artists. A possible issue with this whole approach is that some streaming services might not provide the artists in a sequence data structure, but as a string, which would require detailing to the identifier how to parse these strings for every one of these services, but seen as sites differ in their ways so wildly already, this probably won't actually pose as big of a problem as I'm making it out to be.

**Describe alternatives you've considered**
The way Web Scrobbler currently does things, with a single field for listing the artists of a song, likely has its advantages over a more elaborate scheme, and it may be sensible to want to keep it that way. In that case, we could instead elect to decide on an official syntax for listing multiple artists in this field. A designated delimiter character, like the comma, and a specified grammatical approach, specifically concerning the Oxford comma and whether to use "and", "&", or nothing at all before the last artist, would go a long way to consolidate many manual entries of the same tracks into one.

Issue #3532, opened 2022-12-29
in-browser Spotify scrobbles main artist + featured artist as one artist


krakenattackart:
I usually listen to Spotify on Chrome, and if for example if i'm scrobbling a song by Artist A feat. Artist B, the scrobble will be saved as the artist is "Artist A, Artist B". This creates issue because every song with a feature will be saved as its own artist, with its own album, with its own album cover, and so on.
I am completely oblivious to what could be done to fix this, I am just here to report the issue.

yayuyokitano:
This is a bit of a rough one, and is what the album artist tag is there for. Specifying an album artist will make the album save to that main artist.
However, web scrobbler does not have any logic to set album artist tag automatically at the moment, and I'm not 100% sure how to go about it (though if someone reading this has ideas do speak up).
You can manually add album artist of course, though it's not ideal.

jaccarmac:
See #2946, #3306, #3402. I also remember but can't find one other issue discussing whether artists including features *should* be on Last.fm in the first place. I'm of the mind that they should, that changing the default behavior of the scrobbler would be incorrect, and that album artist selectors are the best way to group tracks by album. Unfortunately, said selector is the squirreliest and there often isn't a clean way to implement it.

PR #6190, opened 2026-08-23
Add option to scrobble only the first artist

### LINKED ISSUE #5496: For songs with multiple artists, add an option to scrobble only the first artist
It would be nice to add an option "Only use the first artist when scrobbling a song made by multiple artists".
For example, this song : https://open.spotify.com/intl-fr/track/3lNIeXZbtNZty4WgNa7Xtb would scrobbled with the "Pendulum" artist, instead of "Pendulum, Devin Townsend" which is not a valid artist.

Doggo785:
Fixes #5496
Related to #5870, #5500, #3532, #4382, #2946. See #3432 for why this is off by default.

### Problem
Many connectors return collaborations as one string, like "Artist A feat. Artist B". That scrobbles as a single artist and creates a wrong entry.

### Solution
New global toggle, "Only scrobble the first artist", off by default (`Options.FIRST_ARTIST_ONLY`). This addresses #3432, always trimming would break bands that contain separators. When enabled, a pipeline step trims the artist before scrobbling and for the metadata lookup. Longest-prefix matching against an allowlist keeps names like "Earth, Wind & Fire" intact.

### Changes
- Pipeline step `src/core/object/pipeline/first-artist.ts` wired into `pipeline.ts`, runs only when the option is enabled
- Extractor `src/core/scrobbler/lastfm/first-artist-extractor.ts` with separator list from `src/core/scrobbler/lastfm/separators.json`
- Allowlist `src/core/scrobbler/lastfm/artist-allowlist.ts` loading `src/static-data/musicbrainz_artist_hashes.bin` (~124k xxh3_64 hashes) via lazy `hash-wasm@4.12.0`
- UI toggle in `src/ui/options/components/options/global-options.tsx` under General, strings in `en/messages.json`, metadata query override in `metadata.ts`
- Build plumbing: copy `static-data` in `vite.configs.ts`, allow `wasm-unsafe-eval` in manifest CSP, update `.prettierignore` and Xcode project

### How to test
- Enable the toggle under Options, General
- Play a track with "Artist A feat. Artist B" and confirm only "Artist A" is scrobbled and used for metadata lookup
- Try allowlisted names like "Earth, Wind & Fire" and confirm they do not split
- Run `npm test` (new cases: first-artist-extractor, pipeline-first-artist, metadata-first-artist, separators anti-divergence, artist-allowlist)
- Disable the toggle and confirm the full string scrobbles as before

### Notes
Separators live in `separators.json` as the shared source of truth for both the extractor and `scripts/data/generate-musicbrainz-allowlist.py`. That script is a clean-room reimplementation of pano-scrobbler's collector. To regenerate, run it and commit the updated `.bin`.

`hash-wasm` is lazy-loaded, so xxh3 runs in wasm only when the option is on. No heavy dependency in the main bundle.

This is pipeline-level and global, unlike #6065 which is YouTube Music only and connector-scoped. No conflict, this supersedes it for the general case. Happy to coordinate.


rob9315:
hi, i have a few questions:
1. why did you choose to use a hash of the keccak family of cryptographic algorithms for this?
2. how did you implement the file generator from pano-scrobbler "clean-room" when you are the only committer?, why do you introduce python to this project that has thus far been javascript/typescript mostly?
3. not really a question, just: please disclose ai-assisted contributions.

Doggo785:
Thanks for the review @rob9315!

1. Not Keccak, XXH3_64 from xxHash. It's non-cryptographic. I used hash-wasm because pano stores its allowlist as XXH3_64 and I need the same output. First try xxhash-wasm had broken XXH3, so ff6accfc switches to hash-wasm with lazy createXXHash3(). The readme lists Keccak/SHA3 among many hashes, that may have caused the confusion.

Web Crypto only does SHA, it can't make XXH3. A Map of 124k names would be ~2-3 MB vs 978 KB for the binary.

2. "clean-room" was loose. I'm the only committer, so no separation. It's reimplemented without copying pano, just inspired. The Python script is build-time only, not shipped. I kept Python because the dump handling is short there. Happy to port it to TypeScript if you prefer.

3. Disclosure: extractor, allowlist loader, separators.json and the five test files were drafted with AI help, then I fixed them by hand (hash swap, longest-prefix scan, real-hash tests). Tests 2214 passing, tsc clean. No GPL code was copied.

Let me know what to change.

PR #6207, opened 2026-09-04
fix(youtube): preserve unattributed cover titles

### LINKED ISSUE #6206: [youtube] cover songs misattributed with original artist - title (cover) in title
**Describe the bug**
Youtube videos with the title format "[song title] - [band name] cover" or similar sometimes get scrobbled incorrectly

**How to reproduce**
Here is an example where a cover of Radiohead's "High and dry" by the band thanks mom with the title "High and Dry - Radiohead cover" gets scrobbled as 
Track: Radiohead cover
Artist: High and dry
[example 1](https://www.youtube.com/watch?v=Gbbcot6ieNU)

Here is an example where a cover of the same song by one Mackenzie Johnson with the title "High And Dry - Radiohead cover" is scrobbled correctly as
Track: High and dry - Radiohead cover
Artist: Mackenzie Johnson
[example 2](https://www.youtube.com/watch?v=ysTbovhLFHI)

The only difference I can see is that the title was "High and Dry" in the first video and "High And Dry" in the second, but it might be that one cover artist is recognized and the other isn't? 

Here's another example where the title "Nude - Radiohead (cover)" is scrobbled incorrectly as
Track: Radiohead (Cover)
Artist: Nude
[example 3](https://www.youtube.com/watch?v=uxhlXG4_NEQ)

Videos with "cover by [cover artist]" or similar in the title seem to be handled correctly. 

**Expected behavior**
Youtube videos with only [og band name], [song title], and "cover" in the title should be scrobbled in some standardised fasion with the artist taken to be the youtube channel name. 

-   OS: Mac
-   Browser: Firefox
-   Extension version: 3.22.0

mikemikimike:
## Summary / Problem

YouTube titles that end with a cover marker, such as `High and Dry - Radiohead cover` or `Nude - Radiohead (cover)`, can be split as if the original artist were the uploader. This produces incorrect artist and track metadata when the covering artist is not included in the title.

Fixes #6206

## Changes

- Detect a standalone trailing `cover` marker without an explicitly named covering artist.
- Preserve the complete video title as the track value so the YouTube connector can use the channel name as the artist.
- Add regression coverage for both reported title formats while keeping the existing `cover by Artist` parsing path unchanged.

## Tests

- `npx vitest run tests/content/util.test.ts` (157 tests passed)
- `npm run test -- --run` (2164 tests passed)
- `npm run checkts`
- `npm run lint`
- `git diff --check`
- `npm run dist chrome` and `npm run dist firefox` in a Linux Docker build environment

## Compatibility / Known limitations

The new behavior only applies to a standalone `cover` marker at the end of a title, with an optional closing bracket. Titles that identify the covering artist, such as `cover by Artist`, continue using the existing parsing rules.

The Windows build could not run locally because the `canvas` native binary was unavailable after its prebuilt download failed during a TLS connection, and Cairo/GTK development headers are not installed. Both Chrome and Firefox distributables built successfully in Linux Docker.

## Issue link

https://github.com/web-scrobbler/web-scrobbler/issues/6206

### DIFF
--- a/src/core/content/util.ts
+++ b/src/core/content/util.ts
@@ -795,6 +795,16 @@ export const ytTitleRegExps = [
 ];
 
 /**
+ * Detect a cover marker that does not identify the covering artist.
+ *
+ * A title such as "Track - Original Artist cover" is not an artist-track
+ * pair: the artist is the channel owner, while the complete title is the
+ * most useful track value. Keep "cover by Artist" on the regular parsing
+ * path, since it already provides an artist.
+ */
+const ytCoverTitleRegExp = /\bcover\b\s*[)\]】）]?\s*$/i;
+
+/**
  * Extract artist and track from Youtube video title.
  * @param videoTitle - Youtube video title
  * @returns Object containing artist and track fields
@@ -832,6 +842,11 @@ export function processYtVideoTitle(
 
 	// MV/PV if ending and with whitespace in front
 	title = title.replace(/\s+(MV|PV)$/i, '');
+
+	// A trailing cover marker without an artist must use the channel name.
+	if (ytCoverTitleRegExp.test(title)) {
+		return { artist, track: title };
+	}
 
 	// Try to match one of the regexps
 	for (const regExp of ytTitleRegExps) {
--- a/tests/content/util.test.ts
+++ b/tests/content/util.test.ts
@@ -471,6 +471,18 @@ const PROCESS_YT_VIDEO_TITLE_DATA = [
 		expected: { artist: 'Artist', track: 'Track' },
 	},
 	{
+		description:
+			'should keep the complete title when a cover artist is not specified',
+		args: ['High and Dry - Radiohead cover'],
+		expected: { artist: null, track: 'High and Dry - Radiohead cover' },
+	},
+	{
+		description:
+			'should keep the complete title when a parenthesized cover marker is used',
+		args: ['Nude - Radiohead (cover)'],
+		expected: { artist: null, track: 'Nude - Radiohead (cover)' },
+	},
+	{
 		description: 'should process tracks with separators and quotes',
 		args: ['Artist - "Track Name"'],
 		expected: { artist: 'Artist', track: 'Track Name' },


src/connectors/youtube.ts
```diff
@@ -512,15 +512,17 @@ function getTrackInfoFromChapters() {
...
 		Util.getTextFromSelectors(videoTitleSelector),
+		artistFromChannel,
 	);
 	if (!artist) {
-		const channelName = Util.getTextFromSelectors(channelNameSelector);
-		const re =
-			// eslint-disable-next-line no-irregular-whitespace
-			/^(?:Mavzu|Тема|الموضوع|ਵਿਸ਼ਾ)\s[–-]\s|(?:(?:\s[-—–]|[:՝])\s(?:Onderwerp|Mövzu|Topik|tema|Tema|téma|Emne|Thema|teema|Topic|gaia|Paksa|Sujet|Isihloko|Efni|Mada|tēma|téma|emne|temat|Tópico|Subiect|aihekanava|Ämne|Chủ đề|Konu|тэма|Тема|Тақырып|Сэдэв|тема|Θέμα|թեմա|נושא|موضوع|عنوان|विषय|বিষয়বস্তু|বিষয়|મુદ્દો|ବିଷୟ|தலைப்பு|అంశం|ವಿಷಯ|വിഷയം|මාතෘකාව|หัวข้อ|ຫົວ​ຂໍ້|ခေါင်းစဉ်|თემა|ርዕስ|ប្រធាន​បទ|主题|主題|トピック|주제)|\s\(tema\))$/;
-		artist = channelName?.replace(re, '') ?? null;
+		artist = artistFromChannel;
```
rob9315: you only have to use the topic channel regex on the channel if processYtVideoTitle didn't match anything.
Topic Channels only publish videos with correct metadata anyways because they get the metadata from the record labels themselves.

mikemikimike: Fixed in 76edf37. getTrackInfoFromTitle now passes the raw channel name to processYtVideoTitle, and applies the Topic-channel normalization only when no artist is parsed.

src/core/content/util.ts
```diff
@@ -357,6 +359,10 @@ export function fillEmptyFields(
 	}
 
 	for (const field of fields) {
+		if (unsafeStateFields.has(field as string)) {
+			continue;
+		}
+
```
rob9315: this is unnecessary
if someone passes these fields in, that's their own fault in connector development.
```suggestion
```

src/core/content/util.ts
```diff
@@ -339,6 +339,8 @@ export function isArtistTrackEmpty(
 	return !(artistTrack && artistTrack.artist && artistTrack.track);
 }
 
+const unsafeStateFields = new Set(['__proto__', 'constructor', 'prototype']);
+
```
rob9315:
```suggestion
```

tests/content/util.test.ts
```diff
@@ -972,6 +972,19 @@ const FILL_EMPTY_FIELDS = [
...
+			{ artist: null },
+			Object.assign(Object.create(null), {
+				['__proto__']: 'polluted',
+				constructor: 'polluted',
+				prototype: 'polluted',
+			}),
+			['__proto__', 'constructor', 'prototype'],
+		],
+		expected: { artist: null },
+	},
```
rob9315:
```suggestion
```
unnecessary, as described elsewhere

